PR #4, opened 2025-07-10
Add `Device` resource + controller and cisco nxos provider support

felix-kaestner:


afritzler (changes requested):

afritzler (changes requested):

afritzler (approved):
@felix-kaestner do you have an idea why the linter is still timeouting? You think we can merge that?

swagner-de (approved):

nikatza (approved):

PR #8, opened 2025-07-11
cisco-nxos-provider: add `ISIS` configuration

nikatza:
Following limitations apply to the current implementation:
- only ISIS (no OSPF)
- only one IGP process per device


felix-kaestner (approved):
Only some nits about comments. Nothing blocking a merge from my perspective. :)

felix-kaestner (approved):
@nikatza Pls squash all commits when merging, so we don't end up with "PR comments" commits in the history.

PR #19, opened 2025-08-04
Add `conditions` utility package for updating status conditions

felix-kaestner:


nikatza (commented):
Sorry I can't finish the review, but I have a couple of comments that repeat themselves and need clarification before I become too repetitive :) 

Where do come the Configured and Operational Conditions from exactly? Meaning: when and where are they set on run-time?  For example: What does it mean for Banner to have an Operational Condition? 

To me it seems this PR is includes a generalized approach that we will later iterate and  fine-tune (for each object). I just want to be sure I am on the right track here.

nikatza (approved):

PR #30, opened 2025-08-19
Add device and provider integration

felix-kaestner:
This patch changes the way that provider-specific configuration is linked to resource, by introducing a generic `providerConfigRef` field that points to a k8s resource that will be read and passed to the provider implementation.

Additionally, the parent/owner device of an interface resource will now be referenced using a required field in the spec. It will now be the responsibility of the interface controller to set the device label and owner reference on the interface resources as they are reconciled.


hardikdr (commented):
I have dropped few optional suggestions.

afritzler (approved):

PR #40, opened 2025-08-26
cisco-nxos-provider: decouple ospf, isis, pim from pkg `iface`

nikatza:
see commit messages from more details


felix-kaestner (approved):
Just some formatting. Otherwise looks nice!

PR #44, opened 2025-09-01
cisco-nxos-provider: refactor `ACL` configuration

felix-kaestner:
This patch rewrites the implementation for configuring access control
lists on Cisco NX-OS devices.

It supports the following configuration:

```
ip access-list ACL-SNMP-VTY
 10 permit ip 10.0.0.0/8 any
```

### DIFF
--- a/internal/provider/cisco/nxos/acl/acl.go
+++ b/internal/provider/cisco/nxos/acl/acl.go
@@ -4,8 +4,10 @@
 package acl
 
 import (
+	"errors"
 	"fmt"
 	"net/netip"
+	"strings"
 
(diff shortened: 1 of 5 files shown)


nikatza (changes requested):
It seems possible to me that it is possible to configure an IPv6 address on `Item.Source`. Which in our case it should rise an error, right? (i mean, i would not let the the operator send the query). If this is correct, could we add an `Option` to do the validation? 

In this line: these ACLs apply only to IPv4 and is not clear in the doc that this is the case. May be we should already some sort of IPv6 configuration Option as well and return an `error` for it?

nikatza (approved):

PR #63, opened 2025-09-12
cisco-nxos-provider: add support for management interface

felix-kaestner:
This patch adds support for querying the management interface (`mgmt0`) as part of the `iface.Exists` and `iface.ShortName` functions. This is provided mainly to add compatibility for cases, when a configuration item references an interface, to have a single point of checks that can be integrated, without checking for the management interface everywhere.


nikatza (changes requested):
Looks good to me, I think we are missing updting the tests thought.

nikatza (approved):

PR #96, opened 2025-10-21
Cisco NXOS vPC Domain 

nikatza:
Multi-Chassis Link Aggregation (MC-LAG) is quite vendor and platform
specific. We don't see much intersection in their respective
configuration to justify a common API type. Instead, we move forward
with a platform specific API exclusive to Cisco NXOS devices.

In this PR we add types, controller, and provider code to configure
virtual Port Channels (vPCs) via the operator as:
```
feature vpc

vpc domain 5
  peer-switch
  role priority 100
  system-priority 10
  peer-keepalive destination 10.1.1.2 source 10.1.1.1 vrf VPC_KEEPALIVE
  delay restore 150
  peer-gateway
  layer3 peer-router
  auto-recovery reload-delay 360
  delay restore interface-vlan 140
  fast-convergence

interface port-channel1
  vpc peer-link
 ```

The `vpcdomain` controller  ensures that the `vpc peer-link` is configured. 
This is because on gNMI this property is configured  at the vpcDom (sub-)containers.
Having this property as a provider-specific resource does not seem a good 
option and the moment as it complicates the code base.

The operational status of the resource is `UP` if the peer is alive and the remote
device returns a positive uptime value for the peer.

### CODE REVIEW

File: api/cisco/nx/v1alpha1/vpcdomain_types.go [hunk 1]
```diff
@@ -0,0 +1,311 @@
...
+	DomainID int16 `json:"domainId"`
+
+	// AdminState is the administrative state of the vPCDomain domain (enabled/disabled).
+	// When disabled, the vPCDomain domain is administratively shut down.
+	// +required
+	AdminState AdminState `json:"adminState"`
+
+	// RolePriority is the role priority for this vPCDomain domain (1-65535).
+	// The switch with the lower role priority becomes the operational primary.
+	// Maps to: "role priority <RolePriority>"
```
felix-kaestner:
```suggestion
```
As agreed, we should not map it to CLI commands. Applies also to other fields.

api/cisco/nx/v1alpha1/vpcdomain_types.go
```diff
@@ -0,0 +1,311 @@
...
+	KeepAlive KeepAlive `json:"keepalive"`
+
+	// AutoRecovery defines auto-recovery settings for restoring vPCDomain after peer failure.
+	// +optional
+	AutoRecovery *AutoRecovery `json:"autoRecovery,omitempty"`
+
+	// Switch enables peer-switch functionality on this peer.
+	// When enabled, both vPCDomain peers use the same spanning-tree bridge ID, allowing both
+	// to forward traffic for all VLANs without blocking any ports.
+	// +required
```
felix-kaestner: Can we make this and the following fields optional and default to `{enabled: false}`?

api/core/v1alpha1/groupversion_info.go
```diff
@@ -75,6 +75,11 @@ const (
 	// intended (e.g., a interface is up). It corresponds to the "oper-status" commonly found
 	// in the OpenConfig models (or "operSt" on Cisco).
 	OperationalCondition = "Operational"
+
+	// DegradedCondition indicates whether the resource is in a degraded state.
+	// This condition indicates whether the resource is experiencing issues that impact its performance
+	// or functionality, but does not render it completely non-functional.
+	DegradedCondition = "Degraded"
```
felix-kaestner: This does not seem to be used.

api/cisco/nx/v1alpha1/vpcdomain_types.go
```diff
@@ -0,0 +1,311 @@
...
+	// +optional
+	PeerStatusMsg string `json:"peerStatusMsg,omitempty"`
+
+	// PeerUptime indicates how long the vPCDomain peer has been up and reachable via keepalive.
+	// +optional
+	PeerUptime metav1.Duration `json:"peerUptime,omitempty"`
+
+	// PeerLinkIf is the name of the interface used as the vPCDomain peer-link.
+	// +optional
+	PeerLinkIf string `json:"peerLinkIf,omitempty"`
```
felix-kaestner: Isn't that already known by the spec?

api/cisco/nx/v1alpha1/vpcdomain_types.go
```diff
@@ -0,0 +1,311 @@
...
+// +kubebuilder:printcolumn:name="Enabled",type=string,JSONPath=`.spec.adminState`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Configured",type=string,JSONPath=`.status.conditions[?(@.type=="Configured")].status`,priority=1
+// +kubebuilder:printcolumn:name="Operational",type=string,JSONPath=`.status.conditions[?(@.type=="Operational")].status`,priority=1
+// +kubebuilder:printcolumn:name="PeerStatus",type=string,JSONPath=`.status.peerStatus`,priority=1
+// +kubebuilder:printcolumn:name="Role",type=string,JSONPath=`.status.role`,priority=1
+// +kubebuilder:printcolumn:name="PeerLinkIf",type="string",JSONPath=".status.peerLinkIf",priority=1
+// +kubebuilder:printcolumn:name="PeerLinkIfOperSt",type="string",JSONPath=".status.peerLinkIfOperStatus",priority=1
+// +kubebuilder:printcolumn:name="KA-PeerStatus",type=string,JSONPath=`.status.keepaliveStatus`,priority=1
+// +kubebuilder:printcolumn:name="KA-PeerUptime",type="string",JSONPath=`.status.peerUptime`,priority=1
```
felix-kaestner: I think we should not use abbreviations here, as it makes it really hard to see what they represent, if you are not familiar with those abbreviations. I generally think writing those words out is better.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2143,6 +2105,168 @@ func (p *Provider) ResetSystemSettings(ctx context.Context) error {
...
+		v.AutoRecovery = NewOption(AdminStDisabled)
+		if peer.AutoRecovery.Enabled {
+			v.AutoRecovery = NewOption(AdminStEnabled)
+		}
+		if peer.AutoRecovery.ReloadDelay > 0 {
+			v.AutoRecoveryReloadDelay = NewOption(uint16(peer.AutoRecovery.ReloadDelay)) // #nosec G115 -- kubebuilder validation
+		}
+	}
+
+	ipaddr := net.ParseIP(peer.KeepAlive.Destination)
```
felix-kaestner: I think we don't need to parse the ipv4 address here, as the validation should already take care that it's correct. Other than that, we don't use the parsed `ipaddr` here. So I think this can be removed.

internal/provider/cisco/nxos/vpc.go
```diff
@@ -1,16 +1,107 @@
 // SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
 // SPDX-License-Identifier: Apache-2.0
-
```
felix-kaestner: This space should be kept. Otherwise some tools interpret the license header as a comment defined on the package statement.

internal/provider/cisco/nxos/intf.go
```diff
@@ -167,11 +167,21 @@ type PortChannelMember struct {
 	Force bool   `json:"isMbrForce,omitempty"`
 }
 
-func NewPortChannelMember(name string) *PortChannelMember {
-	return &PortChannelMember{
+type PortChannelMemberOption func(*PortChannelMember)
+
+func WithForce(force bool) PortChannelMemberOption {
+	return func(m *PortChannelMember) { m.Force = force }
+}
```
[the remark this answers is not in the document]
nikatza: I don't see how this as an exclusive decision, but adds more flexibility to the library and find it helpful for developing. If you want I will remove it.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2143,6 +2105,168 @@ func (p *Provider) ResetSystemSettings(ctx context.Context) error {
...
+	}
+	v.KeepAliveItems.DestIP = peer.KeepAlive.Destination
+
+	ipaddr = net.ParseIP(peer.KeepAlive.Source)
+	if ipaddr == nil {
+		return fmt.Errorf("invalid keep-alive source IP address %q", peer.KeepAlive.Source)
+	}
+	v.KeepAliveItems.SrcIP = peer.KeepAlive.Source
+
+	v.KeepAliveItems.VRF = "default"
```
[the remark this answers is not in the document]
nikatza: As you noticed earlier, this should be `management` and is automatically set by the device. So I will remove this line since is wrong.

internal/controller/cisco/nx/vpcdomain_controller.go
```diff
@@ -0,0 +1,522 @@
...
+					return true
+				},
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return true
+				},
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldVRF := e.ObjectOld.(*corev1.VRF)
+					newVRF := e.ObjectNew.(*corev1.VRF)
+					return oldVRF.Spec.Name != newVRF.Spec.Name ||
+						!equality.Semantic.DeepEqual(oldVRF.Status, newVRF.Status)
```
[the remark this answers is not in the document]
nikatza: I can't reconstruct for`vrf`, but for interface: the switch can shutdown the port-channels that form the peer link, so I wanted to force a reconciliation to update the vpc domain status. If you prefer I can remove it.

internal/provider/cisco/nxos/intf.go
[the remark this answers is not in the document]
felix-kaestner: I guess my comment was more regarding the annotation, which is something the user can control. And I'm just worried that giving out this option, will eventuell lead down the path of always using it, for the lack of better orchestration, while if we don't have this option, we have to ensure that our orchestration works in a correct manner.
However, this isn't a blocker for me, just something I wanted to raise my concerns on.

internal/controller/cisco/nx/vpcdomain_controller.go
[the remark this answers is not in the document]
felix-kaestner: Makes sense. Could we then make the check a bit more specify? E.g. checking for a transition in the operational condition? I'm just a bit worried, that if we check for equality in the whole status subresource, we might end up adding more unrelated properties to the status later, which will then trigger unecessary reconcilations of the vpc domain, for something that doesn't effect the vpc domain. So basically, only checking for things, that we actively account for in the vpc domain.

internal/controller/core/interface_controller.go
```diff
@@ -100,6 +100,15 @@ func (r *InterfaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		return ctrl.Result{}, nil
 	}
 
+	var cfg *provider.ProviderConfig
```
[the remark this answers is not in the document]
nikatza: This should not have been there, apologies.

felix-kaestner (approved):

PR #102, opened 2025-10-30
Refactor `Interface` IPv4 Configuration

felix-kaestner:
This change refactors the ipv4 address assignment of the `Interface` resource to use proper types instead of using strings with an internal format.


nikatza (commented):
This was much necessary. Thanks for pushing these changes!

nikatza (approved):

PR #109, opened 2025-11-04
Add `BGP` and `BGPPeer`  resources and controllers

felix-kaestner:
This PR introduces comprehensive BGP (Border Gateway Protocol) support to the network-operator by adding two new Custom Resource Definitions (CRDs): `BGP` and `BGPPeer`, along with their respective controllers and provider implementations.


nikatza (approved):
very nice.

PR #111, opened 2025-11-05
Add LACPDUs counter information into `InterfaceStatus`

felix-kaestner:


nikatza (approved):
This looks great, thanks a lot.

I have a question since there is no info in the commit message.  Why do we need  to push this info to the interface status? (besides this being nice to have).

PR #121, opened 2025-11-18
Add `EVPNInstance` resource and controller

felix-kaestner:
This change introduces a new custom resource to configure a VXLAN mapping for a VLAN (Bridge Domain) and it's corresponding EVPN control plane context as a MAC-VRF.


nikatza (approved):
Looks good to me. I've left a comment pointing to a PR that might conflict.

PR #122, opened 2025-11-18
Allow to use a IRB Interface as redundant distributed anycast gateway

felix-kaestner:


nikatza (approved):
Looks good to me. 

The `gNMI` call will fail if `feature fabric` is not enabled, or the the MAC address is not configured  on the `nve` interface, right?

PR #123, opened 2025-11-19
Network Virtualization Edge (NVE)

nikatza:
Implements an NVE API core type and an extension for Cisco NX.


api/cisco/nx/v1alpha1/vtepconfig_types.go
```diff
@@ -0,0 +1,81 @@
...
+	corev1alpha1 "github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+)
+
+// +kubebuilder:rbac:groups=nx.cisco.networking.metal.ironcore.dev,resources=vtepconfigs,verbs=get;list;watch
+
+// VTEPConfig defines the Cisco-specific configuration of a VTEP/NVE
+type VTEPConfigSpec struct {
+	// AdvertiseVMAC is equivalent to: `advertise virtual-rmac`
+	// +optional
+	AdvertiseVMAC *bool `json:"advertiseVMAC,omitempty"`
```
felix-kaestner:
```suggestion
	AdvertiseVirtualMAC *bool `json:"advertiseVirtualMac,omitempty"`
```
According to the kubernetes api conventions, this should be written out.

api/cisco/nx/v1alpha1/vtepconfig_types.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// HoldDownTime is equivalent to: `source-interface hold-down-time
```
felix-kaestner: I would rather like to explain what this option does and not mention how it's configured on the CLI. Same applies for the other fields of the spec.

api/core/v1alpha1/prefix_types.go
```diff
@@ -77,3 +77,50 @@ func (in *IPPrefix) DeepCopy() *IPPrefix {
 	in.DeepCopyInto(out)
 	return out
 }
+
+func (in *IPPrefix) First() netip.Addr {
+	if !in.IsValid() {
+		return netip.Addr{}
+	}
```
felix-kaestner:
```suggestion
```
Based on the documentation for `Masked` this is not required.

https://pkg.go.dev/net/netip@go1.24.3#Prefix.Masked
> If p is zero or otherwise invalid, Masked returns the zero [Prefix](https://pkg.go.dev/net/netip@go1.24.3#Prefix).

internal/controller/core/suite_test.go
```diff
@@ -60,6 +61,9 @@ var _ = BeforeSuite(func() {
 	err = v1alpha1.AddToScheme(scheme.Scheme)
 	Expect(err).NotTo(HaveOccurred())
 
+	err = nxosv1alpha1.AddToScheme(scheme.Scheme)
```
felix-kaestner: We should not leak this provider api group into the "core". The core should be standalone, with no dependency on a provider-specific api group.

api/core/v1alpha1/vtep_types.go
```diff
@@ -0,0 +1,163 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceRef is immutable"
+	DeviceRef LocalObjectReference `json:"deviceRef"`
+
+	// ProviderConfigRef is a reference to a resource holding the provider-specific configuration for this VTEP.
+	// +optional
+	ProviderConfigRef *TypedLocalObjectReference `json:"providerConfigRef,omitempty"`
+
+	// Enable indicates whether the interface is administratively up or down.
+	// +required
+	Enabled bool `json:"enabled"`
```
felix-kaestner: Can we use a proper `AdminState` type as on the `Interface`?

api/core/v1alpha1/vtep_types.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// PrimaryInterfaceRef is the reference to the loopback interface used for the primary VTEP IP address.
+	// +required
+	PrimaryInterfaceRef LocalObjectReference `json:"primaryInterfaceRef"`
+
+	// AnycastInterfaceRef is the reference to the loopback interface used for anycast VTEP IP address.
+	// +required
+	AnycastInterfaceRef LocalObjectReference `json:"anycastInterfaceRef,omitempty"`
```
felix-kaestner:
```suggestion
	// +optional
	AnycastInterfaceRef *LocalObjectReference `json:"anycastInterfaceRef,omitempty"`
```
This should actually be optional.

api/core/v1alpha1/vtep_types.go
```diff
@@ -0,0 +1,163 @@
...
+	// +optional
+	MulticastGroup *MulticastGroup `json:"multicastGroup,omitempty"`
+}
+
+// HostReachabilityType defines the method used for host reachability.
+type HostReachabilityType string
+
+const (
+	HostReachabilityTypeBGP           HostReachabilityType = "BGP"
+	HostReachabilityTypeFloodAndLearn HostReachabilityType = "FloodAndLearn"
```
felix-kaestner:
```suggestion
    // HostReachabilityTypeFloodAndLearn uses data-plane learning for MAC addresses.
	HostReachabilityTypeFloodAndLearn HostReachabilityType = "FloodAndLearn"
```

api/core/v1alpha1/vtep_types.go
```diff
@@ -0,0 +1,163 @@
...
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:path=vteps
+// +kubebuilder:resource:singular=vtep
+// +kubebuilder:printcolumn:name="VTEP",type=string,JSONPath=`.spec.name`
+// +kubebuilder:printcolumn:name="Device",type=string,JSONPath=`.spec.deviceRef.name`
+// +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
```
felix-kaestner: Can we add Source Interface, Anycast Source Interface and Host Reachability as columns?

internal/controller/core/vtep_controller.go
```diff
@@ -0,0 +1,427 @@
...
+
+	if s.ProviderConfig != nil {
+		if err := r.validateProviderConfigRef(ctx, s); err != nil {
+			conditions.Set(s.VTEP, metav1.Condition{
+				Type:    v1alpha1.ConfiguredCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  v1alpha1.IncompatibleProviderConfigRef,
+				Message: fmt.Sprintf("ProviderConfigRef is not compatible with Device: %v", err),
+			})
+			return fmt.Errorf("provider config reference is not compatible with the device: %w", err)
```
felix-kaestner: These should use the `reconcile.TerminalError` https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#pkg-functions such that they don't trigger a exponential backoff retrying (as there is no point in retrying). Also applies to the errors returned below.

I would directly set the conditions in the `validateX` funcs and then return a `reconcile.TerminalError` from those funcs and then within this `reconcile` func just `return err` as is.

internal/provider/cisco/nxos/provider.go
```diff
@@ -1737,6 +1661,87 @@ func (p *Provider) DeleteVRF(ctx context.Context, req *provider.VRFRequest) erro
...
+		n.HostReach = HostReachFloodAndLearn
+	default:
+		return fmt.Errorf("vtep: invalid evpn host reachability type %q", req.VTEP.Spec.HostReachability)
+	}
+	n.SuppressARP = req.VTEP.Spec.SuppressARP
+
+	// provider-specific settings
+	vc := new(nxosv1alpha1.VTEPConfig)
+	if err := req.ProviderConfig.Into(vc); err != nil {
+		return fmt.Errorf("vtep: decode provider config: %w", err)
```
```suggestion
		return fmt.Errorf("vtep: failed to decode provider config: %w", err)
```
For me there is a verb missing here.

internal/provider/cisco/nxos/provider.go
```diff
@@ -1737,6 +1661,87 @@ func (p *Provider) DeleteVRF(ctx context.Context, req *provider.VRFRequest) erro
...
+	}
+
+	if vc.Spec.HoldDownTime > 0 {
+		n.HoldDownTime = vc.Spec.HoldDownTime
+	}
```
```suggestion
	n.HoldDownTime = vc.Spec.HoldDownTime
```
With the default of 180 we can omit the if check.

internal/provider/provider.go
```diff
@@ -301,30 +301,30 @@ type VRFRequest struct {
...
-type EnsurePIMRequest struct {
-	PIM            *v1alpha1.PIM
-	Interfaces     []PIMInterface
-	ProviderConfig *ProviderConfig
+	// EnsureVRF call is responsible for VRF realization on the provider.
+	EnsureVTEP(context.Context, *VTEPRequest) error
+	// DeleteVRF call is responsible for VRF deletion on the provider.
+	DeleteVTEP(context.Context, *VTEPRequest) error
+	// GetInterfaceStatus call is responsible for retrieving the current status of the Interface from the provider.
+	GetStatusVTEP(context.Context, *VTEPRequest) (VTEPStatus, error)
```
felix-kaestner:
```suggestion
	GetVTEPStatus(context.Context, *VTEPRequest) (VTEPStatus, error)
```
nit: Would follow this naming scheme.

internal/webhook/cisco/nx/v1alpha1/vtepconfig_webhook.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// validateVTEPConfigSpec performs validation to enforce that the VLAN ranges
+// - are strictly non overlapping
+// - the number of vlans configured does not exceed 512
+// - the IDs must be in the range 1-3967
+func validateVTEPConfigSpec(vc *v1alpha1.VTEPConfig) error {
+	var vlanRanges []rng
+
+	validateRange := func(start, end uint) error {
```
felix-kaestner: Any reason to define this new function here. I think I would just move the logic into the for loop below.

internal/provider/cisco/nxos/nve.go
```diff
@@ -10,19 +10,23 @@ import (
...
-	HoldDownTime     int16         `json:"holdDownTime"`
-	HostReach        HostReachType `json:"hostReach"`
+	HoldDownTime     uint16        `json:"holdDownTime,omitempty"`
+	HostReach        HostReachType `json:"hostReach,omitempty"`
 	McastGroupL2     string        `json:"mcastGroupL2,omitempty"`
 	McastGroupL3     string        `json:"mcastGroupL3,omitempty"`
-	SourceInterface  string        `json:"sourceInterface"`
-	SuppressARP      bool          `json:"suppressARP"`
+	SourceInterface  string        `json:"sourceInterface,omitempty"`
+	SuppressARP      bool          `json:"suppressARP,omitempty"`
```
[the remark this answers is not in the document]
nikatza: Thanks for the explanation. Why should should be using here a `Patch`?

internal/provider/provider.go
```diff
@@ -301,30 +301,30 @@ type VRFRequest struct {
...
 
-type DeletePIMRequest struct {
-	PIM            *v1alpha1.PIM
-	ProviderConfig *ProviderConfig
+type VTEPStatus struct {
+	// OperStatus indicates whether the VTEP is operationally up (true) or down (false).
+	OperStatus bool
+	// OperStatus indicates if the primary and anycast interfaces are operationally up (true) or down (false).
+	OperStatusPrimaryInterface bool
+	OperStatusAnycastInterface bool
```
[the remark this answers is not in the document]
nikatza: My plan was to show this also in the summary, but it might indeed not be necessary. However, this comes for free and is the same as the `Interface`. To my understanding, the `.status` represent the configuration on the remote device. Taking this value from the resource means the waiting for a reconciliation. Is this what we want here?

api/cisco/nx/v1alpha1/vtepconfig_types.go
```diff
@@ -0,0 +1,84 @@
...
+// VLANListItem represents a single VLAN ID or a range start-end. If ID is set, rangeMin and rangeMax must be absent. If ID is absent, both rangeMin
+// and rangeMax must be set.
+// +kubebuilder:validation:XValidation:rule="!has(self.rangeMax) || self.rangeMax > self.rangeMin",message="rangeMax must be greater than rangeMin"
+// +kubebuilder:validation:XValidation:rule="has(self.id) || (has(self.rangeMin) && has(self.rangeMax))",message="either ID or both rangeMin and rangeMax must be set"
+// +kubebuilder:validation:XValidation:rule="!has(self.id) || (!has(self.rangeMin) && !has(self.rangeMax))",message="rangeMin and rangeMax must be omitted when ID is set"
+type VLANListItem struct {
+	// +optional
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=3967
+	ID uint16 `json:"id,omitempty"`
```
felix-kaestner: Should not use unsigned integer types. Also applies to other fields in this file.

internal/controller/core/vtep_controller.go
```diff
@@ -0,0 +1,427 @@
...
+	if err != nil {
+		return fmt.Errorf("vtep: failed to validate primary interface reference: %w", err)
+	}
+
+	anycastIf, err := r.validateInterfaceRef(ctx, s.VTEP.Spec.AnycastInterfaceRef.Name, s)
+	if err != nil {
+		return fmt.Errorf("vtep: failed to validate anycast interface reference: %w", err)
+	}
+
+	if primaryIf.Name == anycastIf.Name {
```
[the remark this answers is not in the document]
felix-kaestner: What you are using here is the Kubernetes resource name, not the Interface handle name (which would be in `.spec.Name`). So what I meant was that you don't need the check, as you already have https://github.com/ironcore-dev/network-operator/blob/b2f80a4d390301b195a4812685a60fe66c678be0/api/core/v1alpha1/nve_types.go#L14 which takes care of requiring a different resource name to be referenced.

Anyways, not being able to have more than one `Interface` resource per device per handle (e.g. 'lo0') should ultimately be enforced. If that is the case, I think we would be good here and don't need to check as part of the NVE controller, right?

Maybe as a suggestion for now, let's allow the referenced interfaces to refer to the same interface handle, and gracefully handle it in the provider implementation. So in the example you shared, the provider implementation would do something like (pseudo code):
```go
if req.AnycastSourceInterfaceName != "" && req.AnycastSourceInterfaceName != req.SourceInterfaceName {
   // .. configure anycast source interface differently
}
```

internal/controller/core/nve_controller.go
```diff
@@ -0,0 +1,518 @@
...
+		if apierrors.IsNotFound(err) {
+			conditions.Set(s.NVE, metav1.Condition{
+				Type:    v1alpha1.ConfiguredCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  v1alpha1.WaitingForDependenciesReason,
+				Message: fmt.Sprintf("interface resource '%s' not found in namespace '%s'", intf.Name, intf.Namespace),
+			})
+			return nil, reconcile.TerminalError(fmt.Errorf("member interface %q not found", s.NVE.Spec.SourceInterfaceRef.Name))
+		}
+		return nil, reconcile.TerminalError(fmt.Errorf("failed to get member interface %q: %w", s.NVE.Spec.SourceInterfaceRef.Name, err))
```
felix-kaestner:
```suggestion
			return nil, reconcile.TerminalError(fmt.Errorf("referenced interface %q not found", s.NVE.Spec.SourceInterfaceRef.Name))
		}
		return nil, err
```
I think the second error case we actually want to retry, as this could be due to some intermittent issues with the api server.

[nit]: not really a member interface.

felix-kaestner (approved):

PR #125, opened 2025-11-20
Fix `gnmiext` list unmarshaling and skipped updates

felix-kaestner:


internal/provider/cisco/gnmiext/v2/client.go
```diff
@@ -326,6 +354,7 @@ func (c *Client) Unmarshal(b []byte, dst any) (err error) {
 	if ok && b[0] == '[' && b[len(b)-1] == ']' {
 		b = b[1 : len(b)-1]
 	}
+	zeroUnknownFields(b, reflect.ValueOf(dst))
```
nikatza: The commit message explains very well why we need this; but I am asking myself if this will/should  always be the behavior for the `Unmarshal` function? May be its worth anyway mentioning this in the function doc.

nikatza (approved):

PR #127, opened 2025-11-21
Default list of allowed vlans switchport interfaces to all VLANs

felix-kaestner:
This change removes the requirement for `Interface` resources that are configured as switchports (L2) to specify their list of allowed vlans (trunk vlans). If unset, the configuration should default to allow all VLANs (1-4094) on the particular link.

Note: No changes were necessary on the provider implementation, as this was already handled in this way.


api/core/v1alpha1/interface_types.go
```diff
@@ -117,7 +117,6 @@ const (
 // Switchport defines the switchport configuration for an interface.
 // +kubebuilder:validation:XValidation:rule="self.mode != 'Access' || has(self.accessVlan)", message="accessVlan must be specified when mode is Access"
 // +kubebuilder:validation:XValidation:rule="self.mode != 'Trunk' || has(self.nativeVlan)", message="nativeVlan must be specified when mode is Trunk"
-// +kubebuilder:validation:XValidation:rule="self.mode != 'Trunk' || has(self.allowedVlans)", message="allowedVlans must be specified when mode is Trunk"
```
nikatza: We could also remove the `nativeVlan` requirement

nikatza (approved):

PR #134, opened 2025-12-03
Add device provisioning

swagner-de:
- **Rename and refactor bootstrap, implement basic provisioning functions**
- **Add provisioning HTTP servers**


swagner-de:
Will open multiple PRs for this

PR #138, opened 2025-12-08
[NX-OS] Add `BorderGateway` resource and controller

felix-kaestner:
This change introduces a new custom resource called `BorderGateway` which is specific to the Cisco NX-OS Platform and allows to configure required features for the border gateway functionality required to deploy Cisco VXLAN EVPN Multi-Site.


api/cisco/nx/v1alpha1/bordergateway_types.go
```diff
@@ -0,0 +1,201 @@
...
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=281474976710655
+	// +kubebuilder:validation:ExclusiveMaximum=false
+	MultisiteID int64 `json:"multisiteId"`
+
+	// SourceInterfaceRef is a reference to the loopback interface used as the source for the
+	// border gateway virtual IP address. A best practice is to use a separate loopback address
+	// for the NVE source interface and multi-site source interface. The loopback interface must
+	// be configured with a /32 IPv4 address. This /32 IP address needs be known by the transient
+	// devices in the transport network and the remote VTEPs.
```
nikatza: Would it make sense to add a webhook to enforce this requirement?  (it doesn't need to be done here if the answer is yes)

internal/provider/cisco/nxos/provider.go
```diff
@@ -2143,6 +2143,139 @@ func (p *Provider) ResetSystemSettings(ctx context.Context) error {
...
+	Interface *v1alpha1.Interface
+	Tracking  nxv1alpha1.InterconnectTrackingType
+}
+
+type BorderGatewayPeer struct {
+	BGPPeer  *v1alpha1.BGPPeer
+	PeerType nxv1alpha1.BGPPeerType
+}
+
+func (p *Provider) EnsureBorderGatewaySettings(ctx context.Context, req *BorderGatewaySettingsRequest) error {
```
nikatza: Usually we pre-append an updated with the required features.

nikatza (approved):
Really nice.

PR #140, opened 2025-12-11
Rename and refactor bootstrap, implement basic provisioning functions

swagner-de:
This renames `DeviceSpec.Bootstrap` to `Provisioning` in order to have a
unified naming. We also implement a ProvisioningProvider interface and
state transitionings in the device controller.


api/core/v1alpha1/device_types.go
```diff
@@ -122,6 +163,44 @@ type DeviceStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
 }
 
+type ProvisioningInfo struct {
+	StartTime  metav1.Time `json:"startTime"`
+	EndTime    metav1.Time `json:"endTime,omitempty"`
+	RebootTime metav1.Time `json:"reboot,omitempty"`
```
felix-kaestner:
```suggestion
	EndTime    metav1.Time `json:"endTime,omitzero"`
	RebootTime metav1.Time `json:"reboot,omitzero"`
```
Should use `omitzero` as `metav1.Time` is a struct.

api/core/v1alpha1/device_types.go
```diff
@@ -122,6 +163,44 @@ type DeviceStatus struct {
...
+	_, err := rand.Read(token)
+	if err != nil {
+		return nil, err
+	}
+	entry := ProvisioningInfo{
+		StartTime: metav1.Now(),
+		Token:     hex.EncodeToString(token),
+	}
+	d.Status.Provisioning = append(d.Status.Provisioning, entry)
+	return &d.Status.Provisioning[len(d.Status.Provisioning)-1], nil
```
felix-kaestner:
```suggestion
	return &entry, nil
```
As we just appended it, we have the last item.

api/core/v1alpha1/device_types_test.go
```diff
@@ -0,0 +1,262 @@
...
+func TestDevice_CreateProvisioningEntry(t *testing.T) {
+	now := metav1.Now()
+	oneHourAgo := metav1.NewTime(now.Add(-1 * time.Hour))
+	twoHoursAgo := metav1.NewTime(now.Add(-2 * time.Hour))
+
+	tests := []struct {
+		name        string
+		device      *Device
+		wantErr     bool
+		errContains string
```
felix-kaestner: I would not assert on the error message. Just checking if an error occurs with `wantErr` is fine. Also callers should never assert/check on an error string.

api/core/v1alpha1/device_types_test.go
```diff
@@ -0,0 +1,262 @@
...
+		validate    func(t *testing.T, device *Device, entry *ProvisioningInfo)
```
felix-kaestner: This validate func feels weird. If validation is that different between test cases, I would simply split them into distinct `Test...` functions, like: 

```go
func TestDevice_CreateProvisioningEntry_Success(t *testing.T) {
  // ...
}

func TestDevice_CreateProvisioningEntry_Success_Completed(t *testing.T) {
  // ...
}

...
```

internal/controller/core/device_controller.go
```diff
@@ -133,29 +126,46 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 		})
 		obj.Status.Phase = v1alpha1.DevicePhaseProvisioning
 		r.Recorder.Event(obj, "Normal", "ProvisioningStarted", "Device provisioning has started")
-		// TODO(swagner-de): Start POAP Process.
-		_ = tmpl // <-- Use the template.
 		return ctrl.Result{RequeueAfter: r.RequeueInterval}, nil
```
felix-kaestner:
```suggestion
		return ctrl.Result{}, nil
```
Can we remove this, so we immediately start the provisioning?

internal/controller/core/device_controller.go
```diff
@@ -133,29 +126,46 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
+			return ctrl.Result{RequeueAfter: 5 * time.Minute}, nil
 		}
-		log.Info("Device provisioning is complete, updating status")
-		conditions.Set(obj, metav1.Condition{
-			Type:    v1alpha1.ReadyCondition,
-			Status:  metav1.ConditionTrue,
-			Reason:  v1alpha1.ReadyReason,
-			Message: "Device is ready for use",
-		})
+		if activeProv.StartTime.Add(1 * time.Hour).After(time.Now()) {
```
felix-kaestner:
```suggestion
		if activeProv.StartTime.Add(time.Hour).After(time.Now()) {
```

internal/controller/core/device_controller.go
```diff
@@ -133,29 +126,46 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
+		activeProv := obj.GetActiveProvisioning()
+		if activeProv == nil {
+			err := errors.New("device went into provisioning completed phase, but no active provisioning found")
+			log.Error(err, "Failed to finalize provisioning")
+			return ctrl.Result{}, err
+		}
+		if !activeProv.RebootTime.IsZero() && activeProv.RebootTime.Time.Add((1 * time.Minute)).Before(time.Now()) {
+			log.Info("Device is rebooting, requeuing")
+			return ctrl.Result{RequeueAfter: 1 * time.Minute}, nil
+		}
```
felix-kaestner:
```suggestion
		if !activeProv.RebootTime.IsZero() && activeProv.RebootTime.Time.Add((time.Minute)).Before(time.Now()) {
			log.Info("Device is rebooting, requeuing")
			return ctrl.Result{RequeueAfter: time.Minute}, nil
		}
```

internal/provider/cisco/nxos/provider.go
```diff
@@ -83,6 +79,29 @@ func (p *Provider) Disconnect(_ context.Context, _ *deviceutil.Connection) error
 	return p.conn.Close()
 }
 
+func (p *Provider) HashProvisioningPassword(password string) (hashed, encryptType string, err error) {
+	salt := make([]byte, 10)
+	if _, err := rand.Read(salt); err != nil {
+		return "", "", err
+	}
+	s := [10]byte{}
+	copy(s[:], salt)
```
felix-kaestner: nit: I think we can directly read into the array.

```suggestion
	s := [10]byte{}
	if _, err := rand.Read(s[:]); err != nil {
		return "", "", err
	}
```

api/core/v1alpha1/device_types.go
```diff
@@ -16,7 +20,7 @@ type DeviceSpec struct {
 	// Bootstrap is an optional configuration for the device bootstrap process.
```
felix-kaestner:
```suggestion
	// Provisioning is an optional configuration for the device provisioning process.
```
nit: this comment should be updated.

internal/controller/core/device_controller.go
```diff
@@ -133,29 +127,46 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
+		return ctrl.Result{RequeueAfter: 20 * time.Minute}, nil
+
+	case v1alpha1.DevicePhaseProvisioningCompleted:
+		// we will finalize the provisioning here, either wait for the reboot to be initiated/completed
+		// or run post-provisioning checks instantly if no reboot is required
+		activeProv := obj.GetActiveProvisioning()
+		if activeProv == nil {
+			err := errors.New("device went into provisioning completed phase, but no active provisioning found")
+			log.Error(err, "Failed to finalize provisioning")
+			return ctrl.Result{}, err
```
felix-kaestner:
```suggestion
			return ctrl.Result{}, reconcile.TerminalError(err)
```
We should use a [reconcile.TerminalError](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#TerminalError) as there is no point in retrying on this error.

felix-kaestner (approved):

Issue #148, opened 2026-01-08
Cisco NX-OS: YANG configuration is not persisted in version <10.6(3) when mixed with feature activation


felix-kaestner:
## Problem

As confirmed by Cisco, NX-OS prior to version 10.6(3) has an issue that deals with the internal candidate DB used with gNMI when mixing feature and dependent configurations. As a result, when feature activation and dependent configuration is mixed within the same gNMI request, the configuration is not persisted to the internal config DB despite being persisted in YANG. When an operator queries YANG for the configuration, it is returned with the expected values despite not showing up e.g. in a `show running-config`.

## Proposed Solution

As suggested also by Cisco, this bug can be avoided by performing two RPC transactions. E.g.

**Features Config**

```
gnmic -a 10.15.33.42:50051 -u admin -p <pass> --skip-verify set --update-path 'System/fm-items/bgp-items' --update-value '{"adminSt":"enabled"}' --update-path 'System/fm-items/nvo-items' --update-value '{"adminSt":"enabled"}' --update-path 'System/fm-items/evpn-items' --update-value '{"adminSt":"enabled"}' 
```

**Dependent Config**

```
gnmic -a 10.15.33.42:50051 -u admin -p <pass> --skip-verify set --update-path 'System/eps-items/epId-items/Ep-list[epId=1]' --update-value '{"adminSt":"enabled","epId":1,"hostReach":"bgp","sourceInterface":"lo1"}
```

## Tasks

- [ ] Implement workaround for versions prior to 10.6(3)
- [ ] Verify that the issue is fixed in 10.6(3) – currently (as of Jan 2026) in internal QA

felix-kaestner:
Verified that the following does work on Cisco NX-OS 10.6(3) and persists configuration correctly:

```bash
gnmic -a 10.15.33.42:50051 -u admin -p <pass> --skip-verify set --update-path 'System/fm-items/bgp-items' --update-value '{"adminSt":"enabled"}' --update-path 'System/fm-items/nvo-items' --update-value '{"adminSt":"enabled"}' --update-path 'System/fm-items/evpn-items' --update-value '{"adminSt":"enabled"}' --update-path 'System/eps-items/epId-items/Ep-list[epId=1]' --update-value '{"adminSt":"enabled","epId":1,"hostReach":"bgp","sourceInterface":"lo1"}'
```

```
leaf1# sh run | sec 'interface nve1'
interface nve1
  no shutdown
  host-reachability protocol bgp
[...]

PR #155, opened 2026-01-19
[NX-OS] Separate feature activation from configuration

felix-kaestner:
## Problem

NX-OS versions prior to 10.6(3) have an issue where YANG configuration is not persisted when feature activation is mixed with dependent configuration in the same gNMI request. The configuration appears in YANG queries but not in the running config.

## Solution

This PR implements a workaround by separating feature activation from dependent configuration into separate gNMI transactions for affected versions (≤10.6(2)).

Relates to #148


nikatza (approved):
nice

PR #166, opened 2026-01-29
Do not wait for interfaces to become ready to configure ospf

swagner-de:
OSPF would never be configured if only one member interface is down.
Hence, only wait for configuration.

### DIFF
--- a/internal/conditions/conditions.go
+++ b/internal/conditions/conditions.go
@@ -70,6 +70,19 @@ func IsReady(target Getter) bool {
 	return condition.Status == metav1.ConditionTrue
 }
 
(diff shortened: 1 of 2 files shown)


internal/conditions/conditions.go
```diff
@@ -70,6 +70,19 @@ func IsReady(target Getter) bool {
 	return condition.Status == metav1.ConditionTrue
 }
 
+// isConfigured looks at the [v1alpha1.ConfiguredCondition] condition type and returns true
```
felix-kaestner:
```suggestion
// IsConfigured looks at the [v1alpha1.ConfiguredCondition] condition type and returns true
```

felix-kaestner (approved):

PR #173, opened 2026-02-04
Make sure salt does not contain zero bytes.

### LINKED ISSUE #172: Cisco NX-OS: 2-5% of Devices Fail ZTP Boot Due to Incorrect Password String Generation
## Description
During Zero Touch Provisioning (ZTP) boot, 2-5% of Cisco NX-OS devices fail to provision with valid admin credentials, requiring manual password reset and preventing scalable automation.

## Credentials Delivery during ZTP
We provision users by sending only scrypt-encrypted, salted password hashes (never plaintext) to prevent credential leakage:
1. **[Generate random 10-byte salt](https://github.com/ironcore-dev/network-operator/blob/d1e3ae48748964d5e7f4da42e35462980ba15caf/internal/provider/cisco/nxos/provider.go#L87-L90)** using crypto/rand
2. **[Create scrypt hash](https://github.com/ironcore-dev/network-operator/blob/d1e3ae48748964d5e7f4da42e35462980ba15caf/internal/provider/cisco/nxos/user.go#L142-L143)** (N=16384, r=8, p=1, keyLen=32)
3. **[Build NX-OS password string](https://github.com/ironcore-dev/network-operator/blob/d1e3ae48748964d5e7f4da42e35462980ba15caf/internal/provider/cisco/nxos/user.go#L146-L147)** using **[Cisco's undocumented custom base64 alphabet](https://github.com/ironcore-dev/network-operator/blob/d1e3ae48748964d5e7f4da42e35462980ba15caf/internal/provider/cisco/nxos/user.go#L154)**: `./0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz`
4. Format as: `$nx-scrypt$<base64_salt>$<base64_key>`
5. Apply as `username $user password 9 $passwordString role network-admin` to the `scheduled-config`, the configuration that is replayed after the POAP boot has finished.

## Suspected Issue
**Incorrect base64 alphabet or other hash failures** Since Cisco provides no official documentation on type 9 password generation, we reverse-engineered the format. The custom base64 alphabet was sourced from some corner of the internet and may be incorrect.

## Reproduction
To confirm our suspicions I had claude [generate a little test tool ](https://gist.github.com/swagner-de/b7eec87136d04c58ef1711c3e741010f#file-main-go). This tool creates different credentials, creates them as different users on the device and tries them out. If I create the all users with the same password, just with different salts, I reach around 1-10 credentials that do not work. I think this pretty much confirms our suspicion that something breaks while generating the password string.

The tool is self-contained an can be run against any Cisco NXOS device using
`go run main.go -configure -count 100 -password my-cisco-test123 -host 1.1.1.1 -admin-user admin -admin-pass Cisco123`

An example result [can be seen here](https://gist.github.com/swagner-de/b7eec87136d04c58ef1711c3e741010f#file-results-txt).

## Vendor Engagement
We are currently seeking vendor support to get official information on the password string generation.

## Workarounds

1. Use pre-validated salt+password combinations (high risk of leaking credentials)
2. Disable ZTP (current approach - blocks automation)
3. Use static ZTP password, rotate post-boot (security risk window)

ZTP disabled until resolution. If vendor support is slow, will implement workaround 3.

swagner-de:
Cisco cannot handle zero bytes in the salt. Those lead to password hash
becoming invalid and the credentials being unusable.
Closes #172.

### DIFF
--- a/internal/provider/cisco/nxos/provider.go
+++ b/internal/provider/cisco/nxos/provider.go
@@ -4,6 +4,7 @@
 package nxos
 
 import (
+	"bytes"
 	"cmp"
 	"context"
 	"crypto/rand"
@@ -85,8 +86,17 @@ func (p *Provider) Disconnect(_ context.Context, _ *deviceutil.Connection) error
 
 func (p *Provider) HashProvisioningPassword(password string) (hashed, encryptType string, err error) {
 	s := [10]byte{}
-	if _, err := rand.Read(s[:]); err != nil {
-		return "", "", err
+	for {
+		// Read cryptographically secure random bytes into the slice.
+		if _, err := rand.Read(s[:]); err != nil {
+			return "", "", err
+		}
+		// Check if the slice contains a zero byte.
+		if !bytes.Contains(s[:], []byte{0}) {
+			// If no zero is found, we're done. Break the loop.
+			break
+		}
+		// If a zero was found, the loop will repeat, overwriting the slice.
 	}
 	e := Scrypt{Salt: s}
 	hashed, pwdEncryptType, err := e.Encode(password)


internal/provider/cisco/nxos/provider.go
```diff
@@ -83,9 +83,30 @@ func (p *Provider) Disconnect(_ context.Context, _ *deviceutil.Connection) error
...
+	return salt, nil
+}
+
 func (p *Provider) HashProvisioningPassword(password string) (hashed, encryptType string, err error) {
-	s := [10]byte{}
-	if _, err := rand.Read(s[:]); err != nil {
+	s, err := generateRandomCiscoSalt()
+	if err != nil {
 		return "", "", err
 	}
```
```suggestion
    // Loop until we have a slice with no zero bytes.
    for {
        if _, err := rand.Read(s[:]); err != nil {
            return "", "", err
	}
        if !bytes.Contains(s[:], []byte{0}) {
            break
        }
    }
```
lets do this directly inline.

felix-kaestner (approved):

PR #176, opened 2026-02-04
Adding AAA TACACS Support

weneghawi:
## Summary

Restructured the core AAA API to align with the [OpenConfig `system/aaa` YANG model](https://openconfig.net/projects/models/schemadocs/jstree/openconfig-system.html), making it vendor-agnostic and suitable for multi-vendor support (Nokia, Juniper, Arista, etc.). All Cisco NX-OS specific configuration has been moved to a dedicated `AAAConfig` provider CRD. RADIUS server group support has been added alongside the existing TACACS+ implementation.

### Core API Changes (`api/core/v1alpha1/aaa_types.go`)

- **Server Groups**: Replaced flat `TACACSServers` + `TACACSGroup` with `ServerGroups []AAAServerGroup` — protocol-agnostic containers with nested servers, following OpenConfig `/system/aaa/server-groups/server-group`. Supports both `TACACS` and `RADIUS` group types.
- **RADIUS Support**: Added `AAAServerRADIUS` struct with `authPort` (default 1812), `acctPort` (default 1813), and `keySecretRef`
- **Method Lists**: Flattened `Authentication`, `Authorization`, and `Accounting` to simple method lists (removed NX-OS specific nesting like `Login.Default/Console` and `ConfigCommands`)
- **Field Renames**: `VRF` -> `VrfName`, `SourceInterface` -> `SourceInterfaceName` (leaves room for future object references)
- **Removed Cisco-specific fields**: `KeyEncryption`, `LoginErrorEnable` moved to Cisco `AAAConfig` CRD
- **Added XValidation rules**:
  - At least one of `serverGroups`, `authentication`, `authorization`, or `accounting` must be set
  - Servers in a TACACS group must have `tacacs` config
  - Servers in a RADIUS group must have `radius` config
  - `groupName` is required when method type is `Group`
  - `deviceRef` is immutable

### Cisco AAAConfig CRD (`api/cisco/nx/v1alpha1/aaaconfig_types.go`)

- Added `ConsoleAuthentication *NXOSMethodList` — NX-OS: `aaa authentication login console`
- Added `ConfigCommandsAuthorization *NXOSMethodList` — NX-OS: `aaa authorization config-commands default`
- Added `RADIUSKeyEncryption` type (Type6/Type7/Clear) with `radiusKeyEncryption` field (default Type7)
- Retains `KeyEncryption` (Type6/Type7/Clear) and `LoginErrorEnable`

### Controller (`internal/controller/core/aaa_controller.go`)

- Updated secret loading to iterate nested `ServerGroups[].Servers[].TACACS.KeySecretRef` and `ServerGroups[].Servers[].RADIUS.KeySecretRef`
- Updated `secretToAAA` watch mapping to trigger reconciliation on changes to both TACACS and RADIUS key secrets

### NX-OS Provider (`internal/provider/cisco/nxos/`)

- **aaa.go**: Added `RadiusProvider`, `RadiusProviderGroup`, `RadiusProviderRef` NX-OS DME structs. Added `MapRADIUSKeyEncryption` helper. Added `groupTypeByName` and `MapRealmFromGroup` to correctly resolve realm as `"radius"` or `"tacacs"` based on the referenced server group type. Removed read-only `Name` and `Realm` fields from `AAADefaultAuthor` (NX-OS rejects writes to these). Added `MapNXOSRealm`, `MapNXOSLocal`, `MapNXOSFallback` helpers. Note: RADIUS on NX-OS requires no feature flag (unlike TACACS+ which requires `feature tacacs+`).
- **provider.go**: Rewrote `EnsureAAA` to iterate `ServerGroups` with a `switch` on group type covering both TACACS and RADIUS. Rewrote `DeleteAAA` with batched resets and RADIUS group/server cleanup. Changed from `Patch` to `Update` (netconf replace).

### Sample YAML (`config/samples/networking_v1alpha1_aaa.yaml`)

- Updated to demonstrate the new structure with `serverGroups`, nested `servers`, flat method lists, and separate Cisco `AAAConfig` with console/config-commands authorization.

## Test Plan

- [x] `go build ./...` — compiles cleanly
- [x] `go test ./api/... ./internal/provider/... ./internal/clientutil/...` — all pass
- [x] `make run-golangci-lint` — 0 issues
- [x] `make generate` — CRDs and deepcopy regenerated
- [x] Verified full AAA lifecycle on NX-OS switch via gNMI (gnmic):
  - Enable TACACS+ feature, create servers + server group with VRF, source interface, server refs
  - Set default authentication, console authentication, config-commands authorization, accounting
  - Read back and verify all TACACS+ configuration
  - Reset all auth/authz/acct to local, delete group and servers, disable feature, verify clean state
  - Create RADIUS server + server group, set default auth realm to `radius`
  - Read back and verify all RADIUS configuration (confirmed: no feature flag needed on NX-OS)
  - Delete RADIUS group and server, verify clean state


api/core/v1alpha1/aaa_types.go
```diff
@@ -0,0 +1,280 @@
...
+	// +required
+	KeySecretRef SecretKeySelector `json:"keySecretRef"`
+
+	// KeyEncryption specifies the encryption type for the key.
+	// Type7 is the Cisco Type 7 encryption (reversible).
+	// Type6 is the AES encryption (more secure).
+	// Clear means the key is sent in cleartext (not recommended).
+	// +optional
+	// +kubebuilder:validation:Enum=Type6;Type7;Clear
+	// +kubebuilder:default=Type7
```
felix-kaestner: We don't want to include cisco specific stuff in the core api (we want this to be vendor agnostic). So at a minimum, we have to remove "Type7". Is Type6/AES universally supported across vendors (Nokia, Juniper, Arista etc.)?

api/core/v1alpha1/aaa_types.go
```diff
@@ -0,0 +1,280 @@
...
+	// +required
+	// +listType=set
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=16
+	Servers []string `json:"servers"`
+
+	// VRF is the VRF to use for communication with the TACACS+ servers.
+	// +optional
+	// +kubebuilder:validation:MaxLength=63
+	VRF string `json:"vrf,omitempty"`
```
felix-kaestner:
```suggestion
	VrfName string `json:"vrfName,omitempty"`
```
Let's use a string name here, to make clear that we are talking about the name of the VRF directly. This would leave us the possibility have "VRF" field in the future, which could then be a resource reference to a "VRF" resource.

config/rbac/role.yaml
```diff
@@ -28,20 +28,10 @@ rules:
   - list
   - update
   - watch
-- apiGroups:
```
felix-kaestner: This should not have been removed.

config/samples/networking_v1alpha1_aaa.yaml
```diff
@@ -0,0 +1,103 @@
...
+spec:
+  # Reference to the Device this AAA configuration belongs to
+  deviceRef:
+    name: my-switch
+
+  # TACACS+ server configuration
+  # Equivalent to:
+  #   tacacs-server host 10.16.8.142 key 7 $TYPE_7_SECRET
+  #   tacacs-server host 10.16.8.32 key 7 $TYPE_7_SECRET
+  #   tacacs-server host 169.145.33.51 key 7 $TYPE_7_SECRET
```
felix-kaestner: Please leave out Cisco NX-OS specific config lines from these samples. These should be vendor-agnostic.

internal/provider/cisco/nxos/aaa.go
```diff
@@ -0,0 +1,195 @@
...
+	_ gnmiext.Configurable = (*TacacsPlusProvider)(nil)
+	_ gnmiext.Configurable = (*TacacsPlusProviderGroup)(nil)
+	_ gnmiext.Configurable = (*AAADefaultAuth)(nil)
+	_ gnmiext.Configurable = (*AAAConsoleAuth)(nil)
+	_ gnmiext.Configurable = (*AAADefaultAuthor)(nil)
+	_ gnmiext.Configurable = (*AAADefaultAcc)(nil)
+)
+
+// TACACSFeature enables/disables the TACACS+ feature on NX-OS.
+// Path: System/fm-items/tacacsplus-items/adminSt
```
felix-kaestner: Please reuse the `Featuer` type from https://github.com/ironcore-dev/network-operator/blob/5b49d2f1a87a32219783abd41a9eb2ff6a192f31/internal/provider/cisco/nxos/feat.go#L14-L25.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2678,6 +2679,197 @@ func separateFeatureActivation(conf []gnmiext.Configurable) (features, others []
...
+		} else {
+			acct.Realm = MapRealmFromMethodType(methods[0].Type, "")
+		}
+		conf = append(conf, acct)
+	}
+
+	return p.Patch(ctx, conf...)
+}
+
+func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest) error {
```
felix-kaestner: Please batch all the updates into a single call, similar to the `EnsureAAA`.

felix-kaestner (changes requested):
@weneghawi Please have a look at the `openconfig-system:system/aaa` yang model. Only configurations that are part of this or are otherwise commonly found on all vendors (Nokia, Juniper, Arista & Co.) should be part of the `core` api. All Cisco NX-OS specific configuration should be refactored into a vendor specific provider config, see e.g. the `ManagementAccess` resource on how this is done. There is a separate api package for cisco specific CRDs.

api/core/v1alpha1/aaa_types.go
weneghawi: **Done.** Renamed to `VrfName string json:"vrfName,omitempty"`.

internal/provider/cisco/nxos/provider.go
weneghawi: **Done.** `DeleteAAA` now batches all resets (auth, console auth, authorization, accounting) into a single `p.Update()` call, followed by group/server deletions.

config/samples/networking_v1alpha1_aaa.yaml
```diff
@@ -0,0 +1,102 @@
...
+# login error messages, console authentication, and config-commands authorization.
+apiVersion: nx.cisco.networking.metal.ironcore.dev/v1alpha1
+kind: AAAConfig
+metadata:
+  name: aaa-tacacs-nxos
+  namespace: default
+spec:
+  keyEncryption: Type7
+  loginErrorEnable: true
+  # NX-OS: aaa authentication login console group GR_TACACS local
```
felix-kaestner: Please leave out the cli commands: https://github.com/ironcore-dev/network-operator/pull/176#discussion_r2773478238.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2791,6 +2792,260 @@ func separateFeatureActivation(conf []gnmiext.Configurable) (features, others []
...
+	if hasTACACS {
+		tacacsFeature := TACACSFeatureDisabled
+		conf = append(conf, &tacacsFeature)
+	}
+
+	if len(conf) > 0 {
+		return p.Update(ctx, conf...)
+	}
+
+	return nil
```
```suggestion
	return p.Update(ctx, conf...)
```

the underlying call will already take care of that.

felix-kaestner (changes requested):

internal/provider/cisco/nxos/aaa.go
```diff
@@ -0,0 +1,301 @@
...
+func MapFallbackFromMethodList(methods []v1alpha1.AAAMethod) string {
+	// If there's more than one method, enable fallback
+	if len(methods) > 1 {
+		return AAAValueYes
+	}
+	return AAAValueNo
+}
+
+// MapNXOSRealm maps an NX-OS method type string to NX-OS realm.
+func MapNXOSRealm(methodType string) string {
```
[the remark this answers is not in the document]
felix-kaestner: This was not consistently done. I still see `NXOSMethodList`, `NXOSMethody`, ...

api/cisco/nx/v1alpha1/aaaconfig_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// Methods is the ordered list of methods.
+	// +required
+	// +listType=atomic
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=4
+	Methods []NXOSMethod `json:"methods"`
+}
+
+// NXOSMethod represents a single AAA method in an NX-OS context.
+type NXOSMethod struct {
```
felix-kaestner: Please reuse the `AAAMethod` type from the `core/v1alpha1` package.

cmd/main.go
```diff
@@ -565,6 +565,18 @@ func main() {
 		os.Exit(1)
 	}
 
+	if err := (&corecontroller.AAAReconciler{
+		Client:           mgr.GetClient(),
+		Scheme:           mgr.GetScheme(),
+		Recorder:         mgr.GetEventRecorderFor("aaa-controller"),
```
felix-kaestner:
```suggestion
		Recorder:         		Recorder:         mgr.GetEventRecorder("aaa-controller"),
```

internal/controller/core/aaa_controller.go
```diff
@@ -0,0 +1,357 @@
...
+type AAAReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	// WatchFilterValue is the label value used to filter events prior to reconciliation.
+	WatchFilterValue string
+
+	// Recorder is used to record events for the controller.
+	// More info: https://book.kubebuilder.io/reference/raising-events
+	Recorder record.EventRecorder
```
felix-kaestner:
```suggestion
	Recorder events.EventRecorder
```

internal/controller/core/aaa_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	Provider provider.ProviderFunc
+
+	// Locker is used to synchronize operations on resources targeting the same device.
+	Locker *resourcelock.ResourceLocker
+}
+
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=aaa,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=aaa/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=aaa/finalizers,verbs=update
+// +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```
felix-kaestner:
```suggestion
// +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
```

internal/controller/core/aaa_controller.go
```diff
@@ -0,0 +1,357 @@
...
+// move the current state of the cluster closer to the desired state.
+//
+// For more details, check Reconcile and its Result here:
+// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.20.2/pkg/reconcile
+//
+// For more details about the method shape, read up here:
+// - https://ahmet.im/blog/controller-pitfalls/#reconcile-method-shape
+func (r *AAAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Result, reterr error) {
+	log := ctrl.LoggerFrom(ctx)
+	log.Info("Reconciling resource")
```
felix-kaestner: Please adapt this controller to the logging scheme introduced in https://github.com/ironcore-dev/network-operator/commit/4e1b59c282d0b11e6d7c81ce687a87d074b5a4ea.

internal/controller/core/aaa_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	if isPaused, requeue, err := paused.EnsureCondition(ctx, r.Client, device, obj); isPaused || requeue || err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if err := r.Locker.AcquireLock(ctx, device.Name, "aaa-controller"); err != nil {
+		if errors.Is(err, resourcelock.ErrLockAlreadyHeld) {
+			log.Info("Device is already locked, requeuing reconciliation")
+			return ctrl.Result{RequeueAfter: time.Second * 5}, nil
```
felix-kaestner:
```suggestion
			return ctrl.Result{RequeueAfter: Jitter(time.Second), Priority: new(LockWaitPriorityDefault)}, nil
```
see https://github.com/ironcore-dev/network-operator/commit/10af9d77b6133298240faa2612c02d93eb905612.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2961,6 +2963,195 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+		switch group.Type {
+		case v1alpha1.AAAServerGroupTypeTACACS:
+			if err := p.client.Delete(ctx, &TacacsPlusProviderGroup{Name: group.Name}); err != nil {
+				return err
+			}
+			for _, server := range group.Servers {
+				if err := p.client.Delete(ctx, &TacacsPlusProvider{Name: server.Address}); err != nil {
+					return err
+				}
+			}
```
felix-kaestner: Can we do a single delete call here with all the servers, instead of doing so many individual gnmi requests?
Also you shouldn't assume that the spec contains all servers, that were ever created. So this `DeleteAAA` func should also take care of removing any potential leftover from previous reconciles.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2961,6 +2963,195 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+		case v1alpha1.AAAServerGroupTypeTACACS:
+			tacacsFeature := TACACSFeature(AdminStEnabled)
+			conf = append(conf, &tacacsFeature)
+
+			for _, server := range group.Servers {
+				srv := &TacacsPlusProvider{
+					Name:   server.Address,
+					KeyEnc: MapKeyEncryption(cfg.Spec.KeyEncryption),
+				}
+				if server.TACACS != nil {
```
felix-kaestner: We still append the server to the updates later on despite server.TACACS == nil? This doesn't seem right.
Isn't it enforced that when the group is tacacs, all servers have to have that setting? Making this check here redundant?

Same also applies to the RADIUS servers.

internal/provider/cisco/nxos/aaa.go
felix-kaestner, line 1: Please also add a test file, as done with the other configuration items including a golden cli script under the `testdata/` directory, i.e. `testdata/aaa.json.txt` and the expected YANG json payload of that as `testdata/aaa.json`.

See how this is done on other items, e.g. the `user.go`/`user_test.go`/`user.json`/`user.json.txt`.

felix-kaestner (changes requested):

api/cisco/nx/v1alpha1/aaaconfig_types.go
weneghawi: there's no `NXOSMethod` struct. `NXOSMethodList.Methods` is `[]v1alpha1.AAAMethod` (reusing the core type)

cmd/main.go
weneghawi: Already `GetEventRecorder`

internal/controller/core/aaa_controller.go
weneghawi: Already `events.EventRecorder`

internal/controller/core/aaa_controller.go
weneghawi: Already `events.k8s.io`

internal/controller/core/aaa_controller.go
weneghawi: Already fully on the V(3)/V(2)/V(1) schem

internal/controller/core/aaa_controller.go
weneghawi: Already using `Jitter` and `LockWaitPriorityDefault`.

internal/provider/cisco/nxos/provider.go
weneghawi: reads from device (not spec), single `Delete` call with all entries, groups before servers

internal/provider/cisco/nxos/aaa.go
weneghawi: `aaa_test.go` and 6 pairs of testdata files, all passing

internal/provider/cisco/nxos/aaa.go
```diff
@@ -0,0 +1,301 @@
...
+	_ gnmiext.Configurable = (*AAADefaultAuthor)(nil)
+	_ gnmiext.Configurable = (*AAADefaultAcc)(nil)
+)
+
+// TACACSFeature enables/disables the TACACS+ feature on NX-OS.
+type TACACSFeature string
+
+func (*TACACSFeature) XPath() string {
+	return "System/fm-items/tacacsplus-items/adminSt"
+}
```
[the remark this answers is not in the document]
felix-kaestner: @weneghawi this is not done.

api/cisco/nx/v1alpha1/aaaconfig_types.go
felix-kaestner: @weneghawi Please remove the 'NXOS' prefix in the `NXOSMethodList`, I've already said this in multiple comments.

internal/provider/cisco/nxos/provider.go
felix-kaestner: Can't we just delete the whole `System/userext-items/tacacsext-items/tacacsplusprovider-items/TacacsPlusProvider-list` (and equally the same for RADIUS)?

config/default/kustomization.yaml
```diff
@@ -24,7 +24,7 @@ resources:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 - ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
```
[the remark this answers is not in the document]
felix-kaestner: Again, this is not reverted @weneghawi

internal/provider/cisco/nxos/aaa.go
```diff
@@ -0,0 +1,336 @@
...
+
+// TacacsPlusProvider represents a TACACS+ server host configuration.
+type TacacsPlusProvider struct {
+	Name         string `json:"name"`
+	Port         int32  `json:"port,omitempty"`
+	Key          string `json:"key,omitempty"`
+	KeyEnc       string `json:"keyEnc,omitempty"`
+	Timeout      int32  `json:"timeout,omitempty"`
+	Retries      int32  `json:"retries,omitempty"`
+	AuthProtocol string `json:"authProtocol,omitempty"`
```
felix-kaestner: When using `omitempty`, please be sure that this value is really not present by default in the YANG payload. Otherwise, this can cause the diff of `gnmiext` to fail and thus trigger a gnmi call on every reconcilation. So when these fields have a default value, the `omitempty` should be omitted. Please check the YANG model for this.

api/cisco/nx/v1alpha1/aaaconfig_types.go
weneghawi: Done removed `NXOSMethodList` to `AAAMethodList`

internal/provider/cisco/nxos/provider.go
weneghawi: Done!

docs/architecture-overview.md
```diff
@@ -16,7 +16,7 @@ sequenceDiagram
 	participant CFG as Config Reconciler
 	participant BOOT as Provisioning HTTP / TFTP
 	participant P as Provider / Transport
-	participant DEVICE as Network Device
+	participant ND as Network Device
```
felix-kaestner: Why did you revert these changes? Previously there was no need for the exclusion in `.typos.toml`.

.typos.toml
```diff
@@ -17,6 +17,8 @@ mininum = "mininum"
 specifc = "specifc"
 # Go built-in keyword
 cpy = "cpy"
+# Documentation
```
felix-kaestner: Please revert with the changes done in the architecture docs.

config/samples/v1alpha1_aaa.yaml
felix-kaestner, line 1: This file should just be called `v1alpha1_aaa.yaml`. For consistency with the other files. Please also add this sample to the Tiltfile.

Also, please add this to the `config/samples/kustomization.yaml`

felix-kaestner (changes requested):
The `config/rbac/kustomization.yaml` file is completely missing the entries for the rbac helper that kubebuilder creates when running `kubebuilder create api`, i.e.

```diff
diff --git a/config/rbac/kustomization.yaml b/config/rbac/kustomization.yaml
index f6aab1b9..d0124924 100644
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -22,6 +22,9 @@ resources:
 # default, aiding admins in cluster management. Those roles are
 # not used by the network-operator itself. You can comment the following lines
 # if you do not want those helpers be installed with your Project.
+- aaa_admin_role.yaml
+- aaa_editor_role.yaml
+- aaa_viewer_role.yaml
 - accesscontrollist_admin_role.yaml
 - accesscontrollist_editor_role.yaml
 - accesscontrollist_viewer_role.yaml
```

These files are missing:
- `config/rbac/aaa_admin_role.yaml`
- `config/rbac/aaa_editor_role.yaml`
- `config/rbac/aaa_viewer_role.yaml`

Same applies to the `PROJECT` file which was not updated.
```diff
diff --git a/PROJECT b/PROJECT
index 24face92..d8001d68 100644
--- a/PROJECT
+++ b/PROJECT
@@ -289,4 +289,12 @@ resources:
   kind: DHCPRelay
   path: github.com/ironcore-dev/network-operator/api/core/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: networking.metal.ironcore.dev
+  kind: AAA
+  path: github.com/ironcore-dev/network-operator/api/core/v1alpha1
+  version: v1alpha1
 version: "3"
```

@weneghawi did you even generate the files by running `kubebuilder create`?

config/samples/networking_v1alpha1_aaa.yaml
```diff
@@ -0,0 +1,92 @@
...
+  namespace: default
+type: Opaque
+stringData:
+  # Replace with your actual TACACS+ server shared secret
+  server-key: "supersecretkey"
+---
+# Cisco NX-OS specific AAA configuration
+# This configures vendor-specific settings like key encryption type,
+# login error messages, console authentication, and config-commands authorization.
+apiVersion: nx.cisco.networking.metal.ironcore.dev/v1alpha1
```
felix-kaestner: Ideally, this would be moved to a `cisco/nx/v1alpha1_aaaconfig.yaml` file, as done in the other provider configs, so that our samples stay vendor-independent.

internal/provider/cisco/nxos/provider.go
weneghawi: We tried this but NX-OS rolls back the entire gNMI SET batch silently if any item fails validation. Including adminSt disabled in the same batch as the auth resets causes NX-OS to reject the whole request because it refuses to disable the feature while groups still exist, and since the batch is atomic the auth resets are reverted too.

The 3 call split is required by NX-OS teardown ordering: reset auth, delete groups and servers, then disable feature. This is not visible when testing with gnmic directly since each CLI invocation is its own gNMI SET RPC.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2678,6 +2679,197 @@ func separateFeatureActivation(conf []gnmiext.Configurable) (features, others []
...
+	var conf []gnmiext.Configurable
+
+	// Enable TACACS+ feature if there are TACACS servers configured
+	if len(req.AAA.Spec.TACACSServers) > 0 {
+		tacacsFeature := TACACSFeatureEnabled
+		conf = append(conf, &tacacsFeature)
+	}
+
+	// Configure TACACS+ server hosts
+	for _, server := range req.AAA.Spec.TACACSServers {
```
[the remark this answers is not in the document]
weneghawi: Done. EnsureAAA now reads the current TACACS and RADIUS server entries from the device after applying the desired config and deletes any that are no longer in the spec. The cleanup runs after the group update so NX-OS does not reject the delete due to the server still being referenced by a group.

config/samples/networking_v1alpha1_aaa.yaml
weneghawi: Done. Removed the CLI command references from the sample header comment.

docs/architecture-overview.md
weneghawi: The architecture docs and typos.toml are already correct on this branch. The docs use the ND shorthand (from commit 9be8b07) and the nd exception in typos.toml (from commit 7aeea49) is already present. No change needed here.

.typos.toml
weneghawi: Same as above, both changes are already in place on this branch.

config/samples/v1alpha1_aaa.yaml
weneghawi: Done. Renamed to `v1alpha1_aaa.yaml`, added to `config/samples/kustomization.yaml`, and added to the Tiltfile.

config/samples/networking_v1alpha1_aaa.yaml
weneghawi: Done. Moved to `config/samples/cisco/nx/v1alpha1_aaaconfig.yaml`. The core `v1alpha1_aaa.yaml` now only contains the vendor-agnostic AAA resource.

internal/provider/cisco/nxos/provider.go
weneghawi: Deleting the whole list after p.Update would also remove the desired servers we just created, requiring a second Update call. Deleting it before p.Update fails because the group still references the old servers. The current read+diff approach uses a single GetConfig and a single batched Delete after the group update — NX-OS requires the server to be dropped from the group's ProviderRef-list first before its host entry can be removed.

internal/provider/cisco/nxos/provider.go
weneghawi: Correct, the XValidation rule on AAAServerGroup enforces that all servers in a TACACS group must have the tacacs field set, so no nil check is needed. The code accesses server.TACACS.Port directly and relies on the API validation to prevent nil.

weneghawi:
**Done.** Added the missing RBAC helper files:
- `config/rbac/aaa_admin_role.yaml`
- `config/rbac/aaa_editor_role.yaml`
- `config/rbac/aaa_viewer_role.yaml`
- `config/rbac/cisco/nx/aaaconfig_admin_role.yaml`
- `config/rbac/cisco/nx/aaaconfig_editor_role.yaml`
- `config/rbac/cisco/nx/aaaconfig_viewer_role.yaml`

All six are now referenced in `config/rbac/kustomization.yaml`. The `PROJECT` file already had the AAA entry (the branch was scaffolded with kubebuilder, the RBAC files just weren't committed).

internal/provider/cisco/nxos/aaa.go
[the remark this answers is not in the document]
weneghawi: Done. The Feature type from feat.go is used directly — see provider.go where we use `&Feature{Name: "tacacsplus", AdminSt: AdminStEnabled}` and `AdminStDisabled`. No separate feature struct exists in aaa.go.

api/cisco/nx/v1alpha1/aaaconfig_types.go
weneghawi: Done. The type is now `AAAMethodList` (no NXOS prefix) in api/cisco/nx/v1alpha1/aaaconfig_types.go.

internal/provider/cisco/nxos/provider.go
weneghawi: The whole-list delete would remove the desired servers we just created, since desired and existing servers share the same path prefix. Deleting before p.Update fails because NX-OS rejects removing a server that is still referenced by a group. The read+diff+delete approach is the minimum required by NX-OS ordering constraints.

config/default/kustomization.yaml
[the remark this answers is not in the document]
weneghawi: Done. config/default/kustomization.yaml has no diff against main — the E2E test change was already reverted.

internal/provider/cisco/nxos/aaa.go
weneghawi: Checked. All fields with `omitempty` fall into one of two categories: (1) fields we always explicitly set to non-empty values when building the struct (`realm`, `fallback`, `local`, `none` — always set via MapFallback/MapLocal/MapRealm), or (2) genuinely optional fields where the empty/zero value is the correct device default (`vrf`, `srcIf`, `key`, `deadtime`). The `authProtocol` field was unused and has been removed from the struct.

weneghawi:
## How to test

To test locally against a real NX-OS device:

**1. Set up SSH tunnel to a QA switch (port-forward gNMI)**
```bash
ssh -f -N -L 9339:<device-ip>:9339 <jump-host>
```

**2. Apply the sample resources** (vendor-agnostic AAA + Cisco-specific AAAConfig)
```bash
kubectl apply -f config/samples/v1alpha1_aaa.yaml
kubectl apply -f config/samples/cisco/nx/v1alpha1_aaaconfig.yaml
```

Adjust `spec.deviceRef.name`, the server addresses, and the key secret to match your environment.

**3. Watch the operator reconcile and verify on the device**
```bash
gnmic -a localhost:9339 -u <user> -p '<pass>' --skip-verify   get --path 'System/fm-items/tacacsplus-items'       --path 'System/userext-items/tacacsext-items'       --path 'System/userext-items/authrealm-items/defaultauth-items'
```

**4. Test stale server cleanup** — remove a server from `spec.serverGroups[].servers` and reconcile; the operator will remove it from the group's ProviderRef-list and then delete the host entry from the device.

**5. Test delete** — delete the AAA CR; the operator resets auth to local, removes all groups and servers, then disables the TACACS feature.

---

Regarding the failing CI check: the `VPCDomain Controller` test failure in `vpcdomain_controller_test.go` is pre-existing and unrelated to this PR — the same tests time out on `main` today (run [24726816994](https://github.com/ironcore-dev/network-operator/actions/runs/24726816994)). This is the known NX-OS gNMI VPCDomain issue tracked in #282.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+	if req.ProviderConfig != nil {
+		if err := req.ProviderConfig.Into(&cfg); err != nil {
+			return err
+		}
+	}
+
+	// Collect desired server addresses per protocol to detect stale entries.
+	desiredTACACS := map[string]bool{}
+	desiredRADIUS := map[string]bool{}
+
```
felix-kaestner, line 3490:
```suggestion
	desiredTACACS := map[string]struct{}
	desiredRADIUS := map[string]struct{}
```
set like data structure in go

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+				SrcIf: group.SourceInterfaceName,
+			}
+			for _, server := range group.Servers {
+				grp.ProviderRefItems.ProviderRefList.Set(&RadiusProviderRef{Name: server.Address})
+			}
+			conf = append(conf, grp)
+		}
+	}
+
+	if req.AAA.Spec.Authentication != nil && len(req.AAA.Spec.Authentication.Methods) > 0 {
```
felix-kaestner, line 3569: When `.spec.Authentication` is `nil` the configuration is not reset to the default value but kept as it is.

I think we should take full ownership of the configuration and the logic should be:
1. If not specified in the spec → set to the default value of the device
2. If specified in the spec → configure as intended.

The same applies to all following configuration blocks.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+	}
+
+	if err := p.Update(ctx, conf...); err != nil {
+		return err
+	}
+
+	// Remove TACACS+ server host entries that are on the device but no longer in the spec.
+	// This must happen after the Update above so the server is first dropped from the group's
+	// ProviderRef-list — NX-OS rejects deleting a server that is still referenced by a group.
+	if len(desiredTACACS) > 0 {
```
felix-kaestner: Similar to my other comment, we should always make sure to clean up any stale entries. Not just in the case where there are entries configured. E.g. when someone only configures TACACS+ servers in the spec, we wouldn't clean up stale RADIUS server, which we should. So I would say the `len(desiredTACACS) > 0` should be removed.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+		if err := p.client.GetConfig(ctx, current); err != nil && !errors.Is(err, gnmiext.ErrNil) {
+			return err
+		}
+		var stale []gnmiext.DataElement
+		for i := range current.ProviderList {
+			if !desiredTACACS[current.ProviderList[i].Name] {
+				stale = append(stale, &TacacsPlusProvider{Name: current.ProviderList[i].Name})
+			}
+		}
+		if len(stale) > 0 {
```
```suggestion
```
The length check is not needed, the `p.client.Delete` is already capable of handling then empty list of deletions.

Same applies below.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+	}
+
+	// Remove TACACS+ server host entries that are on the device but no longer in the spec.
+	// This must happen after the Update above so the server is first dropped from the group's
+	// ProviderRef-list — NX-OS rejects deleting a server that is still referenced by a group.
+	if len(desiredTACACS) > 0 {
+		current := new(TacacsPlusProviderItems)
+		if err := p.client.GetConfig(ctx, current); err != nil && !errors.Is(err, gnmiext.ErrNil) {
+			return err
+		}
```
felix-kaestner: Why do we fetch the list here again, we just configured with based the `desiredTACACS`, so there should be no need in making another get request to the device.

internal/provider/cisco/nxos/provider.go
```diff
@@ -2997,6 +2999,239 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+			toDelete = append(toDelete, new(TacacsPlusProviderGroupItems), new(TacacsPlusProviderItems))
+		case v1alpha1.AAAServerGroupTypeRADIUS:
+			toDelete = append(toDelete, new(RadiusProviderGroupItems), new(RadiusProviderItems))
+		}
+	}
+	if len(toDelete) > 0 {
+		if err := p.client.Delete(ctx, toDelete...); err != nil && !errors.Is(err, gnmiext.ErrNil) {
+			return err
+		}
+	}
```
```suggestion
		if err := p.client.Delete(ctx, new(TacacsPlusProviderGroupItems), new(TacacsPlusProviderItems), new(RadiusProviderGroupItems), new(RadiusProviderItems)); err != nil && !errors.Is(err, gnmiext.ErrNil) {
			return err
		}
```

Just always (unconditionally) remove all TACACS+ and RADIUS servers. Everything should be reset to the default, do not consider the spec.

If you put this as the first step, you can also bundle the rest into a single `.Update` afterwards.

docs/architecture-overview.md
weneghawi: Done. Reverted —  is back to using `DEVICE` and the `nd = "nd"` exclusion has been removed from `.typos.toml`.

.typos.toml
weneghawi: Done. Removed the `nd = "nd"` exclusion.

config/samples/v1alpha1_aaa.yaml
weneghawi: Done. The file is now `config/samples/v1alpha1_aaa.yaml`, added to both `config/samples/kustomization.yaml` and the Tiltfile.

config/samples/networking_v1alpha1_aaa.yaml
weneghawi: Done. The Cisco-specific config (keyEncryption, loginErrorEnable, consoleAuthentication, configCommandsAuthorization) is now in `config/samples/cisco/nx/v1alpha1_aaaconfig.yaml` and the core `v1alpha1_aaa.yaml` sample is vendor-agnostic.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. Changed to `map[string]struct{}{}` for both.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. All four auth blocks (defaultauth, consoleauth, defaultauthor, defaultacc) now always emit a config element. When the corresponding spec field is nil, the device default (local realm, fallback yes) is written — taking full ownership of the configuration.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. Removed both `len(desiredTACACS) > 0` and `len(desiredRADIUS) > 0` guards — stale cleanup now always runs for both protocols regardless of what is configured in the spec.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. Removed both inner `len(stale) > 0` checks — `p.client.Delete` with an empty slice is a no-op.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. The GET for current server lists is now done **before** the `Update` call. We fetch `TacacsPlusProviderItems` and `RadiusProviderItems` upfront to compute the stale set, then apply the desired state via `Update`, and finally delete the stale entries in a single combined call. No fetch after configuring.

internal/provider/cisco/nxos/provider.go
weneghawi: Done. `DeleteAAA` now unconditionally deletes all four containers (`TacacsPlusProviderGroupItems`, `TacacsPlusProviderItems`, `RadiusProviderGroupItems`, `RadiusProviderItems`) in a single call regardless of the spec. We kept the auth reset as a prior step because NX-OS can leave auth in a broken state if the referenced provider group is removed while still active — so we clear the realm references first, then delete the groups and servers, then disable the feature.

charts/network-operator/templates/network-policy/allow-metrics-traffic.yaml
felix-kaestner, line 1: Why was this deleted?

charts/network-operator/Chart.yaml
```diff
@@ -1,5 +1,14 @@
 apiVersion: v2
 name: network-operator
-description: A Helm chart to distribute the project network-operator
+description: A Helm chart to distribute network-operator
 type: application
-version: 0.0.0+dev
+
+version: 0.1.0
+appVersion: "0.1.0"
+
```
felix-kaestner: Please don't modify the versioning. It was done on purpose. In general, I would revert the changes on the `Chart.yaml` as they are unrelated to this PRs intention.

charts/network-operator/values.yaml
```diff
@@ -17,7 +17,9 @@ manager:
 
   image:
     repository: ghcr.io/ironcore-dev/network-operator
-    tag: latest
+    ## Image tag (defaults to Chart.appVersion if not set)
```
felix-kaestner: Please revert the changes.

charts/network-operator/templates/network-policy/allow-metrics-traffic.yaml
weneghawi: These files were unintentionally deleted by the `make charts` run. The templates are gated on `{{- if .Values.networkPolicy.enable }}`, and the chart regeneration also removed the `networkPolicy` section from `values.yaml`, which caused the template files to be dropped as dead code. Restored in the latest commit along with the `networkPolicy` section.

charts/network-operator/Chart.yaml
weneghawi: Reverted. The versioning was changed unintentionally by the `make charts` run — restored to `0.0.0+dev` in the latest commit.

charts/network-operator/values.yaml
weneghawi: Reverted. The `tag` field change was also an unintentional side effect of the `make charts` run — restored to `tag: latest` in the latest commit.

felix-kaestner (approved):

PR #191, opened 2026-02-18
Add peer interface annotation for physical interfaces

felix-kaestner:
This patch introduces the `PhysicalInterfaceNeighborLabel` label which
references the peer Interface resource on the other end of a physical
link. The label can be used by users to declare expected cabling, or
by controllers to reflect actual cabling discovered via neighbor
discovery protocols (e.g., LLDP).

The label value must be the name of another Interface resource in the
same namespace. It is only valid for interfaces of type Physical.

Additionally, a `PhysicalInterfaceNeighborRawAnnotation` annotation is
introduced for interfaces connected to unmanaged devices (devices
without an Interface resource). The annotation value format is
"chassisID::portID" where chassisID is the LLDP chassis identifier
and portID is the LLDP port identifier.

Both the label and annotation are mutually exclusive - an interface
cannot have both set at the same time.

### DIFF
--- a/api/core/v1alpha1/groupversion_info.go
+++ b/api/core/v1alpha1/groupversion_info.go
@@ -63,6 +63,22 @@ const VRFLabel = "networking.metal.ironcore.dev/vrf-name"
 // to trigger certain disruptive operations, such as reboots or firmware upgrades.
 const DeviceMaintenanceAnnotation = "networking.metal.ironcore.dev/maintenance"
 
+// PhysicalInterfaceNeighborLabel identifies the peer Interface resource on the other end of a physical link.
+// The value must be the name of another Interface resource in the same namespace.
+// This label is only valid for interfaces of type Physical.
+// For peers that are not managed as Interface resources, use PhysicalInterfaceNeighborRawAnnotation instead.
+const PhysicalInterfaceNeighborLabel = "networking.metal.ironcore.dev/interface-neighbor"
+
+// PhysicalInterfaceNeighborRawAnnotation stores raw neighbor identification for interfaces
+// connected to unmanaged devices (devices without an Interface resource).
+// The value format is "chassisID::portID" where:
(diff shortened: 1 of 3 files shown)


api/core/v1alpha1/groupversion_info.go
```diff
@@ -63,6 +63,12 @@ const VRFLabel = "networking.metal.ironcore.dev/vrf-name"
 // to trigger certain disruptive operations, such as reboots or firmware upgrades.
 const DeviceMaintenanceAnnotation = "networking.metal.ironcore.dev/maintenance"
 
+// InterfacePeerAnnotation references the peer Interface resource on the other end of a physical link.
+// The value must be a reference to another Interface resource in the format "namespace/name"
+// or just "name" (for same-namespace references).
+// This annotation is only valid for interfaces of type Physical.
+const InterfacePeerAnnotation = "networking.metal.ironcore.dev/peer-interface"
```
nikatza: I think aggregated interfaces should be allowed too: "An An LLDP agent can be configured on an IEEE 802.1AX™ Aggregated Port and/or on any number of the physical ports belonging to the Aggregation" [IEEE8021AB-2016]. May be we could leave a note here or extend the commit message if we want to enforce this condition for now.

SchwarzM:
do we want to use an annotation in this case ? this makes it harder for cli tools to display / filter for it correctly I think

api/core/v1alpha1/groupversion_info.go
nikatza: Fair point. We also discussed it in our alignment and we won't consider link aggregation. In this context, what do you think about renaming the `Annotation` to `PhysicalInterfaceNeighborAnnotation = "networking.metal.ironcore.dev/interface-neighbor"`. This would also be more consistent with the `LLDP` terminology.

nikatza (approved):
Nice, thanks for the update!

PR #198, opened 2026-03-03
Change NVE multicast groups to accept CIDR format

nikatza:
Accept multicast network prefixes (e.g., 239.1.0.0/16) instead of host addresses. The webhook validates no host bits are set, and the provider extracts the base IP before sending to the device.


api/core/v1alpha1/nve_types.go
```diff
@@ -72,12 +72,14 @@ const (
 type MulticastGroups struct {
 	// L2 is the multicast group for Layer 2 VNIs (BUM traffic in bridged VLANs).
 	// +optional
-	// +kubebuilder:validation:Format=ipv4
+	// +kubebuilder:validation:Format=cidr
+	// +kubebuilder:validation:Pattern=`^([0-9]{1,3}\.){3}[0-9]{1,3}/[0-9]{1,2}$`
 	L2 string `json:"l2,omitempty"`
```
felix-kaestner: Given that this becomes a Prefix, could you also change the type here to the `*IPPrefix` type? It will then already be unmarshaled into a `netip.Prefix` saving a  lot of the complexity to do this manually from a string.

felix-kaestner (approved):
@nikatza Can you please rebase and regenerate the helm chart with `make charts` to make the adjustments also reflected there?

PR #209, opened 2026-03-05
Support `LLDP` configuration

nikatza:


charts/network-operator/templates/rbac/manager-role.yaml
felix-kaestner: Are there no rbac files generated for the `LLDPConfig`.

felix-kaestner (approved):
Very nice!

PR #227, opened 2026-03-11
Add L3 port-channel support with member interface layer management

felix-kaestner:
Enable IPv4, VRF membership, and BFD on Aggregate (port-channel) interfaces in the CRD API and NX-OS provider. Handle the member interface layer problem where NX-OS requires Physical members of an L3 port-channel to be Layer3 even without their own IP address.

- Relax CEL validation to allow L3 config on Aggregate interfaces
- Add AggregateParent to EnsureInterfaceRequest for member awareness
- Set Layer3/VRF/medium on Aggregate when IPv4 is configured (NX-OS)
- Set Layer3 on Physical members of L3 Aggregates (NX-OS)
- Add aggregateToMembers watch with field-scoped update predicate
- Add golden files, provider fixtures, and controller integration tests
- Add L3 port-channel sample manifests and Tiltfile entries


nikatza (approved):
Just some minor nits. This is great, thanks! (only tests are not passing)

PR #243, opened 2026-03-15
[NXOS] Fix interface reconciliation loop

nikatza:


felix-kaestner (approved):
Thanks for this fix!

Could you add a little description when you merge why this change was done, i.e. that we currently update the configuration in YANG on every reconcile as we missed the default value of this field?

PR #258, opened 2026-03-25
Add pool/claim types and controllers for numbered resource allocation

felix-kaestner:
Introduces a pool-based allocation system for numbered resources (ASNs, VLAN IDs, IP addresses, prefixes). Three pool types — `IndexPool`, `IPAddressPool`, `IPPrefixPool` — define the available value ranges. A `Claim` references a pool via `spec.poolRef` and the controller allocates the next available value by creating a dedicated allocation object (`Index`, `IPAddress`, or `IPPrefix`).

Allocation objects are first-class Kubernetes resources with a deterministic name derived from the pool name and value (e.g. `asn-pool-64512`). This naming scheme uses the API server's name-uniqueness guarantee as the concurrency guard: two controllers racing for the same value both attempt `Create`, exactly one succeeds, and the other retries with a fresh list. Each allocation object carries a `spec.claimRef` binding it to the owning claim and a `spec.poolRef` back to the pool.

Pools support `Recycle` (default) and `Retain` reclaim policies. On claim deletion, `Recycle` deletes the allocation object to free the value. `Retain` clears the `claimRef` so the allocation persists as reserved but unbound — it can be rebound later using the `allow-binding` annotation. Owner references from the pool to both claims and allocation objects enable automatic garbage collection when a pool is deleted.

Pool types implement a `Pool` interface with `Allocate` and `ListAllocations` methods. Allocation types implement an `Allocation` interface with `GetClaimRef`, `SetClaimRef`, and `GetValue`. The claim controller operates generically through these interfaces with no type-specific logic beyond a single kind switch for pool instantiation.

### Concept

numbered-resources

### Usage

```zsh
$ kubectl get idxpool -o wide
NAME               ALLOCATED   TOTAL      AVAILABLE   AGE
indexpool-sample   1           94968318   True        9s

$ kubectl get ippool -o wide
NAME                   ALLOCATED   TOTAL      AVAILABLE   AGE
ipaddresspool-sample   1           17891328   True        12s

$ kubectl get pfxpool -o wide
NAME                  ALLOCATED   TOTAL   AVAILABLE   AGE
ipprefixpool-sample   1           65536   True        16s

$ kubectl get claim -o wide
NAME              VALUE         ALLOCATED   AGE
claim-index       64512         True        19s
claim-ipaddress   10.0.0.0      True        19s
claim-prefix      10.0.0.0/24   True        18s

$ kubectl get idx -o wide
NAME                     POOL               INDEX   CLAIM         VALID   AGE
indexpool-sample-64512   indexpool-sample   64512   claim-index   True    23s
```


api/core/v1alpha1/index_range.go
```diff
@@ -0,0 +1,92 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
nikatza: nit: `2025` -> `2026` (other new files are also affected)

api/core/v1alpha1/index_range.go
```diff
@@ -0,0 +1,92 @@
...
+	"strings"
+)
+
+// IndexRange represents an inclusive range of indices.
+// +kubebuilder:validation:Type=string
+// +kubebuilder:validation:Pattern=`^[0-9]+\.\.[0-9]+$`
+// +kubebuilder:object:generate=false
+type IndexRange struct {
+	Start uint64 `json:"-"`
+	End   uint64 `json:"-"`
```
nikatza: The [API conventions document](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#primitive-types) does not seem to apply here as this is then later serialized as a string in `IndexRange`. However, this is later used in  `Index.Spec.Index`, which is an `int64`. In theory we could run into an overflow I think. In fact, the `nosec` directive `indexpool_types.go:155` is (mis?)handling this issue. May be I am missing part of the puzzle: Is my understanding here correct?

api/pool/v1alpha1/groupversion_info.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// PreferredValueUnavailableReason indicates that the requested preferred value is not available.
+	PreferredValueUnavailableReason = "PreferredValueUnavailable"
+)
+
+// Annotation keys
+const (
+	// PreferredValueAnnotation is an optional annotation on a Claim that requests a specific
+	// allocation value. The format depends on the pool type:
+	//   - IndexPool: decimal uint64, e.g. "64512"
```
nikatza: If the comments from earlier are valid then this should be `int64`

api/pool/v1alpha1/claim_types.go
```diff
@@ -0,0 +1,131 @@
...
+	// ListAllocations lists allocation objects matching the given options.
+	ListAllocations(ctx context.Context, c client.Client, opts ...client.ListOption) ([]Allocation, error)
+
+	// Allocate picks the next free value from the pool's ranges/prefixes,
+	// given the existing allocations for this pool, and returns a new Allocation
+	// with a ready-to-create object, deterministic name, and the allocated value.
+	// Returns ErrPoolExhausted when no free value remains.
+	Allocate(claim *Claim, existing []Allocation) (Allocation, error)
+}
+type ClaimSpec struct {
```
nikatza, line 53: We need a comment for the function.

api/pool/v1alpha1/claimref_types.go
```diff
@@ -0,0 +1,26 @@
...
+package v1alpha1
+
+import "k8s.io/apimachinery/pkg/types"
+
+const (
+	// AllowBindingAnnotation permits an allocation object whose claimRef name matches
+	// a Claim but whose UID is stale (e.g. after the Claim was deleted and recreated)
+	// to be rebound by updating the UID to the current Claim.
+	AllowBindingAnnotation = "pool.networking.metal.ironcore.dev/allow-binding"
+)
```
nikatza: As a user I would have first looked into `groupversion_info.go`, seems a better home for this. That said, here is also fine.

api/pool/v1alpha1/claimref_types.go
```diff
@@ -0,0 +1,26 @@
...
+// ClaimRef identifies the Claim bound to an allocation object.
+// +structType=atomic
+type ClaimRef struct {
+	// Name is the name of the Claim.
+	// +required
+	Name string `json:"name"`
+
+	// UID is the UID of the Claim, guarding against rebinding to a recycled
+	// Claim with the same name.
+	// +required
```
nikatza, line 20: This is to some extent contradicting with the docs,  they state:
```
When the claim controller finds an allocation whose `claimRef.name` matches but whose UID is
stale (or empty), ...
```

api/pool/v1alpha1/indexpool_types.go
```diff
@@ -0,0 +1,179 @@
...
+	// Example: "64512..65534".
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	Ranges []corev1alpha1.IndexRange `json:"ranges"`
+
+	// ReclaimPolicy controls what happens to an allocation when a claim is deleted.
+	// Recycle returns the allocation to the pool. Retain keeps it reserved.
+	// +optional
+	// +kubebuilder:default=Recycle
+	ReclaimPolicy ReclaimPolicy `json:"reclaimPolicy,omitempty"`
```
nikatza, line 31: I am wondering if we would be safer for now by making this `immutable`. At the moment, one user could change the Policy, which -if I understand correctly- would imply:
- `Retain` to `Recycle`: here existing allocations that would be preserved on claim deletion will be permanently deleted, breaking any business logic that relied on rebinding setup. 

I am aware that it some K8s resources like `PersistentVolumes` allow to [change the Policy at the Volume level](https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/). I am not sure if we have the same requirements at the moment. What are your thoughts on this?

api/pool/v1alpha1/indexpool_types.go
```diff
@@ -0,0 +1,179 @@
...
+	allocs := make([]Allocation, len(list.Items))
+	for i := range list.Items {
+		allocs[i] = &list.Items[i]
+	}
+	return allocs, nil
+}
+
+// Allocate finds the first free index across all ranges and returns an Index
+// allocation object for the given claim.
+func (p *IndexPool) Allocate(claim *Claim, existing []Allocation) (Allocation, error) {
```
nikatza, line 130: I think we should try this approach for now. However, I am wondering if you have an idea about how we could expose the time this takes as a Prometheus query. I guess a good first step would be to have metrics for the reconciliation. That way we would easily get early knowledge if this has performance issues. What do you think?

api/pool/v1alpha1/ipaddresspool_types.go
```diff
@@ -0,0 +1,191 @@
...
+	// Prefixes defines the CIDR ranges that can be allocated.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	Prefixes []corev1alpha1.IPPrefix `json:"prefixes"`
+
+	// ReclaimPolicy controls what happens to an allocation when a claim is deleted.
+	// Recycle returns the allocation to the pool. Retain keeps it reserved.
+	// +optional
+	// +kubebuilder:default=Recycle
+	ReclaimPolicy ReclaimPolicy `json:"reclaimPolicy,omitempty"`
```
nikatza, line 33: Same question as above wrt. the immutability of this field.

api/pool/v1alpha1/ipaddresspool_types.go
```diff
@@ -0,0 +1,191 @@
...
+func (p *IPAddressPool) Allocate(claim *Claim, existing []Allocation) (Allocation, error) {
+	allocated := make(map[netip.Addr]struct{}, len(existing))
+	for _, obj := range existing {
+		allocated[obj.(*IPAddress).Spec.Address.Addr] = struct{}{}
+	}
+	for _, prefix := range p.Spec.Prefixes {
+		masked := prefix.Masked()
+		for addr := masked.Addr(); masked.Contains(addr); addr = addr.Next() {
+			if _, taken := allocated[addr]; !taken {
+				value := addr.String()
```
nikatza, line 154: docstring of `.String()` says: 
`...IPv6 with zone ("fe80:db8::1%eth0"`

the sanitizeValue function would not replace the `%` and we would error. Is my understanding correct?

api/pool/v1alpha1/indexpool_types.go
```diff
@@ -0,0 +1,179 @@
...
+	allocated := make(map[int64]struct{}, len(existing))
+	for _, obj := range existing {
+		idx := obj.(*Index)
+		if idx.Spec.Index >= 0 {
+			allocated[idx.Spec.Index] = struct{}{}
+		}
+	}
+	for _, r := range p.Spec.Ranges {
+		for idx := r.Start; idx <= r.End; idx++ {
```
nikatza, line 139: Corner case: `r.End` equals to `MaxUint64` , then we would end in an infinite loop. May be we need an additional CEL rule to forbid that specific range?

api/pool/v1alpha1/ipprefixpool_types.go
```diff
@@ -0,0 +1,237 @@
...
+
+	corev1alpha1 "github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+)
+
+// IPPrefixPoolSpec defines the desired state of IPPrefixPool
+type IPPrefixPoolSpec struct {
+	// Prefixes defines the base prefixes and target prefix lengths to allocate from.
+	// +required
+	// +kubebuilder:validation:MinItems=1
+	Prefixes []IPPrefixPoolPrefix `json:"prefixes"`
```
nikatza: With this specification an end-user could do something like:
```
prefixes:
  - prefix: "10.0.0.0/28"
    prefixLength: 30   # produces: 10.0.0.0/30, 10.0.0.4/30, 10.0.0.8/30, 10.0.0.12/30
  - prefix: "10.0.0.0/28"
    prefixLength: 29   # produces: 10.0.0.0/29, 10.0.0.8/29
```

if I understand correctly, we potentiall could then have two different claims getting overlapping IP ranges. I understand that this is a user problem. However, I also think this is not straight-forward by looking just at the spec. So, I am wondering if should add something like `.spec.AllowOverlaps` with a default value and a webhook, or if we could just add a comment. 

I am interested to know what would you propose.

docs/concepts/numbered-resources.md
```diff
@@ -0,0 +1,183 @@
...
+
+<img :src="allocationImage" alt="Numbered Resource Allocation Flow" style="margin: 2em 0;" />
+
+## Pool Types
+
+Three pool types are available, each targeting a different kind of resource:
+
+| Pool Kind       | Allocates                   | Allocation Object | Example use case                       |
+| --------------- | --------------------------- | ----------------- | -------------------------------------- |
+| `IndexPool`     | `uint64` Index              | `Index`           | ASN, VLAN ID, VNI, Route Distinguisher |
```
nikatza: This is `int64` in `pool.Index.spec.Index`, isn't it?

api/pool/v1alpha1/indexpool_types.go
```diff
@@ -0,0 +1,179 @@
...
+						Name:      fmt.Sprintf("%s-%s", p.Name, value),
+						Namespace: p.Namespace,
+					},
+					Spec: IndexSpec{
+						PoolRef: corev1alpha1.TypedLocalObjectReference{
+							APIVersion: GroupVersion.String(),
+							Kind:       "IndexPool",
+							Name:       p.Name,
+						},
+						Index: int64(idx), // #nosec G115
```
nikatza: In relation to the my comment in `index_range.go`: This directive might be wrongly applied here. I'm thinking the following: If a user applies this we wold run into an overflow here, don't we?
```
  apiVersion: pool.networking.metal.ironcore.dev/v1alpha1
  kind: IndexPool
  metadata:
    name: my-pool
  spec:
    ranges:
      - "9223372036854775808..9223372036854775810"
```

internal/controller/pool/claim_controller.go
```diff
@@ -0,0 +1,519 @@
...
+		// Watches enqueues Claims for updates in referenced IPPrefixPool resources.
+		// Triggers on create, delete, and update events when the allocated count changes.
+		Watches(
+			&poolv1alpha1.IPPrefixPool{},
+			handler.EnqueueRequestsFromMapFunc(r.claimsForPoolRef),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldPool := e.ObjectOld.(*poolv1alpha1.IPPrefixPool)
+					newPool := e.ObjectNew.(*poolv1alpha1.IPPrefixPool)
+					return oldPool.Status.Allocated != newPool.Status.Allocated
```
nikatza: Should we consider also `.Status.Total` here? If the pool is exhausted, we change the spec and add new ranges. The controller won't trigger a reconciliation and we are anyhow stucked due to the `TerminalError()` from the exhaustion. 

```suggestion
                    return oldPool.Status.Allocated != newPool.Status.Allocated ||
    oldPool.Status.Total != newPool.Status.Total
```

internal/controller/pool/claim_controller.go
```diff
@@ -0,0 +1,519 @@
...
+		// Watches enqueues Claims for updates in referenced IPAddressPool resources.
+		// Triggers on create, delete, and update events when the allocated count changes.
+		Watches(
+			&poolv1alpha1.IPAddressPool{},
+			handler.EnqueueRequestsFromMapFunc(r.claimsForPoolRef),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldPool := e.ObjectOld.(*poolv1alpha1.IPAddressPool)
+					newPool := e.ObjectNew.(*poolv1alpha1.IPAddressPool)
+					return oldPool.Status.Allocated != newPool.Status.Allocated
```
nikatza: Same question as below.

internal/controller/pool/claim_controller.go
```diff
@@ -0,0 +1,519 @@
...
+		// Watches enqueues Claims for updates in referenced IndexPool resources.
+		// Triggers on create, delete, and update events when the allocated count changes.
+		Watches(
+			&poolv1alpha1.IndexPool{},
+			handler.EnqueueRequestsFromMapFunc(r.claimsForPoolRef),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldPool := e.ObjectOld.(*poolv1alpha1.IndexPool)
+					newPool := e.ObjectNew.(*poolv1alpha1.IndexPool)
+					return oldPool.Status.Allocated != newPool.Status.Allocated
```
nikatza: Same question as below.

internal/controller/pool/claim_controller.go
```diff
@@ -0,0 +1,519 @@
...
+		})
+		return reconcile.TerminalError(err)
+	case errors.Is(err, poolv1alpha1.ErrPoolExhausted):
+		conditions.Set(claim, metav1.Condition{
+			Type:    poolv1alpha1.AllocatedCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  poolv1alpha1.PoolExhaustedReason,
+			Message: "Referenced pool is exhausted",
+		})
+		return reconcile.TerminalError(err)
```
nikatza, line 392: Assuming that we can change the pool size in run-time, we might want how we handle this case. If we keep the `TerminalError` then we probably need to modify the watch above to consider the `.Status.Total`. Alternatively, we could also check on the `Available` condition. By this we might ensure a retry if someone frees resources.

internal/controller/pool/claim_controller_test.go
```diff
@@ -0,0 +1,464 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	corev1alpha1 "github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	poolv1alpha1 "github.com/ironcore-dev/network-operator/api/pool/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/conditions"
+)
+
+var _ = Describe("Claim Controller", func() {
```
nikatza, line 19: I was wondering if we should be more exhaustive here and consider the following tests:
- Pool owns the claim and it owns the allocation
- Pool `Available` condition is correctly set
- Claim conditions are correctly set
- Allocation `claimRef` has correct UID
- Re-reconciliation doesn't double allocate
- On conflict retry with the next slot
- Exhausted claims retry once a slot is free
- Exhausted claims retry after pool is expanded
- Behavior when pool is reduced?
- Pool deletion cascades on the claims

nikatza (changes requested):
This is fantastic contribution!! 🚀🚀🚀

Since this is a very large PR I came with a larger number of comments than usual, but they are mostly to help me understand the new concepts, so please bear with me.

Most of my comments reduce to one of these 3 points:
- Pools are quite central to the `Fabric` controller (or users). Hence, we may want to increase our test coverage to make sure we don't run into regressions by accident in the future. 
- Unless there is a valid use case, I think I would try to make `ReclaimPolicy` immutable and avoid complex management corner cases for the time being. 
- We should extend our user docs with the expected behavior for a resizing of the pool. In this regard, are we covering it correctly? For example, after downsizing a pool, e.g., an allocated prefix gets removed,  the controllers sets an`ValueOutOfRangeReason`,  but the Claim remains Allocated and continues to report its value to the controller.


Happy to hear your thoughts on this. Please let me know if any of the comments needs further explanation. But again, amazing work, looking forward merging this PR!

api/pool/v1alpha1/groupversion_info.go
felix-kaestner: This got removed in a later commit.

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: Yes, I'd rather first look into the reconcilation metrics, which already exposes the time for an entire reconcilation before we start measuring this function in particular.

api/pool/v1alpha1/ipaddresspool_types.go
felix-kaestner: I don't think this would be a problem here, as we compute `addr` from `prefix.Masked().Addr().Next()`. And the prefix comes from the spec which is validated to be a CIDR, so naturally we will never have zones.

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: Why would this trigger an infinite loop? The guard is on `idx <= r.End` so this can't overflow.

api/core/v1alpha1/index_range.go
felix-kaestner: Changed in all files.

api/core/v1alpha1/index_range.go
felix-kaestner: I've changed this to int64. This also avoids all the type conversions.

api/pool/v1alpha1/claim_types.go
felix-kaestner: Added

api/pool/v1alpha1/claimref_types.go
felix-kaestner: Moved to that file

api/pool/v1alpha1/claimref_types.go
felix-kaestner: Updated the comment to explain it better.

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: Made it immutable.

api/pool/v1alpha1/ipaddresspool_types.go
felix-kaestner: Changed.

api/pool/v1alpha1/ipprefixpool_types.go
felix-kaestner: Made the prefixLength a field and the spec, so it applies to all prefixes and made the list of prefixes a set, such that each prefix is unique. I think an overlap would only be possible now if you e.g. provide `10.0.0.0/8` and `10.0.0.0/16`. I guess we will need a validation webhook for this. I'll do that as a follow-up.

docs/concepts/numbered-resources.md
felix-kaestner: Changed.

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: Changed the range to work on int64.

internal/controller/pool/claim_controller.go
felix-kaestner: Changed.

internal/controller/pool/claim_controller.go
felix-kaestner: Changed.

internal/controller/pool/claim_controller.go
felix-kaestner: Changed.

internal/controller/pool/claim_controller.go
felix-kaestner: Changed the watcher, so this is not a problem any longer.

internal/controller/pool/claim_controller_test.go
felix-kaestner: Added these test cases.

api/pool/v1alpha1/indexpool_types.go
nikatza: `IndexRange` uses int64. If `idx`  and `r.End` both have value `MaxInt64` then we would still enter the scope of the `for`. At its end we increase `idx` which then becomes MinInt64, so the next time we check the condition will again succeed. 

This wouldn't happen if r.End can't take value MaxInt64.

docs/assets/numbered-resources-dark.png
nikatza, line 1: Could we have these in svg?

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: > if r.End can't take value MaxInt64.

I think this would need a validation webhook.

nikatza: That is fine for me, but lets make sure we address it. Shall we create an issue to track it? (feel free to close this comment).

api/pool/v1alpha1/ipaddresspool_types.go
```diff
@@ -37,7 +40,7 @@ type IPAddressPoolStatus struct {
 
 	// Allocated is the number of allocated IP addresses.
 	// +optional
-	Allocated string `json:"allocated,omitempty"`
+	Allocated int32 `json:"allocated"`
```
nikatza: `int32` only holds for IPv4, but `corev1alpha1.IPAddr` supports Ipv6. May be `big.Int`, a `16[byte]`, or two `uint64`s?

api/pool/v1alpha1/ipaddresspool_types.go
```diff
@@ -117,31 +97,16 @@ func (p *IPAddressPool) Total() *big.Int {
...
-}
-
 // IsExhausted returns true if all available IP addresses have been allocated.
 func (p *IPAddressPool) IsExhausted() bool {
 	total := p.Total()
 	if total.Sign() == 0 {
 		return true
 	}
-	allocated := big.NewInt(int64(p.Allocated()))
+	allocated := big.NewInt(int64(p.Status.Allocated))
```
nikatza: Given my comment above, we might need to remove the inner cast to `int64`.

api/pool/v1alpha1/claim_types.go
```diff
@@ -4,15 +4,52 @@
...
 )
 
-// ClaimSpec defines the desired state of Claim
+// Allocation is an interface implemented by allocation objects (Index, IPAddress, IPPrefix).
+// +kubebuilder:object:generate=false
+type Allocation interface {
+	client.Object
+
+	// GetValue returns the allocated value as a string.
+	GetValue() string
```
nikatza: We could make this just `Value()` (this would apply to several other places)

api/pool/v1alpha1/claim_types.go
```diff
@@ -4,15 +4,52 @@
...
+
+	// GetClaimRef returns the ClaimRef bound to this allocation, or nil if unbound.
+	GetClaimRef() *ClaimRef
```
nikatza: Similarly this could be just  `ClaimRef()`

api/pool/v1alpha1/claim_types.go
```diff
@@ -4,15 +4,52 @@
...
+	client.Object
+
+	// IsExhausted reports whether all allocatable resources in the pool are taken.
+	IsExhausted() bool
+
+	// GetReclaimPolicy returns the pool's configured reclaim policy.
+	GetReclaimPolicy() ReclaimPolicy
+
+	// ListAllocations lists allocation objects matching the given options.
+	ListAllocations(ctx context.Context, c client.Client, opts ...client.ListOption) ([]Allocation, error)
```
nikatza, line 43: And this could be just `Allocations()`.

I am not going to add further nits wrt. to naming as these repeat through out all the files. Just wanted to raise the point. Feel free to ignore these comments.

api/pool/v1alpha1/claim_types.go
```diff
@@ -4,15 +4,52 @@
...
+// Pool is an interface that abstracts over the different pool types (IndexPool, IPAddressPool, IPPrefixPool) that a Claim can reference.
+// +kubebuilder:object:generate=false
+type Pool interface {
+	client.Object
+
+	// IsExhausted reports whether all allocatable resources in the pool are taken.
+	IsExhausted() bool
+
+	// GetReclaimPolicy returns the pool's configured reclaim policy.
+	GetReclaimPolicy() ReclaimPolicy
```
nikatza: Just `ReclaimPolicy`?

docs/api-reference/index.md
```diff
@@ -10,6 +10,9 @@
 
 ## networking.metal.ironcore.dev/v1alpha1
 
+SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+SPDX-License-Identifier: Apache-2.0
```
nikatza: I guess this was auto-generated? Don't know why this is added but it contains a repetion of line 18.

api/pool/v1alpha1/indexpool_types.go
felix-kaestner: Added in https://github.com/ironcore-dev/network-operator/pull/258/changes/807eadab8c6c088e44411bd9cf76cfce1b8339ee.

api/pool/v1alpha1/ipprefixpool_types.go
felix-kaestner: Added in https://github.com/ironcore-dev/network-operator/pull/258/changes/807eadab8c6c088e44411bd9cf76cfce1b8339ee.

api/pool/v1alpha1/ipprefixpool_types.go
```diff
@@ -0,0 +1,237 @@
...
+		sum := uint16(b[i]) + carry
+		b[i] = uint8(sum) // #nosec G115
+		carry = sum >> 8
+	}
+	result := netip.AddrFrom16(b)
+	if addr.Is4() {
+		return result.Unmap()
+	}
+	return result
+}
```
nikatza, line 178: I think this is not correct here. The problem is that when add we might overflow in a v4 address, transforming it into a IPv6. Here is a demo:


```
package main

import (
	"fmt"
	"net/netip"
)

// stepAddr unmodified version
func stepAddr(addr netip.Addr, n int) netip.Addr {
	b := addr.As16()
	// Add 2^n to the 128-bit value. Work from the least significant byte
	// (index 15) upward, carrying as needed.
	//
	// Example: advancing 10.0.0.192 by 2^6 (n=6, one /26 step):
	//	carry = 1 << (6%8) = 64,  start at byte 15 - 6/8 = 15
	//	byte 15: 192 + 64 = 256 → store 0, carry 1
	//	byte 14:   0 +  1 =   1 → store 1, carry 0  (stop)
	//	result: 10.0.1.0
	carry := uint16(1) << (n % 8)
	for i := 15 - n/8; i >= 0 && carry > 0; i-- {
		sum := uint16(b[i]) + carry
		b[i] = uint8(sum) // #nosec G115
		carry = sum >> 8
	}
	result := netip.AddrFrom16(b)
	if addr.Is4() {
		return result.Unmap()
	}
	return result
}

// stepAddrFixed - version with overflow detection for both IPv4 and IPv6
func stepAddrFixed(addr netip.Addr, n int) netip.Addr {
	b := addr.As16()
	carry := uint16(1) << (n % 8)
	for i := 15 - n/8; i >= 0 && carry > 0; i-- {
		sum := uint16(b[i]) + carry
		b[i] = uint8(sum)
		carry = sum >> 8
	}
	result := netip.AddrFrom16(b)
	if addr.Is4() {
		unmapped := result.Unmap()
		// IPv4 overflow: carry corrupted the ::ffff: prefix, result is no longer IPv4
		if !unmapped.Is4() {
			return netip.Addr{}
		}
		return unmapped
	}
	// IPv6 overflow: carry propagated past byte 0
	if carry > 0 {
		return netip.Addr{}
	}
	return result
}

func main() {
	testCases := []struct {
		addr     string
		n        int
		expected string
		overflow bool
	}{
		{"10.0.0.192", 6, "10.0.1.0", false},
		{"255.255.255.254", 1, "OVERFLOW", true},
		{"ffff:ffff:ffff:ffff:ffff:ffff:ffff:fffe", 1, "OVERFLOW", true},
	}

	for _, tc := range testCases {
		addr := netip.MustParseAddr(tc.addr)
		orig := stepAddr(addr, tc.n)
		fixed := stepAddrFixed(addr, tc.n)

		fmt.Printf("%s + 2^%d:\n", tc.addr, tc.n)
		fmt.Printf("  expected: %s\n", tc.expected)
		fmt.Printf("  stepAddr:     %s\n", orig)
		fmt.Printf("  stepAddrFixed: %s\n", fixed)
		fmt.Println("")
	}
}
``` 

which yields:

```
10.0.0.192 + 2^6:
  expected: 10.0.1.0
  stepAddr:     10.0.1.0
  stepAddrFixed: 10.0.1.0

255.255.255.254 + 2^1:
  expected: OVERFLOW
  stepAddr:     ::1:0:0:0
  stepAddrFixed: invalid IP

ffff:ffff:ffff:ffff:ffff:ffff:ffff:fffe + 2^1:
  expected: OVERFLOW
  stepAddr:     ::
  stepAddrFixed: invalid IP
```

api/pool/v1alpha1/ipprefixpool_types.go
```diff
@@ -18,10 +18,17 @@ import (
...
 	// +kubebuilder:validation:MinItems=1
-	Prefixes []IPPrefixPoolPrefix `json:"prefixes"`
+	// +listType=set
+	Prefixes []corev1alpha1.IPPrefix `json:"prefixes"`
+
+	// PrefixLength is the prefix length to allocate within each base prefix.
+	// +required
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Maximum=128
+	PrefixLength int32 `json:"prefixLength"`
```
nikatza: The naming is at first a bit confusing. Could we use something like either `PrefixLength -> PrefixPartitionLength` or `Prefixes -> BasePrefixes`?

nikatza (commented):
Great additions, just a couple of thing I noticed. Looking forward merging this in the next hours :)

docs/assets/numbered-resources-dark.png
felix-kaestner: Already converted to svg in a later commit.

api/pool/v1alpha1/ipaddresspool_types.go
felix-kaestner: Changed Status.Allocated from int32 to int64 on all three pool types and removed the redundant int64() / int32() casts in the controllers and IsExhausted() methods.

In practice this field will never exceed int32 since it counts Kubernetes objects (and the API server has its own etcd limits long before 2B resources). But int64 avoids the lint-suppressed narrowing casts (#nosec G115) and is consistent with Total() which already returns int64 / *big.Int.

api/pool/v1alpha1/ipaddresspool_types.go
felix-kaestner: See https://github.com/ironcore-dev/network-operator/pull/258#discussion_r3595013360.

api/pool/v1alpha1/claim_types.go
felix-kaestner: Renamed.

api/pool/v1alpha1/claim_types.go
felix-kaestner: Renamed.

api/pool/v1alpha1/claim_types.go
felix-kaestner: Renamed.

api/pool/v1alpha1/claim_types.go
felix-kaestner: Here I think this shouldn't be just `Allocations()`, because this is not a getter, but actually performs an API call with the kubernetes client. As such, I think I would prefer to leave it as `ListAllocations`, especially for consistency with the `List` method called on the kubernetes client.

docs/api-reference/index.md
felix-kaestner: I would assume this is because we now have files with a different copyright year in them.

api/pool/v1alpha1/ipprefixpool_types.go
felix-kaestner: Added overflow detection to stepAddr. It now returns a zero netip.Addr{} if the addition wraps past the address space (carry left over after the oop, or IPv4 unmaps to non-v4). 

The `Allocate` loop breaks naturally since masked.Contains is false for a zero addr. Added your test cases to verify.

api/pool/v1alpha1/ipprefixpool_types.go
felix-kaestner: Renamed to `AllocationPrefixLength`.

docs/api-reference/index.md
felix-kaestner: I've added a postprocessing step to the docs generation, which simply strips these lines.

nikatza (approved):
nicely done 🚀

PR #263, opened 2026-03-27
feat: Implement DeviceProvider for IOS-XR

sven-rosenzweig:
Implement DeviceProvider for IOS-XR

### DIFF
--- a/internal/provider/cisco/iosxr/intf.go
+++ b/internal/provider/cisco/iosxr/intf.go
@@ -15,16 +15,20 @@ import (
 var (
 	bundleEtherRE       = regexp.MustCompile(`^Bundle-Ether(\d+)(?:\.\d+)?$`)
 	physicalInterfaceRE = regexp.MustCompile(`^(TenGigE|TwentyFiveGigE|FortyGigE|HundredGigE|GigabitEthernet)(\d){1}(\/\d){2}(\/\d+){1}(.\d{1,5})?$`)
+	loopbackInterfaceRE = regexp.MustCompile(`^(Loopback|Lo)\d+$`)
+	mgmtEthInterfaceRe  = regexp.MustCompile(`^MgmtEth\d+\/RP\d+\/CPU\d+\/\d+$`)
 )
 
-type IFaceSpeed string
+type IFaceOwner string
 
 const (
-	Speed10G    IFaceSpeed = "TenGigE"
-	Speed25G    IFaceSpeed = "TwentyFiveGigE"
-	Speed40G    IFaceSpeed = "FortyGigE"
-	Speed100G   IFaceSpeed = "HundredGigE"
-	EtherBundle IFaceSpeed = "etherbundle"
+	Speed10G    IFaceOwner = "TenGigE"
+	Speed25G    IFaceOwner = "TwentyFiveGigE"
+	Speed40G    IFaceOwner = "FortyGigE"
+	Speed100G   IFaceOwner = "HundredGigE"
+	EtherBundle IFaceOwner = "etherbundle"
+	LoopBack    IFaceOwner = "Loopback"
+	MgmtEth     IFaceOwner = "MgmtEth"
 )
 
 type BundlePortActivity string
@@ -45,6 +49,16 @@ const (
 	StateAdminDown PhysIfStateType = "im-state-admin-down"
 	StateShutDown  PhysIfStateType = "im-state-shutdown"
 )
+
+type Ifaces struct {
+	PhysIfList []struct {
+		Name string `json:"interface-name"`
+	} `json:"interface-configuration"`
+}
+
+func (*Ifaces) XPath() string {
+	return "Cisco-IOS-XR-ifmgr-cfg:interface-configurations"
+}
 
 // Iface represents physical and bundle interfaces as part of the same struct as they share a lot of common configuration
 // and only differ in a few attributes like the interface name and the presence of bundle configuration or not.
@@ -183,10 +197,20 @@ func ValidateInterfaceName(name string) error {
 	return fmt.Errorf("unsupported interface name format: %s", name)
 }
 
-func ExtractInterfaceSpeedFromName(ifaceName string) (IFaceSpeed, error) {
-	// Owner of bundle interfaces is 'etherbundle'
+// Extract the owner of an interface based on the interface name.
+// For physical interfaces the owner matches the speed extracted from the interface name, e.g. 'TenGigE' for interface TwentyFiveGigE0/0/0/3
+// For bundle interfaces the owner is 'etherbundle'
+func ExtractOwnerFromInterfaceName(ifaceName string) (IFaceOwner, error) {
 	if bundleEtherRE.MatchString(ifaceName) {
 		return EtherBundle, nil
+	}
+
+	if loopbackInterfaceRE.MatchString(ifaceName) {
+		return LoopBack, nil
+	}
+
+	if mgmtEthInterfaceRe.MatchString(ifaceName) {
+		return MgmtEth, nil
 	}
 
 	// Match the port_type in an interface name <port_type>/<rack>/<slot/<module>/<port>
@@ -194,7 +218,7 @@ func ExtractInterfaceSpeedFromName(ifaceName string) (IFaceSpeed, error) {
 	re := regexp.MustCompile(`^\D*`)
 	speed := string(re.Find([]byte(ifaceName)))
 	if speed == "" {
-		return "", fmt.Errorf("failed to extract speed from interface name %s", ifaceName)
+		return "", fmt.Errorf("failed to extract interface speed from interface name %q", ifaceName)
 	}
 
 	switch speed {
@@ -207,7 +231,24 @@ func ExtractInterfaceSpeedFromName(ifaceName string) (IFaceSpeed, error) {
 	case string(Speed100G):
 		return Speed100G, nil
 	default:
-		return "", fmt.Errorf("unsupported interface type %s", speed)
+		return "", fmt.Errorf("unsupported interface speed %q", speed)
+	}
+}
+
+func MapInterfaceOwnerToSpeed(speed IFaceOwner) (int32, error) {
+	switch speed {
+	case Speed10G:
+		return 10_000, nil
+	case Speed25G:
+		return 25_000, nil
+	case Speed40G:
+		return 40_000, nil
+	case Speed100G:
+		return 100_000, nil
+	case EtherBundle, LoopBack, MgmtEth:
+		return 0, nil
+	default:
+		return 0, fmt.Errorf("unsupported interface owner %s", speed)
 	}
 }
 
--- a/internal/provider/cisco/iosxr/provider.go
+++ b/internal/provider/cisco/iosxr/provider.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"time"
 
 	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
 	"github.com/ironcore-dev/network-operator/internal/deviceutil"
@@ -20,6 +21,7 @@ import (
 
 var (
 	_ provider.Provider          = &Provider{}
+	_ provider.DeviceProvider    = &Provider{}
 	_ provider.InterfaceProvider = &Provider{}
 )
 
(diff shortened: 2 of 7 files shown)


internal/provider/cisco/iosxr/intf.go
```diff
@@ -46,6 +46,16 @@ const (
 	StateShutDown  PhysIfStateType = "im-state-shutdown"
 )
 
+type Ifaces struct {
+	PhysIfList []struct {
+		Name string `json:"interface-name"`
+	} `json:"interface-configuration"`
+}
+
+func (i *Ifaces) XPath() string {
```
felix-kaestner:
```suggestion
func (*Ifaces) XPath() string {
```

internal/provider/cisco/iosxr/intf.go
```diff
@@ -211,6 +224,21 @@ func ExtractInterfaceSpeedFromName(ifaceName string) (IFaceSpeed, error) {
 	}
 }
 
+func MapInterfaceSpeedToNumeric(speed IFaceSpeed) (int32, error) {
+	switch speed {
+	case Speed10G:
+		return 10000, nil
```
felix-kaestner:
```suggestion
		return 10_000, nil
```
nit: you can format numbers like this. Makes it a bit easier to read perhaps.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,71 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+		s, err := ExtractInterfaceSpeedFromName(intf.Name)
+		if err != nil {
+			continue
+		}
+
+		n, err := MapInterfaceSpeedToNumeric(s)
+		if err != nil {
+			return []provider.DevicePort{}, fmt.Errorf("failed to map interface speed to numeric value: %w", err)
+		}
+		if s != "" {
```
felix-kaestner: I think that ExtractInterfaceSpeedFromName should never return an empty string with a non-nil `error`. So in theory this check shouldn't be needed.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,71 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+	dp := make([]provider.DevicePort, 0, len(iFaces.PhysIfList))
+	for _, intf := range iFaces.PhysIfList {
+		s, err := ExtractInterfaceSpeedFromName(intf.Name)
+		if err != nil {
+			continue
+		}
+
+		n, err := MapInterfaceSpeedToNumeric(s)
+		if err != nil {
+			return []provider.DevicePort{}, fmt.Errorf("failed to map interface speed to numeric value: %w", err)
```
felix-kaestner:
```suggestion
			return nil, fmt.Errorf("failed to map interface speed to numeric value: %w", err)
```

internal/provider/cisco/iosxr/system.go
```diff
@@ -0,0 +1,50 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package iosxr
+
+import "time"
+
+const Manufacturer = "Cisco"
+
+// Model is the chassis model of the device, e.g. "NCS-57C3-MOD-SYS".
```
felix-kaestner: Comments should start with the Name of the type. also applies to the `SystemTime`.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,67 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+}
+
+func (p *Provider) GetDeviceInfo(ctx context.Context) (*provider.DeviceInfo, error) {
+	i := new(BasicDeviceInfo)
+
+	if err := p.client.GetState(ctx, i); err != nil {
+		return nil, err
+	}
+
+	return &provider.DeviceInfo{
```
felix-kaestner, line 88: We will be adding the `Hostname` to this struct in https://github.com/ironcore-dev/network-operator/pull/292/changes/6ecc0c9f0c8e151aa8359c0bc95f832e84ec8a15. Could you already prepare for that here (i.e. fetch the value and have it commented out on the return, so we can simply use it then)?

sven-rosenzweig: Fetching this information from the device. Added a ToDo to set this in the provider config once Enrics PR is merged.

internal/provider/cisco/iosxr/system_test.go
```diff
@@ -0,0 +1,29 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
nikatza, line 1: nit: 2025 -> 2026 (if you are to rebase anyhow, same applies to the next file)

internal/provider/cisco/iosxr/system.go
```diff
@@ -0,0 +1,55 @@
...
+	TimeZone    string `json:"time-zone"`
+	Year        int    `json:"year"`
+}
+
+func (*SystemTime) XPath() string {
+	return "Cisco-IOS-XR-shellutil-oper:/system-time"
+}
+
+func (c *Clock) ConvertToTime() time.Time {
+	return time.Date(c.Year, time.Month(c.Month), c.Day, c.Hour, c.Minute, c.Second, c.Millisecond, time.UTC)
```
nikatza: I think this is ignoring the time-zone and always using `UTC`, but may be `iosxr` does not fully support it. If it does then something like this might help:
```
	loc, err := time.LoadLocation(c.TimeZone)
	if err != nil {
		loc = time.UTC
	}
	return time.Date(c.Year, time.Month(c.Month), c.Day, c.Hour, c.Minute, c.Second, c.Millisecond, loc)
```

If it should always return  UTC then we probably should comment this in the function.

internal/provider/cisco/iosxr/intf.go
```diff
@@ -183,31 +193,54 @@ func ValidateInterfaceName(name string) error {
 	return fmt.Errorf("unsupported interface name format: %s", name)
 }
 
-func ExtractInterfaceSpeedFromName(ifaceName string) (IFaceSpeed, error) {
+func ExtractOwnerFromInterfaceName(ifaceName string) (IFaceSpeed, error) {
```
nikatza: The name of this function is a bit confusing: it returns an `IFaceSpeed` type but the name refers to an "Owner".  May be we can add an example explaining what it does so that it is easier to understand the reasoning :)

internal/provider/cisco/iosxr/intf.go
```diff
@@ -183,31 +193,54 @@ func ValidateInterfaceName(name string) error {
...
 	// Owner of bundle interfaces is 'etherbundle'
 	if bundleEtherRE.MatchString(ifaceName) {
 		return EtherBundle, nil
 	}
+	speed := ExtractInterfaceSpeedFromName(ifaceName)
 
+	if speed == "" {
```
nikatza: I am curious why `ExtractInterfaceSpeedFromName` does not return an error itself. The comparison with `""` is a bit odd, as it could have several meanings (depending on the reader).

internal/provider/cisco/iosxr/intf.go
```diff
@@ -183,31 +193,54 @@ func ValidateInterfaceName(name string) error {
...
-		return Speed25G, nil
+		return Speed25G
 	case string(Speed40G):
-		return Speed40G, nil
+		return Speed40G
 	case string(Speed100G):
-		return Speed100G, nil
+		return Speed100G
+	default:
+		return ""
```
nikatza: As mentioned above, may be this could return an error or on `SpeedUndefined` (Although in this case the device did actually return something, just that we have it mapped). Do we have a reference where these values come from that we could add here?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+		Manufacturer:    Manufacturer,
+		Model:           i.Model,
+		SerialNumber:    i.SerialNumber,
+		FirmwareVersion: i.FirmwareVersion,
+		// TODO(sven-rosenzweig): add hostname to provider.DeviceInfo once the PR containing the field is merged
+		// Hostname:        i.Hostname,
```
nikatza: PR is now merged :)

internal/provider/cisco/iosxr/provider.go
```diff
@@ -64,7 +129,7 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
 	// SubInterface <PotySpeed><rack><slot><port>.<vlan-id> e.g TwentyFiveGigE0/0/0/3
 	// Bundle Interface/Port Channel Bundle-Ether<BundleID>
 	// Vlans over Bundle Bundle-Ether<BundleID>.<vlan-id>
-	_, err := ExtractInterfaceSpeedFromName(name)
+	_, err := ExtractOwnerFromInterfaceName(name)
```
nikatza: something like this might be more idiomatic:

```suggestion
if _, err := ExtractOwnerFromInterfaceName(name); err != nil {
		return err
}
```

internal/provider/cisco/iosxr/intf.go
nikatza: Actually in the `MapInterfaceSpeedToNumeric` we return an `error`. Should `ExtractInterfaceSpeedFromName` adopt the same behavior?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
 	return p.conn.Close()
 }
 
+func (p *Provider) ListPorts(ctx context.Context) ([]provider.DevicePort, error) {
+	iFaces := new(Ifaces)
+	err := p.client.GetConfig(ctx, iFaces)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list ports: %w", err)
+	}
```
nikatza, line 58:
```suggestion
	if err := p.client.GetConfig(ctx, iFaces); err != nil {
		return nil, fmt.Errorf("failed to list ports: %w", err)
	}
```

might be more idiomatic

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+
+	dp := make([]provider.DevicePort, 0, len(iFaces.PhysIfList))
+	for _, intf := range iFaces.PhysIfList {
+		s := ExtractInterfaceSpeedFromName(intf.Name)
```
nikatza: Since `s` can be empty as per "ExtractInterfaceSpeedFromName", could/should we already early skip this iteration already in that case?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+		n, err := MapInterfaceSpeedToNumeric(s)
+		if err != nil {
+			return nil, fmt.Errorf("failed to map interface speed to numeric value: %w", err)
+		}
+		if s != "" {
```
nikatza: Isn't `s` always non-empty and has a valid value? (otherwise `MapInterfaceSpeedToNumeric` would have failed)

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+func (p *Provider) GetLastRebootTime(ctx context.Context) (time.Time, error) {
+	t := new(SystemTime)
+	if err := p.client.GetState(ctx, t); err != nil {
+		return time.Time{}, err
+	}
+	uptimeDuration := time.Second * time.Duration(t.Uptime.Uptime) * -1
+	return t.CurrTime.ConvertToTime().Add(uptimeDuration), nil
+}
+
+func (p *Provider) Reboot(ctx context.Context, conn *deviceutil.Connection) error {
```
nikatza:
```suggestion
func (p *Provider) Reboot(_ context.Context, conn *deviceutil.Connection) error {
```

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+	return errors.New("IOS XR Provider does not support rebooting the device")
+}
+
+func (p *Provider) FactoryReset(ctx context.Context, conn *deviceutil.Connection) error {
```
nikatza:
```suggestion
func (p *Provider) FactoryReset(_ context.Context, conn *deviceutil.Connection) error {
```

internal/provider/cisco/iosxr/provider.go
```diff
@@ -48,6 +50,69 @@ func (p *Provider) Disconnect(ctx context.Context, conn *deviceutil.Connection)
...
+	return errors.New("IOS XR Provider does not support factory reset")
+}
+
+func (p *Provider) Reprovision(cxt context.Context, conn *deviceutil.Connection) error {
```
nikatza:
```suggestion
func (p *Provider) Reprovision(_ context.Context, conn *deviceutil.Connection) error {
```

internal/provider/cisco/iosxr/system.go
```diff
@@ -0,0 +1,55 @@
...
+
+	// FirmwareVersion is the firmware version of the device, e.g. "25.2.2".
+	FirmwareVersion string `json:"firmware-version"`
+
+	// Hostname is the hostname of the device, e.g. "router1".
+	Hostname string `json:"name"`
+}
+
+func (*BasicDeviceInfo) XPath() string {
+	return "Cisco-IOS-XR-platform-inventory-oper:/platform-inventory/racks/rack[name=0]/attributes/basic-info"
```
nikatza, line 26: Is `rack[name=0]` constant? If not we could add a comment to the function about why we do it (for example I do not know :) )

nikatza (commented):
I have some comments, mostly small things than can be fixed quickly. Let me know if some lack clarity. Thanks for pushing this!

internal/provider/cisco/iosxr/intf.go
sven-rosenzweig: discussed via slack - renamed to ``ExtractInterfaceOwner``

internal/provider/cisco/iosxr/system_test.go
sven-rosenzweig: updated.

internal/provider/cisco/iosxr/provider.go
sven-rosenzweig: Returns an error now.

internal/provider/cisco/iosxr/system.go
sven-rosenzweig: added a comment. rack defines the chassis number. can be set to 0 as we have single chassis system

nikatza (approved):
Nice! IOS XR incoming 🚀

PR #271, opened 2026-03-31
Remove Nix shell environment

felix-kaestner:
Nix is not a common tool in the Kubernetes operator ecosystem and adds maintenance overhead without meaningful benefit. No one within the development team behind this project is actively using nix.

As a result, we drop shell.nix, .envrc, and disable Nix generation in Makefile.maker.yaml.


nikatza (approved):
nice!

PR #277, opened 2026-04-02
Move `gnmiext` into new `transport` package

felix-kaestner:
As we plan to use our `gnmiext` package also for other provider
implementations of other hardware vendors than Cisco, we now move it
outside of the cisco provider package scope.

With this change we are introducing a new `transport` package which will
house multiple packages related to transport/protocol specific utilities
and clients.

Additionally a new `transport/grpcext` package is added which refactors
out all the grpc related code from the `deviceutil` package and add an 
additional interceptor for handling non-retryable grpc error codes.


internal/deviceutil/deviceutil.go
```diff
@@ -216,3 +220,37 @@ func UnaryDefaultTimeoutInterceptor(timeout time.Duration) grpc.UnaryClientInter
...
+}
+
+// terminalCodes holds the set of gRPC codes that are considered terminal.
+// That is, if an error has one of these codes, retrying the operation
+// is not expected to succeed.
+// This list is based on the gRPC documentation at https://grpc.io/docs/guides/status-codes.
+var terminalCodes = []codes.Code{
+	codes.Unknown,
+	codes.InvalidArgument,
+	codes.NotFound,
```
nikatza: I'm not so sure about this one. It  am thinking of a `Get` that fails with this error code -not sure if this is even the case- and for which the relevant data is later added by the action of a different controller. Selectively picking the doc it states: "[...]", such as gradual feature rollout [...]". So data may appear later and thus maybe this shouldn't be a `TerminalError`.

I acknowledge we don't have the necessary operational experience to make a sound decision at the moment. This seems however a nice concept we could use.

internal/transport/gnmiext/empty.go
```diff
@@ -39,7 +43,7 @@ func (e *Empty) UnmarshalJSON(b []byte) error {
 		*e = false
 		return nil
 	}
-	if string(b) != "[null]" {
+	if !nullRe.MatchString(string(b)) {
```
nikatza, line 46: We could think of only checking this only if we find an IOS-XR in `client.capabilities.capabilities`. That said, this is likely too much for now as we are unlikely to find this being used as valid value in different platform. On the other hand,  as we increase the number of platforms supported we may have more of these particularities coming in. Anyhow, I would leave this at it is unless you think otherwise.

internal/transport/gnmiext/client_test.go
```diff
@@ -105,7 +105,7 @@ func TestClient_GetConfig(t *testing.T) {
 	tests := []struct {
 		name    string
 		conn    grpc.ClientConnInterface
-		conf    []Configurable
+		conf    []DataElement
```
nikatza: may be we can rename `conf` to  `configs` to be in line with the terminology above. If we want it generic then `d`, `de`, or `data` might be other options.

internal/transport/gnmiext/client_test.go
```diff
@@ -326,7 +326,7 @@ func TestClient_GetConfig(t *testing.T) {
 					}, nil
 				},
 			},
-			conf:    []Configurable{new(Hostname)},
+			conf:    []DataElement{new(Hostname)},
```
nikatza: If we are not generic then we could rename this to `updates`

nikatza (approved):

internal/deviceutil/deviceutil.go
felix-kaestner: > which the relevant data is later added by the action of a different controller

I think this would potentially be a bad pattern, as controllers should ideally model their dependencies and validate them and then take full ownership of their data. So in this case, I feel like it might be better to have the `TerminalError` instead of it eventually working due to the  exponential backoff. So in that case I would hope that it intentionally doesn't work to give us that insight.

I would propose we keep it for now, and if we see it leading to false results, we can still remove it later.

internal/transport/gnmiext/empty.go
felix-kaestner: I actually like this a lot! Will use our `gnmiext.Marshaler` for this purpose, as it was made exactly for this.

felix-kaestner: Hmm, thinking about it again, if we would go with this, then we'd have something like:

```go
// UnmarshalYANG implements [Marshaler] for Empty.
func (e *Empty) UnmarshalYANG(c *Capabilities, b []byte) error {
	if slices.ContainsFunc(c.SupportedModels, func(m Model) bool {
		return strings.HasPrefix(strings.ToLower(m.Name), "cisco-ios-xr")
	}) {
		if len(b) == 0 || string(b) == "null" {
			*e = false
			return nil
		}
		if !nullRe.MatchString(string(b)) {
			return fmt.Errorf("gnmiext: invalid empty value: %s", string(b))
		}
		*e = true
		return nil
	}
	return json.Unmarshal(b, e)
}
```

which means that for all the providers, we always iterate through all the YANG models in the capabilities on every unmarshal. This feels like a way greater performance hit than using a regex match vs. string comparison. 

So I'd say keeping it like this and just using the regex, might be preferable. What do you think?

internal/deviceutil/deviceutil.go
nikatza: I mistakenly used the word "different". I meant the same controller, in an posterior reconciliation which we here prevent.

internal/transport/gnmiext/empty.go
nikatza: This was my exact thought as well and also thought to better leave it this way for now.

PR #284, opened 2026-04-07
Reduce log noise by introducing a consistent verbosity scheme

felix-kaestner:
All log.Info calls across controllers and supporting packages have been re-leveled based on how frequently they fire, so that operators see only meaningful output by default and can opt into more detail as needed.

Verbosity levels:

  V(0) default  Errors only. Meaningful state transitions that a user
                should always be aware of: resource pausing/unpausing.

  V(1)          One-time lifecycle events per resource: finalizer added,
                status conditions initialized. Actual device configuration
                changes pushed via gNMI: updating, deleting, resetting
                config to default.

  V(2)          Infrequent operational events: watch-triggered enqueue
                requests, lock acquired/released, TLS certificate setup,
                gNMI no-op (configuration already up-to-date).

  V(3)          Per-loop noise hidden by default: reconciling resource,
                resource not found, resource being deleted, device locked
                and requeuing, reconciliation paused steady-state, internal
                resource lock state details.

### DIFF
--- a/.github/workflows/checks.yaml
+++ b/.github/workflows/checks.yaml
@@ -29,7 +29,7 @@ jobs:
         uses: actions/setup-go@v6
         with:
           check-latest: true
-          go-version: 1.26.1
+          go-version: 1.26.2
       - name: Run prepare make target
         run: make generate
       - name: Run golangci-lint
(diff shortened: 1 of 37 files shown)


nikatza (approved):
Nice, the scheme seems very reasonable and better than what we currently have.  I think the `dhcp_controller` we merged last week is missing.

PR #286, opened 2026-04-08
Add NX-API JSON-RPC client and hack/nxapi helper

felix-kaestner:
Introduce a new `nxapi` package in `internal/transport`, which includes a thin HTTP client for the Cisco NX-OS NX-API JSON-RPC endpoint.

Add `hack/nxapi`, a minimal CLI that sends one or more NX-API commands to a device and pretty-prints the JSON results.

### DIFF
--- a/hack/nxapi/main.go
+++ b/hack/nxapi/main.go
@@ -1,0 +1,105 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// nxapi is a minimal CLI tool for sending NX-API JSON-RPC commands to a
+// Cisco NX-OS device and printing the results.
+//
+// Usage:
+//
+//	go run ./hack/nxapi [flags] <cmd> [<cmd> ...]
+//
+// Example:
+//
+//	go run ./hack/nxapi -address 10.0.0.1:443 "show version" "show interface brief"
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"flag"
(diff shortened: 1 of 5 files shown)


hack/nxapi/main.go
```diff
@@ -0,0 +1,105 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
nikatza: nit: 2026

internal/transport/nxapi/nxapi.go
```diff
@@ -0,0 +1,254 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("nxapi: failed to read response body: %w", err)
+	}
+
+	res, err := decode(body)
+	if err != nil {
+		return nil, fmt.Errorf("nxapi: failed to decode response: %w", err)
+	}
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
nikatza, line 95: minor: We could first check the the status code and then read the body if there was no issue.

internal/provider/cisco/nxos/nxapi_test.go
```diff
@@ -0,0 +1,283 @@
...
+		})
+	}
+}
+
+func trim(t *testing.T, s string) string {
+	t.Helper()
+	s = strings.TrimSpace(s)
+	s = strings.Join(strings.Fields(s), "")
+	return s
+}
```
nikatza: Could we use `json.Compact` instead to keep the spaces?

nikatza (approved):

PR #287, opened 2026-04-08
Refactor GitHub Actions Workflows

felix-kaestner:
With this PR we stop using go-makefile-maker for GitHub Actions workflow generation. With recent additions of custom workflows, maintaining consistency between auto-generated and manually written pipelines became impractical. The effort to update all workflows already requires reviewing each one, so auto-generation provides diminishing value. Additionally, some generated workflows like CodeQL are redundant (enabled at repo level) while others like test-chart or publishing the docker image to `ghcr.io` required manual modifications anyway.

Going forward, all workflows will be manually maintained for this project, providing full control over CI/CD configuration.


.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,89 @@
...
+            type=ref,event=branch
+            type=ref,event=pr
+            # https://github.com/docker/metadata-action#typesha
+            type=sha,format=long
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+        id: setup-buildx
+      - name: Login to GitHub Container Registry
```
sven-rosenzweig, line 62: nit: After doing a little research on github actions, it seems to me the login step follows directly after the checkout step.

.github/workflows/publish-image.yml
```diff
@@ -0,0 +1,89 @@
...
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Go Cache for Docker
+        uses: actions/cache@v4
+        id: cache
+        with:
+          path: |
+            go-pkg-mod
+            go-build-cache
+          key: cache-mount-${{ hashFiles('go.sum') }}
+      - name: Restore Docker Cache Mounts
```
sven-rosenzweig, line 76: From my understanding this is necessary as we decided against pushing caches to the container registry and still want to reuse cache between different builds.  That's why we need the ``buildkit-cache-dance`` runner, right?

```
cache-from: type=registry,ref=user/app:buildcache
cache-to: type=registry,ref=user/app:buildcache,mode=max
```

If my understanding is correct, either a short comment or enhancing the git commit message, would be appreciated.

hack/validate-kustomize.sh
```diff
@@ -0,0 +1,18 @@
...
+BASEDIR=$(cd -- "$(dirname -- "$0")" && pwd)
+
+while IFS= read -r -d '' kustomization; do
+  dir=$(dirname "$kustomization")
+  name=${dir#"$BASEDIR/../"}
+  if kustomize build "$dir" >/dev/null 2>&1; then
+    echo "OK: $name"
+  else
+    echo "FAILED: $name"
+    exit 1
```
sven-rosenzweig: nit: we could fail at the end, after executing all kustomizations (if we see an error)

.github/workflows/size-label.yml
```diff
@@ -0,0 +1,25 @@
+# SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+# SPDX-License-Identifier: Apache-2.0
+#
+name: Size Label
+
+on:
+  pull_request_target:
+    types:
+      - opened
```
sven-rosenzweig, line 9: ``ready_for_review`` or ``reopend`` needed? 
https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target

charts/network-operator/templates/manager/manager.yaml
```diff
@@ -74,6 +68,9 @@ spec:
         - containerPort: 8080
           name: provisioning
           protocol: TCP
+        - containerPort: 1069
```
sven-rosenzweig, line 71: I guess this was detected by the new workflow. Maybe mentioning that this commits fixes uncommited changes as well, would be nice

Makefile.maker.yaml
```diff
@@ -56,15 +56,12 @@ testPackages:
   except: '/test'
```
sven-rosenzweig, line 56: Why don't we remove this file completely? Presence of that file might indicate that we are still using the Makefile.maker.

.github/workflows/publish-image.yml
felix-kaestner: The steps are modular, you can order them as you want as long as you make sure that you login to the container registry before pushing an image (obviously). So putting this after the checkout is purely preference. I think putting it before the build-and-push action makes most sense for use, so you co-locate the login→push flow.

.github/workflows/publish-image.yml
felix-kaestner: The commit message for this change already includes the link to https://docs.docker.com/build/ci/github-actions/cache/#github-cache. Where to put the cache is arguably a bit of preference, but for us GH Actions felt the most natural, as are already on GitHub+Actions. 

However, the cache backend for the container build (`cache-from: type=gha; cache-to: type=gha,mode=max`) and the `buildkit-cache-dance` are two distinct mechanisms.

.github/workflows/size-label.yml
felix-kaestner: Both `ready_for_review` or `reopend` don't effect the size of the PR, so re-running on those events seems unnecessary.

Makefile.maker.yaml
felix-kaestner: We are still using gomm for the Makefile itself. With this PR we are only removing the generated pipelines.

hack/validate-kustomize.sh
felix-kaestner: Adjusted.

charts/network-operator/templates/manager/manager.yaml
felix-kaestner: Done.

sven-rosenzweig (approved):

PR #290, opened 2026-04-09
Pause child resources based on device reachability and phase

felix-kaestner:
#### Summary

 - Introduce a dedicated Reachable condition on the Device that tracks connectivity independently from Ready. When the controller cannot connect, Reachable=False and Ready=Unknown (per K8s API conventions: state is unobservable, not definitively unhealthy). A HeartbeatInterval flag governs periodic requeuing.
 - Extend computeCondition in internal/paused to pause child resources when the device is unreachable (Reachable != True) or not in the Running phase, in addition to the existing spec.paused check.
 - Update all Device watch predicates across the resource controllers to use a new DevicePausedChanged helper, so child resources are paused promptly on reachability or phase changes rather than waiting for the next periodic requeue.

### DIFF
--- a/internal/controller/core/device_controller.go
+++ b/internal/controller/core/device_controller.go
@@ -50,9 +50,9 @@ type DeviceReconciler struct {
 	// Provider is the driver that will be used to create & delete the interface.
 	Provider provider.ProviderFunc
 
-	// RequeueInterval is the duration after which the controller should requeue the reconciliation,
+	// HeartbeatInterval is the duration after which the controller requeues the reconciliation,
 	// regardless of changes.
-	RequeueInterval time.Duration
+	HeartbeatInterval time.Duration
 }
 
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=devices,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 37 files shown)


internal/controller/core/device_controller.go
```diff
@@ -279,69 +279,81 @@ func (r *DeviceReconciler) reconcile(ctx context.Context, device *v1alpha1.Devic
...
+
+		log := ctrl.LoggerFrom(ctx)
+		if device.Labels == nil {
+			device.Labels = map[string]string{}
+		}
+		if serial := strings.ToLower(device.Status.SerialNumber); serial != "" {
+			if device.Labels[v1alpha1.DeviceSerialLabel] == "" {
+				device.Labels[v1alpha1.DeviceSerialLabel] = serial
+			} else if device.Labels[v1alpha1.DeviceSerialLabel] != serial {
+				log.Info("Device serial label does not match observed device serial number", "labelSerial", device.Labels[v1alpha1.DeviceSerialLabel], "observedSerial", serial)
```
nikatza, line 323: We might set this with `warn` severity,  since this only happens once after the new device is booted. I am concerned that this information will be eventually lost and the user won't be aware that the value in the label is wrong.

nikatza (approved):
May be we could extend `docs/concepts/pausing.md` with additional info -included already in the commit messages- about the `Phases` and the  failed connections.

internal/controller/core/device_controller.go
felix-kaestner: @nikatza there is no WARN severity. Only `log.Info` and `log.Error`.

PR #291, opened 2026-04-09
Fix inverted provisioning timeout check in device controller

elinalin:
## Summary

This PR fixes an inverted timeout condition in the device controller provisioning flow.

Previously, the controller marked a device as failed when:

`activeProv.StartTime.Add(time.Hour).After(time.Now())`

That condition is true while the provisioning deadline is still in the future, which means the controller could fail provisioning before the timeout was actually reached.

This PR changes the logic so that provisioning only fails after the deadline has passed.

## Changes

- fix the timeout check in the provisioning reconcile path
- add boundary tests for:
  - provisioning still in progress before the 1 hour timeout
  - provisioning failing after the timeout threshold

## Why this is needed

Without this fix, a device can transition to `Failed` too early and emit a misleading `ProvisioningFailed` event even though the provisioning window has not expired yet.

## Validation

Validated locally with:
- `gofmt`
- `go test -run '^$' ./internal/controller/core`
- `go test ./internal/provisioning`

### DIFF
--- a/internal/controller/core/device_controller_test.go
+++ b/internal/controller/core/device_controller_test.go
@@ -16,74 +16,6 @@ import (
 	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
 )
 
-func drainRecordedEvents() {
-	if testRecorder == nil {
-		return
-	}
-
-	for {
-		select {
-		case <-testRecorder.Events:
-		default:
-			return
-		}
-	}
-}
-
-func waitForRecordedEvent(expected string) {
-	Eventually(func(g Gomega) string {
-		select {
-		case event := <-testRecorder.Events:
-			return event
-		default:
-			return ""
-		}
-	}).Should(ContainSubstring(expected))
-}
-
-func setActiveProvisioningEntry(key client.ObjectKey, startTime time.Time) {
-	Eventually(func(g Gomega) {
-		resource := &v1alpha1.Device{}
-		g.Expect(k8sClient.Get(ctx, key, resource)).To(Succeed())
-
-		patch := resource.DeepCopy()
-		patch.Status.Phase = v1alpha1.DevicePhaseProvisioning
-		patch.Status.Provisioning = []v1alpha1.ProvisioningInfo{{
-			Token:     "test-token",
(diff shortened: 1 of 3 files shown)


internal/controller/core/device_controller_test.go
```diff
@@ -183,6 +228,65 @@ var _ = Describe("Device Controller", func() {
...
+
+			By("Verifying the device transitions to Failed phase")
+			Eventually(func(g Gomega) {
+				resource := &v1alpha1.Device{}
+				g.Expect(k8sClient.Get(ctx, key, resource)).To(Succeed())
+				g.Expect(resource.Status.Phase).To(Equal(v1alpha1.DevicePhaseFailed))
+			}).Should(Succeed())
+
+			By("Verifying the timeout event is emitted")
+			waitForRecordedEvent("ProvisioningFailed")
```
felix-kaestner: Let's please not validate on any recorded events for now. Checking that the device is in Phase `Failed` is already sufficient. This means we can also revert all the changes in regards to the events recorder (incl. `drainRecordedEvents`).

internal/controller/core/device_controller_test.go
```diff
@@ -16,6 +16,74 @@ import (
...
+		select {
+		case event := <-testRecorder.Events:
+			return event
+		default:
+			return ""
+		}
+	}).Should(ContainSubstring(expected))
+}
+
+func setActiveProvisioningEntry(key client.ObjectKey, startTime time.Time) {
```
felix-kaestner: Let's please move the code in this and the `newProvisioningDevice` funcs inline where they are used. They are only used in two places, and the minimal code duplication is not worth the tradeoff for readability here.

felix-kaestner (changes requested):

internal/controller/core/device_controller_test.go
elinalin: Reverted.

internal/controller/core/device_controller_test.go
elinalin: thanks Felix, fixed.

felix-kaestner (approved):

PR #292, opened 2026-04-09
Validate interfaces neighbors via LLDP

nikatza:
I am opening this PR even though there is still one action item missing: making the provider reject some of the naming schemes for the physical interfaces. I fear this would make this PR even longer than it is. So I will add this in a follow up PR. As for now, validation via labels and annotations work in the lab in both cases.

This is what an interface resource looks like with neighbor validation using an annotation (i skip the label as it is similar):
```
kind: Interface
metadata:
  annotations:
[...]
    networking.metal.ironcore.dev/interface-neighbor-raw: router2::Ethernet1/10
[...]
spec:
  adminState: Up
  description: vPC to Host1
  deviceRef:
    name: leaf1
  mtu: 1500
  name: eth1/10
  type: Physical
status:
[...]
  neighbors:
  - chassisId: 0c3f.2300.010a
    chassisIdType: 4
    expirationTime: "2026-04-13T14:47:50Z"
    portDescription: Ethernet1/10
    portId: Ethernet1/10
    portIdType: 5
    systemDescription: |-
      Cisco Nexus Operating System (NX-OS) Software 10.6(2)
      TAC support: http://www.cisco.com/tac
      Copyright (c) 2002-2025, Cisco Systems, Inc. All rights reserved.
    systemName: router2
    validation: Verified
```

 I have added context in the commit message, but please let me know if something is not clear.


api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+
+// Neighbor represents an LLDP neighbor discovered on an interface.
+type Neighbor struct {
+	// NeighborInfo contains the information about the LLDP neighbor discovered on this interface.
+	NeighborInfo `json:",inline"`
+	// ExpirationTime is the time when the LLDP neighbor information expires.
+	// It is calculated based on the TTL.
+	ExpirationTime *metav1.Time `json:"expirationTime,omitempty"`
+	// Validation indicates whether the LLDP neighbor information matches the information in the label or annotations of the interface.
+	Validation NeighborValidation `json:"validation"`
```
felix-kaestner: As by the default. These fields would now be required. Can we make this explicit?

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+
+// NeighborInfo represents the information retrieved via LLDP.
+type NeighborInfo struct {
+	SystemName        string `json:"systemName,omitempty"`
+	SystemDescription string `json:"systemDescription,omitempty"`
+	ChassisID         string `json:"chassisId,omitempty"`
+	ChassisIDType     uint8  `json:"chassisIdType,omitempty"`
+	PortID            string `json:"portId,omitempty"`
+	PortIDType        uint8  `json:"portIdType,omitempty"`
+	PortDescription   string `json:"portDescription,omitempty"`
```
felix-kaestner: Maybe just a very brief sentence for each field might be helpful.

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
 }
 
+type NeighborValidation string
+
+const (
+	// NeighborValidationUndefined indicates that there is not a label or annotation with the necessary information to validate the LLDP neighbor information against, or that the validation has not been performed yet.
+	NeighborValidationUndefined NeighborValidation = "Undefined"
```
felix-kaestner: Should we have Undefined? Or should the validation just be omitted, i.e. not show up in the status, when no label or annotation is set?

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+	// NeighborValidationError indicates that an error occurred while trying to validate the LLDP neighbor information against the interface labels or annotations.
+	NeighborValidationError NeighborValidation = "Error"
```
felix-kaestner: Having such a generic `Error` case feels weird, as a user wouldn't know any more details on why it failed. Looking into the controller, it seems like it's only used, when there is a `err != nil` case, in which the error is returned, causing a exponential backoff-retry. In such cases, you wouldn't typically update the status, as this would then result in flapping status updates when an operator faces intermittent issues that are then resolved by retrying. So I would use specific errors and a TerminalError when the validation can't be performed, or simply return the error as is, without status updates, when it's an intermittent issue (e.g. failed to retrieve a resource from the api server).

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+	SystemName        string `json:"systemName,omitempty"`
+	SystemDescription string `json:"systemDescription,omitempty"`
+	ChassisID         string `json:"chassisId,omitempty"`
+	ChassisIDType     uint8  `json:"chassisIdType,omitempty"`
+	PortID            string `json:"portId,omitempty"`
+	PortIDType        uint8  `json:"portIdType,omitempty"`
+	PortDescription   string `json:"portDescription,omitempty"`
 }
 
+type NeighborValidation string
```
felix-kaestner, line 527: I think this should have a description and a kubebuilder Enum statement.

api/core/v1alpha1/interface_types.go
```diff
@@ -399,7 +440,6 @@ type InterfaceStatus struct {
 // +kubebuilder:printcolumn:name="Operational",type=string,JSONPath=`.status.conditions[?(@.type=="Operational")].status`,priority=1
 // +kubebuilder:printcolumn:name="Paused",type=string,JSONPath=`.status.conditions[?(@.type=="Paused")].status`,priority=1
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
-
```
felix-kaestner, line 402: With this line removed, it breaks the rule, that a comment should always start with the type name (`Interface` in this case). So we should keep this separating line. For kubebuilder markers, they are still applied to the `Interface` resource.

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+	NeighborInfo `json:",inline"`
+	// ExpirationTime is the time when the LLDP neighbor information expires.
+	// It is calculated based on the TTL.
+	ExpirationTime *metav1.Time `json:"expirationTime,omitempty"`
+	// Validation indicates whether the LLDP neighbor information matches the information in the label or annotations of the interface.
+	Validation NeighborValidation `json:"validation"`
+}
+
+// NeighborInfo represents the information retrieved via LLDP.
+type NeighborInfo struct {
```
felix-kaestner: Is there a reason we define this struct here? As we either way inline this information into the `Neighbor` above. Can't this fields then just be put into the `Neighbor` struct as is?

internal/controller/core/interface_controller.go
```diff
@@ -541,6 +542,11 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
 		return fmt.Errorf("failed to get interface status: %w", err)
 	}
 
+	// Neighbor adjacencies is only metadata and should not prevent reconciliation
+	if err := r.updateNeighborAdjacenciesStatus(ctx, s, &status); err != nil {
+		ctrl.LoggerFrom(ctx).V(3).Info("failed to update neighbor adjacency status", "error", err, "interface", klog.KObj(s.Interface))
```
felix-kaestner: Should we still log this as an `ctrl.LoggerFrom(ctx).Error(err, ...)`?

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+// updateNeighborAdjacenciesStatus updates the Interface status with the LLDP neighbor adjacencies returned by the provider.
+// It validates the adjacencies by looking at the corresponding label/annotation.
+// It first attempts to validate through label (neighbor is managed by the operator and exists as a kubernetes resource).
+// If that fails, it attempts to validate through annotation (neighbor is not managed by the operator).
+// Only returns an error if there is an issue during the validation process, but does not return an error if the validation fails (i.e. the adjacency is marked as invalid).
+func (r *InterfaceReconciler) updateNeighborAdjacenciesStatus(ctx context.Context, s *scope, status *provider.InterfaceStatus) error {
+	if s.Interface.Spec.Type != v1alpha1.InterfaceTypePhysical || len(status.LLDPAdjacencies) == 0 {
+		s.Interface.Status.Neighbors = nil
+		return nil
+	}
```
felix-kaestner: I would like to do this check on the caller side and then not even call this function. Similar to the sub-reconciler checks, e.g. when `reconcileVRF` is called.

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+	Neighbors []Neighbor `json:"neighbors,omitempty"`
+}
+
+// Neighbor represents an LLDP neighbor discovered on an interface.
+type Neighbor struct {
+	// NeighborInfo contains the information about the LLDP neighbor discovered on this interface.
+	NeighborInfo `json:",inline"`
+	// ExpirationTime is the time when the LLDP neighbor information expires.
+	// It is calculated based on the TTL.
+	ExpirationTime *metav1.Time `json:"expirationTime,omitempty"`
```
felix-kaestner: If this field is required and always set, can it be a non-pointer?

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+				ChassisIDType:     adj.ChassisIDType,
+				PortID:            adj.PortID,
+				PortIDType:        adj.PortIDType,
+				PortDescription:   adj.PortDescription,
+			},
+			Validation: v1alpha1.NeighborValidationUndefined,
+		}
+
+		// The TTL reported by the device and the operator clock can drift one from each other for more than one second.
+		// If we use the TTL to compute ExpirationTime on each reconciliation, then ExpirationTime will keep changing, triggering unnecessary updates to the Interface status.
```
felix-kaestner: nit: Instead, we could also adjust the predicates to not trigger a reconciliation, if only the expiration time changed.

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+		// Thus, we preserve the existing ExpirationTime as long as the neighbor is still valid, at the cost of some inaccuracy.
+		if existing, found := existingNeighbors[key]; found && existing.ExpirationTime != nil && existing.ExpirationTime.After(time.Now()) {
+			adjacency.ExpirationTime = existing.ExpirationTime
+		} else {
+			t := metav1.NewTime(time.Now().Add(time.Duration(adj.TTL) * time.Second).Truncate(time.Second))
+			adjacency.ExpirationTime = &t
+		}
+
+		if err := r.validateLLDPAdjacencyThroughLabel(ctx, s.Interface, &adjacency); err != nil {
```
felix-kaestner: I feel like this function should have a return value (e.g. the `NeighborValidation` enum), instead of setting something implicitly in the passed `&adjacency` object which is then checked. Might make it easier to follow the code.

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+	return kerrors.NewAggregate(errs)
+}
+
+func (r *InterfaceReconciler) validateLLDPAdjacencyThroughLabel(ctx context.Context, intf *v1alpha1.Interface, n *v1alpha1.Neighbor) error {
+	neighborLabelValue, ok := intf.Labels[v1alpha1.PhysicalInterfaceNeighborLabel]
+	if !ok {
+		n.Validation = v1alpha1.NeighborValidationUndefined
+		return nil
+	}
+
```
felix-kaestner: Again, would move this to the caller. Then the `!= Undefined` check might also become easier.

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+	key := client.ObjectKey{
+		Name:      neighborLabelValue,
+		Namespace: intf.Namespace,
+	}
+
+	remoteIntf := new(v1alpha1.Interface)
+	if err := r.Get(ctx, key, remoteIntf); err != nil {
+		n.Validation = v1alpha1.NeighborValidationError
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get neighbor interface %q: %w", neighborLabelValue, err)
```
felix-kaestner: What about a specific NeighborNotFound state?

internal/controller/core/interface_controller.go
```diff
@@ -560,6 +566,149 @@ func (r *InterfaceReconciler) reconcile(ctx context.Context, s *scope) (reterr e
...
+	}
+
+	if remoteDevice.Status.Hostname == "" {
+		n.Validation = v1alpha1.NeighborValidationError
+		return fmt.Errorf("the neighbor device does not have a hostname yet, cannot validate adjacency: neighborInterface=%q", remoteIntf.Name)
+	}
+
+	if remoteDevice.Status.Hostname != n.SystemName {
+		n.Validation = v1alpha1.NeighborDeviceMismatch
+		log.Info("the neighbor device hostname does not match", "expected", n.SystemName, "actual", remoteDevice.Status.Hostname)
```
felix-kaestner: Can all these log statements have a `.V(\d)`? Probably something very low, perhaps just `1`. Just so we keep adhering to logging levels for all logs.

internal/controller/core/interface_lldp_test.go
felix-kaestner, line 1: Any reason why this is a new file and not put into the already existing `interface_controller_test.go`?

internal/controller/core/lldp_controller.go
```diff
@@ -433,28 +433,36 @@ func (r *LLDPReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager)
...
+					return false
+				},
+			}),
+		).
+		// Watches enqueues LLDPs for updates in referenced Interface resources.
+		// This ensures LLDP reconciles when a referenced Interface is created or updated.
+		Watches(
+			&v1alpha1.Interface{},
+			handler.EnqueueRequestsFromMapFunc(r.interfaceToLLDPs),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
```
felix-kaestner: For interfaces we are only looking at their `spec.name` (and the referenced device is immutable). So we might just use

```go
			builder.WithPredicates(predicate.Funcs{
					return false
				},
				GenericFunc: func(e event.GenericEvent) bool {
					return false
				},
			}),
```
i.e. only create and delete events.

internal/controller/core/lldp_controller.go
```diff
@@ -534,3 +542,41 @@ func (r *LLDPReconciler) deviceToLLDPs(ctx context.Context, obj client.Object) [
...
+	if !ok {
+		panic(fmt.Sprintf("Expected an Interface but got a %T", obj))
+	}
+
+	log := ctrl.LoggerFrom(ctx, "Interface", klog.KObj(intf))
+
+	list := new(v1alpha1.LLDPList)
+	if err := r.List(ctx, list,
+		client.InNamespace(intf.Namespace),
+		client.MatchingLabels{v1alpha1.DeviceLabel: intf.Spec.DeviceRef.Name},
```
felix-kaestner: Let's please use the device field indexer for reasons explained in https://github.com/ironcore-dev/network-operator/commit/144dc2ceb193a4d0333cd10d12253fffdb0851ea.

internal/provider/cisco/nxos/provider.go
```diff
@@ -1332,19 +1332,50 @@ func (p *Provider) GetInterfaceStatus(ctx context.Context, req *provider.Interfa
...
+				ChassisIDType:   adj.ChassisIDT,
+				PortID:          adj.PortIDV,
+				PortIDType:      adj.PortIDT,
+				PortDescription: adj.PortDesc,
+				SysName:         adj.SysName,
+				SysDescription:  adj.SysDesc,
+				TTL:             uint32(adj.TTL), // #nosec G115
+			}
+			lldpAdjacencies = append(lldpAdjacencies, neighbor)
+		}
```
felix-kaestner: Why are we doing this twice?

internal/provider/cisco/nxos/provider.go
```diff
@@ -1332,19 +1332,50 @@ func (p *Provider) GetInterfaceStatus(ctx context.Context, req *provider.Interfa
...
-		if err := p.client.GetState(ctx, phys); err != nil && !errors.Is(err, gnmiext.ErrNil) {
+		lldpAdj := new(LLDPAdjacencyItems)
+		lldpAdj.ID = name
+		if err := p.client.GetState(ctx, phys, lldpAdj); err != nil && !errors.Is(err, gnmiext.ErrNil) {
 			return provider.InterfaceStatus{}, err
 		}
 		operSt = phys.OperSt
 		operMsg = phys.OperStQual
 
+		for _, adj := range lldpAdj.AdjItems.AdjEpList {
```
felix-kaestner, line 1352: nit: a preallocation might bring a small benefit here:

```go		
lldpAdjacencies = make([]provider.LLDPAdjacency, 0, len(lldpAdj.AdjItems.AdjEpList))
```

internal/provider/cisco/nxos/lldp_test.go
```diff
@@ -22,3 +26,32 @@ func init() {
 
 	Register("lldp", lldp)
 }
+
+func TestLLDPAdjacencyItems_XPath(t *testing.T) {
```
felix-kaestner: Do we really need a test case for a simple `fmt.Sprintf()`?

internal/provider/provider.go
```diff
@@ -137,6 +137,20 @@ type InterfaceStatus struct {
...
+// LLDPAdjacency represents information about a directly connected neighbor on an interface, as discovered through LLDP.
+type LLDPAdjacency struct {
+	SysName         string
+	SysDescription  string
+	ChassisID       string
+	ChassisIDType   uint8
+	PortID          string
+	PortIDType      uint8
+	PortDescription string
+	TTL             uint32
```
felix-kaestner: Can we make this typed as `time.Duration`? Otherwise it's not clear from this type, that is represents seconds. But instead of calling it `TTLSeconds` it would rather just directly go with the `time.Duration` type here.

api/core/v1alpha1/interface_types.go
```diff
@@ -381,8 +381,49 @@ type InterfaceStatus struct {
...
+	// Validation indicates whether the LLDP neighbor information matches the information in the label or annotations of the interface.
+	Validation NeighborValidation `json:"validation"`
+}
+
+// NeighborInfo represents the information retrieved via LLDP.
+type NeighborInfo struct {
+	SystemName        string `json:"systemName,omitempty"`
+	SystemDescription string `json:"systemDescription,omitempty"`
+	ChassisID         string `json:"chassisId,omitempty"`
+	ChassisIDType     uint8  `json:"chassisIdType,omitempty"`
```
felix-kaestner: By kubernetes conventions, we shouldn't use unsigned integers in the api types. Also, I think just expose raw integers without a semantic meaning might be rather confusing. Could we at least add a comment explaining these?

api/core/v1alpha1/interface_types.go
nikatza: The intention is to have clarity about which fields are computed and which ones are just data holders. Since this is not clear I will put everything together as you suggest.

internal/controller/core/interface_controller.go
nikatza: I would prefer to use the current approach. I think would be easier to maintain in the long run since they are just a few lines of code.

felix-kaestner: Sure, just be aware that there will always be two reconciles and as such two calls to the device, when the TTL updates, as it will then trigger a new reconciliation, once the TTL in the status changes.

internal/controller/core/interface_lldp_test.go
nikatza: No, fixed.

internal/controller/core/lldp_controller.go
nikatza: This was introduced after i started working on and it slipped in, thanks for spotting it.

internal/controller/core/interface_controller.go
nikatza: I think this is a valid concern, specially considering that this affects the `interface` resource, the default initial TTL values are typically maximum 5 min, and we have several dozens of interfaces per device. I am going to adjust it.

nikatza: This has been now addressed differently. Relevant code is 364153023383989d3d3eeb60da6514059081cc4b.

internal/controller/core/interface_controller.go
```diff
@@ -417,6 +418,46 @@ func (r *InterfaceReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
...
+	predicate.Funcs
+}
+
+// Update implements predicate.Predicate.
+func (interfaceUpdatePredicate) Update(e event.UpdateEvent) bool {
+	// Generation changes on any spec or metadata update — always reconcile.
+	if e.ObjectOld.GetGeneration() != e.ObjectNew.GetGeneration() {
+		return true
+	}
+	// Only status changed. Skip if the sole difference is neighbor ExpirationTime.
```
felix-kaestner: nit: we could also have something like this to really only focus on the expiry time differences here.

```diff
diff --git a/internal/controller/core/interface_controller.go b/internal/controller/core/interface_controller.go
index 26dad652..8c2ed7c9 100644
--- a/internal/controller/core/interface_controller.go
+++ b/internal/controller/core/interface_controller.go
@@ -290,7 +290,7 @@ func (r *InterfaceReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 	}
 
 	bldr := ctrl.NewControllerManagedBy(mgr).
-		For(&v1alpha1.Interface{}, builder.WithPredicates(interfaceUpdatePredicate{})).
+		For(&v1alpha1.Interface{}, builder.WithPredicates(predicate.Or(predicate.GenerationChangedPredicate{}, interfaceUpdatePredicate{}))).
 		Named("interface").
 		WithEventFilter(filter)
```

felix-kaestner (approved):

PR #293, opened 2026-04-10
feat: introduce subinterface interface type

sven-rosenzweig:
Add a new "subinterface" interface type to support Layer 3 subinterfaces. Layer 3 subinterfaces allow a single physical router interface to handle traffic from multiple VLANs by creating separate logical interfaces per VLAN.

Validation rules:
- Encapsulation is allowed for subinterfaces only
- Supported encapsulation types: 802.1Q and 802.1ad
- Validate VLAN ID range
- 802.1ad requires two VLAN IDs (outer and inner VLAN)
- Subinterfaces cannot be configured as aggregate, switchport, or vlanRef


api/core/v1alpha1/interface_types.go
```diff
@@ -104,6 +107,15 @@ type InterfaceSpec struct {
 	// When omitted, ethernet parameters use their default values (e.g., FEC mode defaults to auto).
 	// +optional
 	Ethernet *Ethernet `json:"ethernet,omitempty"`
+
+	// Encapsulation defines the subinterfaces config for an L3 interface.
+	// +optional
+	Encapsulation *Encapsulation `json:"encapsulation,omitempty"`
+
+	// ParentInterfaceRef is a reference to the parent physical interface for this subinterface.
```
felix-kaestner:
```suggestion
	// ParentInterfaceRef is a reference to the parent interface for this subinterface.
```
I think also aggregate/bundle interfaces can have subinterfaces. So this doesn't only apply to physical interfaces.

sven-rosenzweig: Yes, you are absolutely right.

api/core/v1alpha1/interface_types.go
```diff
@@ -132,6 +144,8 @@ const (
 	InterfaceTypeAggregate InterfaceType = "Aggregate"
 	// InterfaceTypeRoutedVLAN indicates that the interface is a routed VLAN interface (SVI/IRB).
 	InterfaceTypeRoutedVLAN InterfaceType = "RoutedVLAN"
+	// InterfaceTypeSubinterface indicates that the interface is a subinterface of a physical interface.
```
felix-kaestner:
```suggestion
	// InterfaceTypeSubinterface indicates that the interface is a subinterface of a interface.
```
Can also be of a aggregate interface.

felix-kaestner: @sven-rosenzweig 🔝

sven-rosenzweig: Thanks for the finding. Sorry I missed that ....

api/core/v1alpha1/interface_types.go
```diff
@@ -165,6 +179,42 @@ type Switchport struct {
...
+const (
+	// EncapsulationTypeDot1Q indicates IEEE 802.1Q encapsulation.
+	EncapsulationTypeDot1Q EncapType = "802.1q"
+	// EncapsulationTypeQinQ indicates IEEE 802.1ad encapsulation.
+	EncapsulationTypeQinQ EncapType = "802.1ad"
+)
+
+// Encapsulation defines config for an L3 subinterface.
+// +kubebuilder:validation:XValidation:rule="self.type != '802.1q' || has(self.tag)", message="tag must be specified for interfaces of type 802.1q"
+// +kubebuilder:validation:XValidation:rule="self.type != '802.1ad' || (has(self.innerTag) && has(self.outerTag))", message="innerTag and outerTag must be specified for interfaces of type 802.1ad"
```
felix-kaestner, line 195: I think we are missing a rule here to disallow 'tag' for 802.1ad and similarly disallow 'innerTag' and 'outerTag' for 802.1q.

sven-rosenzweig: Added them. My first intention was to leave them out as I personally do not like the stack of Validation rules. But from a userperspective, I guess this make sense.

felix-kaestner (approved):

PR #302, opened 2026-04-15
Add vrfRef to BGP and propagate VRF through BGPPeer controller

felix-kaestner:
BGP.Spec gains an optional VrfRef field that scopes the BGP instance to a VRF. Both the BGP and BGPPeer controllers resolve this reference and pass the VRF to all provider calls.

The BGPPeer controller gains a reconcileVRF helper that resolves BGP.Spec.VrfRef and sets ConfiguredCondition on failure; it returns nil when unset so the provider uses the default VRF.

A vrfToBGPPeers mapper and VRF watcher (create/delete only) enqueue affected peers when their BGP's VRF definition changes. The BGP watcher predicate is extended to trigger on any ready state change and when VrfRef changes.

### DIFF
--- a/internal/controller/core/bgp_controller.go
+++ b/internal/controller/core/bgp_controller.go
@@ -36,6 +36,9 @@ import (
 	"github.com/ironcore-dev/network-operator/internal/resourcelock"
 )
 
+// bgpVrfRefIndexKey is the field index key for BGP.Spec.VrfRef.Name.
+const bgpVrfRefIndexKey = ".spec.vrfRef.name"
+
 // BGPReconciler reconciles a BGP object
 type BGPReconciler struct {
 	client.Client
@@ -62,6 +65,7 @@ type BGPReconciler struct {
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=bgp,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=bgp/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=bgp/finalizers,verbs=update
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=vrfs,verbs=get;list;watch
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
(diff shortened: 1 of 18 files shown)


internal/controller/core/bgp_controller.go
```diff
@@ -309,6 +334,15 @@ func (r *BGPReconciler) reconcile(ctx context.Context, s *bgpScope) (reterr erro
 }
 
 func (r *BGPReconciler) finalize(ctx context.Context, s *bgpScope) (reterr error) {
+	var vrf *v1alpha1.VRF
+	if s.BGP.Spec.VrfRef != nil {
+		var err error
+		vrf, err = r.reconcileVRF(ctx, s.BGP, s.Device)
```
nikatza: Claude made me aware that this function can return a `TerminalError`.

If a user deletes the VRF resource manually the BGP goes into `WaitingForDependencies`. If the user might then tries to delete the BGP resource the `reconcileVRF` in the finalizer returns `TerminalError`. This leads to the resource deletion being stucked. 

I don't think we can directly reuse `reconcileVRF` here unfortunately.

internal/controller/core/bgp_controller.go
```diff
@@ -309,6 +347,21 @@ func (r *BGPReconciler) reconcile(ctx context.Context, s *bgpScope) (reterr erro
...
 func (r *BGPReconciler) finalize(ctx context.Context, s *bgpScope) (reterr error) {
+	var vrf *v1alpha1.VRF
+	if s.BGP.Spec.VrfRef != nil {
+		vrf = new(v1alpha1.VRF)
+		if err := r.Get(ctx, types.NamespacedName{
+			Name:      s.BGP.Spec.VrfRef.Name,
+			Namespace: s.BGP.Namespace,
+		}, vrf); err != nil {
+			// If the VRF is not found, we can assume it was deleted and with it the BGP reference, so we can proceed with deletion without error.
+			return client.IgnoreNotFound(err)
```
swagner-de, line 358: Should we issue a delete request regardless and let the provider decide if it acts on it? After all we don't know if a deletion of the VRF cascades on any platform.

swagner-de (commented):
https://github.com/ironcore-dev/network-operator/blob/vrfref/internal/provider/cisco/nxos/provider.go#L382-L384

Shouldn't we check here, that the BGPs VRF is empty? Otherwise we would delete the BGP process whenever we delete a VRFs BGP?

swagner-de (changes requested):

felix-kaestner:
@swagner-de I've update the deletion procedure with the following changes:

  - `EnsureBGP`: cerror if the device already has a BGP instance with a different ASN or a BGP domain with a different router ID, rather than silently overwriting
  - `EnsureBGP`: write a marker peer group ("__operator-managed__") into every BGP domain the operator configures, to track ownership
  - `DeleteBGP`: delete only the VRF-scoped BGP domain rather than the entire BGP instance
  - `deleteBGP` **(new)**: after deleting the named bgp domain, scan remaining domains — if none carry the ownership marker and none have any SAFIs or peer groups configured, delete the global BGP instance (System/bgp-items/inst-items) as well; this cleans up on last-CR-deletion without assuming full operator ownership over the device's BGP config
  - `DeleteVRF`: co-delete the VRF's BGP domain via deleteBGPDom, since NX-OS does not remove it automatically when the VRF is deleted

internal/controller/core/bgp_controller.go
felix-kaestner: Changed this. Also please see [my comment](https://github.com/ironcore-dev/network-operator/pull/302#issuecomment-4305704795) below.

internal/controller/core/bgp_controller.go
felix-kaestner: Please see [my comment](https://github.com/ironcore-dev/network-operator/pull/302#issuecomment-4305704795) below for an updated approach (as discussed offline).

swagner-de (approved):

PR #303, opened 2026-04-15
Add routing policy refs to BGPPeerAddressFamily

felix-kaestner:
BGPPeerAddressFamily gains two optional fields — InboundRoutingPolicyRef and OutboundRoutingPolicyRef — referencing a RoutingPolicy resource to apply per address family for inbound and outbound route filtering.

The NX-OS provider maps the resolved policy names to RtCtrlP-list entries under each peer address family item, using a new RtCtrlDirection type and BGPPeerAfRtCtrlP struct following the existing pattern.


api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -118,6 +118,16 @@ type BGPPeerAddressFamily struct {
 	// for this specific address family. Defaults to false.
 	// +optional
 	RouteReflectorClient bool `json:"routeReflectorClient,omitempty"`
+
+	// InboundRoutingPolicyRef references a RoutingPolicy applied to routes received from this peer
+	// for this address family.
+	// +optional
+	InboundRoutingPolicyRef *LocalObjectReference `json:"inboundRoutingPolicyRef,omitempty"`
```
nikatza, line 125: It took me some time to find the OpenConfig counterpart — I believe it's `/network-instances/network-instance/protocols/protocol/bgp/neighbors/neighbor/afi-safis/afi-safi/apply-policy/config/import-policy`. Can you confirm?

If so, may be we could add the link to the commit message. OpenConfig also defines a `default-import-policy` / `default-export-policy` (defaulting to `REJECT_ROUTE`). Do we need something similar here? If not, it might be worth documenting the expected behavior when no routing policy is referenced (e.g., accept all / reject all).

nikatza (approved):

api/core/v1alpha1/bgp_peer_types.go
felix-kaestner: Yes, that is correct. I opted to not add a default policy options, even if Openconfig offers them, we currently don't have the need on NX-OS, so I opted to leave that out. Will update the commit description, thanks.

PR #305, opened 2026-04-15
Add redistribute direct support to BGP address-family

felix-kaestner:
Adds RedistributeDirectRoutes to BGPAddressFamily in the API, allowing operators to configure redistribution of directly connected routes via a RoutingPolicy reference. Covers both IPv4 and IPv6 unicast families.

The NX-OS provider maps this to InterLeakP entries in the BGP domain address-family DME model. The BGP controller resolves the referenced RoutingPolicy objects and watches for their creation/deletion to trigger re-reconciliation.


swagner-de (approved):
LGTM

PR #318, opened 2026-04-20
Use Option[string] for description and drop mtu omitempty

felix-kaestner:
Description is omitted completely in the YANG payload when unset or removed with 'no description', the field must be omitted entirely. Model this with Option[string] so the zero value removes the entry in YANG.

MTU is always present in the YANG payload (defaulting to 1500 when not explicitly configured), so remove the omitempty tag to ensure it is never omitted from the serialized payload.

### DIFF
--- a/internal/provider/cisco/nxos/intf.go
+++ b/internal/provider/cisco/nxos/intf.go
@@ -35,12 +35,12 @@ var (
 	_ gnmiext.DataElement = (*FabricFwdIf)(nil)
 )
 
-// Loopback represents a loopback interface on a NX-OS device.
+// Loopback represents a loopback interface.
 type Loopback struct {
-	AdminSt       AdminSt2   `json:"adminSt"`
-	Descr         string     `json:"descr"`
-	ID            string     `json:"id"`
-	RtvrfMbrItems *VrfMember `json:"rtvrfMbr-items,omitempty"`
+	AdminSt       AdminSt2       `json:"adminSt"`
+	Descr         Option[string] `json:"descr"`
+	ID            string         `json:"id"`
+	RtvrfMbrItems *VrfMember     `json:"rtvrfMbr-items,omitempty"`
 }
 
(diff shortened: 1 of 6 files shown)


internal/provider/cisco/nxos/intf.go
```diff
@@ -64,15 +65,15 @@ const (
...
 type PhysIf struct {
 	AccessVlan    string         `json:"accessVlan"`
 	AdminSt       AdminSt2       `json:"adminSt,omitempty"`
-	Descr         string         `json:"descr"`
+	Descr         Option[string] `json:"descr"`
 	FecMode       FecMode        `json:"FECMode"`
 	ID            string         `json:"id"`
 	Layer         Layer          `json:"layer"`
-	MTU           int32          `json:"mtu,omitempty"`
+	MTU           int32          `json:"mtu"`
```
nikatza, line 76: The `interface` API we declares the MTU as optional and the commit message says that it must be included. Couldn't we use here `Option[]` as well?

felix-kaestner: No. The "mtu" is always set in the YANG payload.

The Option[T] type is meant for cases, where a field is omitted completely from the YANG payload. Its functionality is a) when a value is present, compare against desired value and b) when absent, enforce field is removed from YANG entirely, by sending a DME_UNSET_PROPERTY_MARKER.

The "mtu" field however has a default value of 1500 and is always present in YANG payload. Our implementation therefore correctly defaults to a value of 1500 if not specified in the spec and otherwise sets a desired value from the spec. As such the "mtu" field is never left empty/unset, and omitempty therefore also doesn't make sense.

felix-kaestner: For the "descr" you may have a look at executing "no description" under the interface configuration and then inspecting the YANG config. You'll see that the "descr" is not there at all. This signifies the use of Option[T].

nikatza (approved):

PR #320, opened 2026-04-20
Makefile: set containerd registry config_path for kind local registry

nikatza:
Patch the containerd config to set config_path to /etc/containerd/certs.d, enabling containerd to discover the hosts.toml files created in step 3. Without this, containerd inside the kind node cannot resolve localhost:<port> to the kind-registry container.


hack/kind-with-registry.sh
```diff
@@ -31,6 +31,10 @@ nodes:
 - role: control-plane
   labels:
     topology.kubernetes.io/zone: eu-de-1a
+containerdConfigPatches:
```
felix-kaestner, line 34: @nikatza which version of kind are you running? In https://kind.sigs.k8s.io/docs/user/local-registry they mentioned that:

> NOTE: the containerd config patch is not necessary with images from kind v0.27.0+
 It may enable some older images to work similarly.

So I though these changes would not be necessary. Are they?

nikatza: They aren't, indeed. Closing this PR as this is not needed.

PR #321, opened 2026-04-21
feat: implement config backup controller

elinalin:
## Summary

This PR introduces the `ConfigBackup` controller for on-device configuration backups.

The controller supports:
- `type: Local` for timestamped backups written to device-local storage
- `type: Startup` for persisting the running configuration as `startup-config`

The initial provider implementation targets Cisco NX-OS.

## What Is Included

- Added the `ConfigBackup` CRD with spec and status
- Implemented the `ConfigBackup` controller reconcile flow
- Added provider interfaces for config backup operations
- Implemented NX-OS provider support for:
  - local filesystem backups
  - startup-config backups via NX-API
  - backup inventory discovery
  - storage statistics collection
  - retention-based rotation for local backups
- Added status handling, conditions, metrics, and event recording
- Added documentation and usage examples
- Added a reproducible local E2E example for startup backup validation

## Validation Performed

I validated this change at multiple levels.

### 1. Controller and provider tests

I ran the controller and provider-side tests to validate:
- reconciliation flow
- scheduling behavior
- retention/rotation behavior
- storage-threshold handling
- provider integration
- startup and local backup logic

### 2. Live device validation for local backup behavior

On the lab device, I verified the local backup path directly:
- creating a backup file on device-local storage
- listing the backup from the device filesystem
- deleting the backup successfully

This validated the device-side behavior for `type: Local`.

### 3. Full local E2E validation for startup backup

I also validated the full local end-to-end flow for `type: Startup`:

- ran the operator locally
- used a local kind cluster
- applied Kubernetes resources including:
  - `Device`
  - `ConfigBackupConfig`
  - `ConfigBackup`
- forwarded gNMI and NX-API ports from the lab device to the local machine
- verified that:
  - `Device` reconciled to `Running` / `Reachable=True`
  - `ConfigBackup` reconciled to `Ready=True`
  - `status.lastBackup.location` was set to `startup-config`
  - a `BackupCompleted` event was recorded
- read back the device startup-config over NX-API and confirmed the saved timestamp

This addresses the previously missing local operator E2E validation for the startup backup path.

## Notes

- For NX-OS, `type: Startup` requires NX-API because gNMI cannot execute `copy running-config startup-config`.
- To support local E2E validation where gNMI and NX-API are exposed on different ports, this PR adds an NX-OS-specific `ConfigBackupConfig` that can override the NX-API address for `ConfigBackup`.

## Example E2E Assets

- `examples/configbackup-startup-local-e2e/manifest.yaml`
- `examples/configbackup-startup-local-e2e/README.md`


api/core/v1alpha1/configbackup_types.go
felix-kaestner, line 1: @elinalin how did you scaffold this file? These api definitions are created by running `kubebuilder create api` which also generates all necessary files. ref/ https://book.kubebuilder.io/cronjob-tutorial/new-api.html

The following are missing.

1. Addition in `PROJECT`

```diff
diff --git a/PROJECT b/PROJECT
index 24face92..b8ed63fc 100644
--- a/PROJECT
+++ b/PROJECT
@@ -289,4 +289,12 @@ resources:
   kind: DHCPRelay
   path: github.com/ironcore-dev/network-operator/api/core/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: networking.metal.ironcore.dev
+  kind: ConfigBackup
+  path: github.com/ironcore-dev/network-operator/api/core/v1alpha1
+  version: v1alpha1
 version: "3"
```

2. Addition in `config/crd/kustomization.yaml`

```diff
diff --git a/config/crd/kustomization.yaml b/config/crd/kustomization.yaml
index 182f057f..2fde877a 100644
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -26,6 +26,7 @@ resources:
 - bases/networking.metal.ironcore.dev_vlans.yaml
 - bases/networking.metal.ironcore.dev_vrfs.yaml
 - bases/networking.metal.ironcore.dev_lldps.yaml
+- bases/networking.metal.ironcore.dev_configbackups.yaml
 - bases/nx.cisco.networking.metal.ironcore.dev_bordergateways.yaml
 - bases/nx.cisco.networking.metal.ironcore.dev_managementaccessconfigs.yaml
 - bases/nx.cisco.networking.metal.ironcore.dev_networkvirtualizationedgeconfigs.yaml
```

3. Addition in `config/rbac/kustomization.yaml`

```diff
diff --git a/config/rbac/kustomization.yaml b/config/rbac/kustomization.yaml
index f6aab1b9..496449b0 100644
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -37,6 +37,9 @@ resources:
 - certificate_admin_role.yaml
 - certificate_editor_role.yaml
 - certificate_viewer_role.yaml
+- configbackup_admin_role.yaml
+- configbackup_editor_role.yaml
+- configbackup_viewer_role.yaml
 - device_admin_role.yaml
 - device_editor_role.yaml
 - device_viewer_role.yaml
```

4. RBAC Helper Files

- config/rbac/configbackup_admin_role.yaml

```yaml
# This rule is not used by the project network-operator itself.
# It is provided to allow the cluster admin to help manage permissions for users.
#
# Grants full permissions ('*') over networking.metal.ironcore.dev.
# This role is intended for users authorized to modify roles and bindings within the cluster,
# enabling them to delegate specific permissions to other users or groups as needed.

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/name: network-operator
    app.kubernetes.io/managed-by: kustomize
  name: configbackup-admin-role
rules:
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups
  verbs:
  - '*'
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups/status
  verbs:
  - get
```

- config/rbac/configbackup_editor_role.yaml

```yaml
# This rule is not used by the project network-operator itself.
# It is provided to allow the cluster admin to help manage permissions for users.
#
# Grants permissions to create, update, and delete resources within the networking.metal.ironcore.dev.
# This role is intended for users who need to manage these resources
# but should not control RBAC or manage permissions for others.

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/name: network-operator
    app.kubernetes.io/managed-by: kustomize
  name: configbackup-editor-role
rules:
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups
  verbs:
  - create
  - delete
  - get
  - list
  - patch
  - update
  - watch
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups/status
  verbs:
  - get
```

- config/rbac/configbackup_viewer_role.yaml

```yaml# This rule is not used by the project network-operator itself.
# It is provided to allow the cluster admin to help manage permissions for users.
#
# Grants read-only access to networking.metal.ironcore.dev resources.
# This role is intended for users who need visibility into these resources
# without permissions to modify them. It is ideal for monitoring purposes and limited-access viewing.

apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  labels:
    app.kubernetes.io/name: network-operator
    app.kubernetes.io/managed-by: kustomize
  name: configbackup-viewer-role
rules:
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups
  verbs:
  - get
  - list
  - watch
- apiGroups:
  - networking.metal.ironcore.dev
  resources:
  - configbackups/status
  verbs:
  - get
```

felix-kaestner (changes requested):
@elinalin did you even test your changes at all? If so, how? When I run the operator with these changes I get a clear

```
ConfigBackup.networking.metal.ironcore.dev", "error": "no matches for kind \"ConfigBackup\" in version \"networking.metal.ironcore.dev/v1alpha1\""
```

Meaning that the operator won't even start or doesn't know this api type. Again, because the scaffolding from kubebuilder is missing.

api/core/v1alpha1/configbackup_types.go
```diff
@@ -0,0 +1,227 @@
...
+
+const (
+	// ConfigBackupTypeLocal stores the running configuration in a device-local file path.
+	ConfigBackupTypeLocal ConfigBackupType = "Local"
+	// ConfigBackupTypeStartup stores the running configuration as the device startup configuration.
+	ConfigBackupTypeStartup ConfigBackupType = "Startup"
+)
+
+// ConfigBackupSpec defines the desired state of ConfigBackup.
+// +kubebuilder:validation:XValidation:rule="self.type == 'Startup' ? !has(self.path) || size(self.path) == 0 : has(self.path) && size(self.path) > 0",message="path must be set for Local backups and omitted for Startup backups"
```
felix-kaestner:
```suggestion
// // +kubebuilder:validation:XValidation:rule="self.type == 'Startup' && !has(self.path) || size(self.path) == 0",message="path must be omitted for Startup backups"
// +kubebuilder:validation:XValidation:rule="self.type == 'Local' && has(self.path) && size(self.path) > 0",message="path must be set for Local backups"
```
I think having two validation rules makes the messages a bit better.

api/core/v1alpha1/configbackup_types.go
```diff
@@ -0,0 +1,227 @@
...
+	Type ConfigBackupType `json:"type"`
+
+	// Path is the device-local destination path for Local backups.
+	// Different providers may accept different path formats, such as bootflash:///backups/.
+	// +optional
+	Path string `json:"path,omitempty"`
+
+	// Retention configures automatic cleanup of older backups.
+	// +optional
+	Retention ConfigBackupRetention `json:"retention,omitempty"`
```
felix-kaestner: This field only applies to Local typed backups, as there is no way of storing multiple startup-configs. A validation rule should be added.

api/core/v1alpha1/configbackup_types.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// StorageThreshold defines the minimum free space that must remain before creating a new backup.
+	// +optional
+	StorageThreshold *ConfigBackupStorageThreshold `json:"storageThreshold,omitempty"`
```
felix-kaestner, line 55: Same here. Also only applies to Local typed backups.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+	WatchFilterValue string
+	Recorder         events.EventRecorder
+	Provider         provider.ProviderFunc
+	Locker           *resourcelock.ResourceLocker
+	Now              func() time.Time
+}
+
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=configbackups,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=configbackups/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=core,resources=events,verbs=create;patch
```
felix-kaestner:
```suggestion
// +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
```

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+// ConfigBackupReconciler reconciles a ConfigBackup object.
+type ConfigBackupReconciler struct {
+	client.Client
+	Scheme *runtime.Scheme
+
+	WatchFilterValue string
+	Recorder         events.EventRecorder
+	Provider         provider.ProviderFunc
+	Locker           *resourcelock.ResourceLocker
+	Now              func() time.Time
```
felix-kaestner:
```suggestion
	client.Client
	Scheme *runtime.Scheme
	
	// WatchFilterValue is the label value used to filter events prior to reconciliation.
	WatchFilterValue string

	// Recorder is used to record events for the controller.
	// More info: https://book.kubebuilder.io/reference/raising-events
	Recorder events.EventRecorder

	// Provider is the driver that will be used to create & delete the interface.
	Provider provider.ProviderFunc

	// Locker is used to synchronize operations on resources targeting the same device.
	Locker *resourcelock.ResourceLocker

	// Now is a function that returns the current time. It can be overridden in tests to control time-dependent behavior.
	Now func() time.Time
```

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+	var cfg *provider.ProviderConfig
+	if obj.Spec.ProviderConfigRef != nil {
+		cfg, err = provider.GetProviderConfig(ctx, r, obj.Namespace, obj.Spec.ProviderConfigRef)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	orig := obj.DeepCopy()
+	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition, v1alpha1.ConfiguredCondition, v1alpha1.OperationalCondition) {
```
felix-kaestner:
```suggestion
	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition) {
```
Just use the single Ready condition as standalone. There are no Configured or Operational Conditions.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+		log.V(1).Info("Initializing status conditions")
+		return ctrl.Result{}, r.Status().Update(ctx, obj)
+	}
+
+	defer func() {
+		if !equality.Semantic.DeepEqual(orig.ObjectMeta, obj.ObjectMeta) {
+			if err := r.Patch(ctx, obj.DeepCopy(), client.MergeFrom(orig)); err != nil {
```
felix-kaestner, line 178:
```suggestion
			// Pass obj.DeepCopy() to avoid Patch() modifying obj and interfering with status update below
			if err := r.Patch(ctx, obj.DeepCopy(), client.MergeFrom(orig)); err != nil {
```
nit: comment might be helpful.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+
+	s := &configBackupScope{
+		Device:         device,
+		ConfigBackup:   obj,
+		Connection:     conn,
+		ProviderConfig: cfg,
+		Provider:       prov,
+	}
+
+	return r.reconcile(ctx, s)
```
felix-kaestner:
```suggestion
	res, err := r.reconcile(ctx, s)
	if err != nil {
		log.Error(err, "Failed to reconcile resource")
		return ctrl.Result{}, err
	}

	return res, nil
```

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+	if !controllerutil.HasControllerReference(s.ConfigBackup) {
+		if err := controllerutil.SetOwnerReference(s.Device, s.ConfigBackup, r.Scheme, controllerutil.WithBlockOwnerDeletion(true)); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	defer func() {
+		conditions.RecomputeReady(s.ConfigBackup)
+	}()
+
```
felix-kaestner:
```suggestion
```
Just set the Ready condition directly and don't compute it.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+			Reason:  v1alpha1.ScheduleInvalidReason,
+			Message: err.Error(),
+		})
+		conditions.Set(s.ConfigBackup, metav1.Condition{
+			Type:    v1alpha1.OperationalCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  v1alpha1.ScheduleInvalidReason,
+			Message: "Backup schedule is invalid",
+		})
+		return ctrl.Result{}, nil
```
felix-kaestner:
```suggestion
		conditions.Set(s.ConfigBackup, metav1.Condition{
			Type:    v1alpha1.ReadyCondition,
			Status:  metav1.ConditionFalse,
			Reason:  v1alpha1.ScheduleInvalidReason,
			Message: err.Error(),
		})
		return ctrl.Result{}, nil
```

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+			Message: "Device is not reachable",
+		})
+		observeConfigBackupFailure(s.ConfigBackup.Spec.Type, 0)
+		return ctrl.Result{}, fmt.Errorf("failed to connect to provider: %w", err)
+	}
+	defer func() {
+		if err := s.Provider.Disconnect(ctx, s.Connection); err != nil {
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+	}()
```
felix-kaestner, line 320:
```suggestion
	if err := s.Provider.Connect(ctx, s.Connection); err != nil {
		return ctrl.Result{}, fmt.Errorf("failed to connect to provider: %w", err)
	}
	defer func() {
		if err := s.Provider.Disconnect(ctx, s.Connection); err != nil {
			reterr = kerrors.NewAggregate([]error{reterr, err})
		}
	}()
```

no status update here, this might be an intermittent situation, so simply return the error which will requeue and backoff-retry.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+			Message: err.Error(),
+		})
+		conditions.Set(s.ConfigBackup, metav1.Condition{
+			Type:    v1alpha1.OperationalCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  v1alpha1.ScheduleInvalidReason,
+			Message: "Backup schedule is invalid",
+		})
+		return ctrl.Result{}, nil
+	}
```
felix-kaestner:
```suggestion
	var schedule *cron.Schedule
	if s.ConfigBackup.Spec.Schedule != "" {
		sh, err := cron.ParseStandard(s.ConfigBackup.Spec.Schedule)
		// ...
		schedule = &sh
	}
```
Just write it like this instead of all these tiny helper functions. Makes the code more readable.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,676 @@
...
+		Watches(
+			&v1alpha1.Device{},
+			handler.EnqueueRequestsFromMapFunc(r.deviceToConfigBackups),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldDevice := e.ObjectOld.(*v1alpha1.Device)
+					newDevice := e.ObjectNew.(*v1alpha1.Device)
+					return oldDevice.Spec.Paused != newDevice.Spec.Paused
+				},
+				GenericFunc: func(e event.GenericEvent) bool { return false },
```
felix-kaestner:
```suggestion
				UpdateFunc: func(e event.UpdateEvent) bool {
					return paused.DevicePausedChanged(e.ObjectOld, e.ObjectNew)
				},
				GenericFunc: func(e event.GenericEvent) bool {
					return false
				},
```

elinalin:
> @elinalin did you even test your changes at all? If so, how? When I run the operator with these changes I get a clear
> 
> ```
> ConfigBackup.networking.metal.ironcore.dev", "error": "no matches for kind \"ConfigBackup\" in version \"networking.metal.ironcore.dev/v1alpha1\""
> ```
> 
> Meaning that the operator won't even start or doesn't know this api type. Again, because the scaffolding from kubebuilder is missing.

I did test the controller/provider logic, including real NX-OS Local backup behavior（the lab container) , but you are right that I missed the fresh deployment/install path.

I checked the code after your comment:
- the Go type is registered in the scheme
- the controller is wired into the manager
- the CRD base exists in the repo

So this is not missing Go-side kubebuilder scaffolding.

The real issue is that the ConfigBackup CRD is not being delivered/installed in the deployment path you used, so the API server does not recognize the kind and the operator cannot start cleanly against that cluster.

I’ll fix the install artifacts and validate again from a fresh deployment.

api/core/v1alpha1/configbackup_types.go
elinalin: Thanks Felix. This was a useful catch. I had the API type and controller wired up already, but I had indeed missed part of the Kubebuilder scaffolding around the install path, specifically the PROJECT entry, the CRD/RBAC kustomization updates, and the helper RBAC manifests. 
I‘ve added those now.

.github/workflows/lint.yml
felix-kaestner, line 1: Why was this changed?

api/cisco/nx/v1alpha1/configbackupconfig_types.go
felix-kaestner, line 1: The provider config pattern is used to add additional vendor-specific configuration to a configuration item (such as the `ConfigBackup`), which this resource is clearly not doing. At the same time, I also don't see a reason, why we would need to define another Address of the NX-API Endpoint which is distinct from the endpoint address in `.spec.endpoint.address`. So I would strongly suggest that we remove this again.

api/core/v1alpha1/configbackup_types.go
```diff
@@ -0,0 +1,227 @@
...
+	// Schedule is an optional cron expression. If omitted, the controller performs a one-shot backup per generation.
+	// +optional
+	Schedule string `json:"schedule,omitempty"`
+
+	// Type determines whether the backup is saved as a local file or as startup-config.
+	// +required
+	Type ConfigBackupType `json:"type"`
+
+	// Path is the device-local destination path for Local backups.
+	// Different providers may accept different path formats, such as bootflash:///backups/.
```
felix-kaestner:
```suggestion
	// Different providers may accept different path formats, such as "bootflash:///backups/".
```

api/core/v1alpha1/groupversion_info.go
```diff
@@ -184,6 +184,21 @@ const (
...
+	BackupPendingReason = "BackupPending"
+
+	// BackupFailedReason indicates that a backup operation failed.
+	BackupFailedReason = "BackupFailed"
+
+	// ScheduleInvalidReason indicates that the configured cron schedule is invalid.
+	ScheduleInvalidReason = "ScheduleInvalid"
+
+	// StorageThresholdExceededReason indicates that a configured storage safety threshold was exceeded.
+	StorageThresholdExceededReason = "StorageThresholdExceeded"
```
felix-kaestner, line 203: Maybe add a new section for this specific reasons that only apply to `ConfigBackup` resources (see below in the file).

config/crd/kustomization.yaml
felix-kaestner, line 1: Please keep the formatting in this file the same and only add lines.

config/default/kustomization.yaml
felix-kaestner, line 1: Why were all these files changed in this file?

config/rbac/kustomization.yaml
felix-kaestner, line 1: Please don't reformat this file

docs/concepts/config-backup.md
```diff
@@ -0,0 +1,66 @@
...
+- Status reporting for last backup result, next scheduled backup, and discovered backup inventory
+- Best-effort checksum reporting when the provider can retrieve it
+
+## Provider Scope
+
+The initial implementation targets Cisco NX-OS devices.
+
+- `type: Startup` uses NX-API because NX-OS does not allow a local config save via gNMI
+- `type: Local` stores timestamped managed backup files on device-local storage
+- File inventory and storage statistics are collected from the device and surfaced through `status`
```
felix-kaestner: I would leave out the provider implementation specifics in this page. The documentation here should just discuss the concepts and stay vendor-agnostic.

docs/concepts/config-backup.md
```diff
@@ -0,0 +1,66 @@
...
+    name: leaf-switch-1
+  type: Startup
+```
+
+## Notes
+
+- `Startup` backups always keep a single logical copy.
+- Local backup rotation only applies to `type: Local`.
+- `checksum` and `sizeBytes` are optional status fields and depend on what the provider can retrieve from the device.
+- Backup execution depends on provider support. Providers that do not implement `ConfigBackup` report a `NotImplemented` status.
```
felix-kaestner: This is true for all resources. As such, I would leave it out here.

examples/configbackup-startup-local-e2e/manifest.yaml
felix-kaestner, line 1: We already have the samples in `config/samples` doing exactly the same thing. So we don't need to add it again here.

examples/configbackup-startup-local-e2e/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+This example expects:
+
+- gNMI on `127.0.0.1:9339`
+- NX-API on `127.0.0.1:8443`
+
+For the current shared lab device, this was validated with:
+
+```bash
+kubectl --context clabernetes -n lab--i583364-confirmedcommits port-forward svc/cisco-nx-10-6-2-router 8443:443 9339:9339
```
felix-kaestner: Please leave out internal development setup/landscapes from this public open source repo.

Also, there is a Tiltfile located in this project, specifically designed for local development. So overall, this is not needed to have in the repo.

internal/controller/core/suite_test.go
```diff
@@ -87,14 +91,16 @@ var _ = BeforeSuite(func() {
 	Expect(cfg).NotTo(BeNil())
 
 	k8sManager, err = ctrl.NewManager(cfg, ctrl.Options{
-		Scheme: scheme.Scheme,
-		Logger: GinkgoLogr,
+		Scheme:  scheme.Scheme,
+		Logger:  GinkgoLogr,
+		Metrics: metricsserver.Options{BindAddress: "0"},
```
felix-kaestner, line 95: Why do we intentionally disable the metrics server here?

internal/controller/core/suite_test.go
```diff
@@ -459,6 +496,9 @@ func NewProvider() *Provider {
 		EVIs:            sets.New[int32](),
 		PrefixSets:      sets.New[string](),
 		RoutingPolicies: sets.New[string](),
+		StorageTotal:    func() *int64 { v := int64(1024 * 1024 * 100); return &v }(),
+		StorageUsed:     func() *int64 { v := int64(0); return &v }(),
+		StorageFree:     func() *int64 { v := int64(1024 * 1024 * 100); return &v }(),
```
felix-kaestner:
```suggestion
		StorageTotal:    new(int64(1024 * 1024 * 100)), // 100 MiB total storage
		StorageUsed:     new(int64(0)),                 // 0 bytes used
		StorageFree:     new(int64(1024 * 1024 * 100)), // 100 MiB free
```

internal/controller/core/suite_test.go
```diff
@@ -442,6 +475,10 @@ type Provider struct {
 	NVE             *v1alpha1.NetworkVirtualizationEdge
 	LLDP            *v1alpha1.LLDP
 	DHCPRelay       *v1alpha1.DHCPRelay
+	ConfigBackups   []provider.ConfigBackupFile
+	StorageTotal    *int64
+	StorageUsed     *int64
+	StorageFree     *int64
```
felix-kaestner: Why are these pointers if they are always set?

internal/controller/core/suite_test.go
```diff
@@ -888,6 +926,121 @@ func (p *Provider) GetNVEStatus(_ context.Context, _ *provider.NVERequest) (prov
...
+		if !remove.Has(file.Name) {
+			filtered = append(filtered, file)
+		}
+	}
+	p.ConfigBackups = filtered
+	p.recalculateStorage()
+	return nil
+}
+
+func (p *Provider) recalculateStorage() {
```
felix-kaestner: Instead of making this a function that must be called and then sets fields on the `(p *Provider)`, just make it a Getter function

```go
func (p *Provider) FreeStorageBytes() uint64 { ... }
func (p *Provider) UsedStorageBytes() uint64 { ... }
```
This makes everything a lot simpler.

internal/controller/core/configbackup_controller_test.go
```diff
@@ -0,0 +1,294 @@
...
+				g.Expect(current.Status.LastBackup).NotTo(BeNil())
+				g.Expect(current.Status.LastBackup.Checksum).To(Equal("sha256:test-checksum"))
+				g.Expect(current.Status.LastBackup.Location).To(ContainSubstring("bootflash:///backups/"))
+				g.Expect(current.Status.TotalBackups).To(Equal(int32(1)))
+				cond := meta.FindStatusCondition(current.Status.Conditions, v1alpha1.ReadyCondition)
+				g.Expect(cond).NotTo(BeNil())
+				g.Expect(cond.Status).To(Equal(metav1.ConditionTrue))
+			}).Should(Succeed())
+
+			Eventually(recordedEvents).Should(ContainElement(ContainSubstring("BackupCompleted")))
```
felix-kaestner: I've already said this in the other PR. Please don't assert on a specific event. Assert on the status of the resource.

Please also remove all the added Event Recorder Helpers again.

internal/provider/cisco/nxos/configbackup.go
```diff
@@ -0,0 +1,378 @@
...
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/provider"
+	"github.com/ironcore-dev/network-operator/internal/transport/nxapi"
+)
+
+var _ provider.ConfigBackupProvider = (*Provider)(nil)
+
+type nxapiDoer interface {
+	Do(context.Context, nxapi.Request) ([]json.RawMessage, error)
+}
```
felix-kaestner: Why does that even exist? Just use the NX-API client in `p.nxapi`.

internal/provider/cisco/nxos/configbackup.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+	base := req.ConfigBackup.Spec.Path
+	if strings.HasSuffix(base, "/") {
+		return base + req.TargetName
+	}
+	return base + "/" + req.TargetName
+}
+
+func parseConfigBackupInventory(raw json.RawMessage, req *provider.ConfigBackupRequest) (*provider.ConfigBackupInventory, error) {
+	var decoded any
```
felix-kaestner: You get a perfectly formatted json in the response. Why don't you just decode that to a Go Struct instead of all this `any` type checking/walking?

internal/provider/configbackup.go
```diff
@@ -0,0 +1,60 @@
...
+	// ListConfigBackups lists the backups currently discovered for the ConfigBackup policy.
+	ListConfigBackups(context.Context, *ConfigBackupRequest) (*ConfigBackupInventory, error)
+	// DeleteConfigBackups removes the provided backups from the device.
+	DeleteConfigBackups(context.Context, *DeleteConfigBackupsRequest) error
+}
+
+type ConfigBackupRequest struct {
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+	ManagedPrefix  string
```
felix-kaestner: This field is not used for anything. Why do we have it?

internal/provider/configbackup.go
```diff
@@ -0,0 +1,60 @@
...
+	TargetName     string
```
felix-kaestner: This `TargetName` doesn't really make sense for me. The `ConfigBackup` already contains the desired config backup. Also, having a implicit magic `"startup-config"` string as this value is certainly a bad abstraction. Just use the `ConfigBackup` resource directly in the NX-OS provider implementation.

internal/provider/configbackup.go
```diff
@@ -0,0 +1,60 @@
...
+type DeleteConfigBackupsRequest struct {
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+	Backups        []ConfigBackupFile
+}
+
+type ConfigBackupResult struct {
+	Name      string
+	Path      string
+	SizeBytes *int64
```
felix-kaestner: If we currently can't support this field on NX-OS (which is our main implementation currently) lets leave this out of the API completely. We can always add things later, if we which to. (Same also goes for the `*ConfigBackupFile`)

internal/provider/configbackup.go
```diff
@@ -0,0 +1,60 @@
...
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+	ManagedPrefix  string
+	TargetName     string
+}
+
+type DeleteConfigBackupsRequest struct {
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+	Backups        []ConfigBackupFile
```
felix-kaestner: Why does this get a list of backups? The provider implementation should take care of formulating the requested `*v1alpha1.ConfigBackup` into specific delete operations performed on the provider.

internal/provider/configbackup.go
```diff
@@ -0,0 +1,60 @@
...
+	Checksum  string
+	CreatedAt time.Time
+	Duration  time.Duration
+}
+
+type ConfigBackupInventory struct {
+	Backups    []ConfigBackupFile
+	TotalBytes *int64
+	UsedBytes  *int64
+	FreeBytes  *int64
```
felix-kaestner:
```suggestion
	TotalBytes int64
	UsedBytes  int64
	FreeBytes  int64
```
don't see a reason why these should be pointers.

internal/provider/configbackup.go
felix-kaestner, line 1: Why was this moved to it's own file instead of being put in the `internal/provider/provider.go` file like all others?

.golangci.yaml
felix-kaestner, line 1: Why did you remove this? Just please revert these changes.

felix-kaestner (changes requested):

internal/controller/core/configbackup_controller.go
elinalin: Throughout this controller, Configured and Operational are used as independent conditions, with Ready recomputed based on them. Adjusting only the initialization would make the condition model inconsistent.
Adopting a Ready-only model would therefore require a more comprehensive refactor of the controller, rather than a simple one-line change.

internal/controller/core/configbackup_controller.go
elinalin: In the current controller design, Configured and Operational are maintained as standalone child conditions during reconciliation, and Ready is derived from them afterwards.

internal/controller/core/configbackup_controller.go
elinalin: These two messages are not intended to represent the exact same thing: Configured=False carries the detailed validation/parsing error, while Operational=False is meant to provide a higher-level status summary.

internal/controller/core/configbackup_controller.go
elinalin: Agree. Fixed.

.github/workflows/lint.yml
elinalin: Reverted

api/cisco/nx/v1alpha1/configbackupconfig_types.go
elinalin: Thank you Felix. Removed.

config/default/kustomization.yaml
elinalin: my vscode setting did the yaml format, i have reverted those files.

docs/concepts/config-backup.md
elinalin: Agree. I removed the NX-OS / NX-API specific notes and kept the documentation focused on the resource behavior and API semantics.

examples/configbackup-startup-local-e2e/manifest.yaml
elinalin: Removed

internal/controller/core/suite_test.go
elinalin: Fixed

internal/controller/core/suite_test.go
elinalin: I applied the readability cleanup, but used a small int64Ptr(...) helper instead of new(int64(...)) since the latter is not valid Go syntax.

internal/controller/core/suite_test.go
elinalin: These stay as pointers intentionally because they mirror the provider contract, where storage statistics may be unavailable. In that case, nil represents “unknown / not reported” rather than 0. The test provider usually initializes them, but keeping them as pointers allows us to preserve the real API semantics and still simulate missing storage information when needed.

internal/controller/core/suite_test.go
elinalin: Agree with simplifying the fake provider here. I removed the cached StorageUsed / StorageFree fields and the recalculateStorage() mutator, and now compute the storage usage on demand via getter-style helpers when building the inventory response. I keptStorageTotal pointer-based so the fake provider can still represent “storage total unknown”, which matches the provider contract semantics better than switching everything to plain numeric getters.

internal/provider/cisco/nxos/configbackup.go
elinalin: You are right!  The response we rely on is structured enough to decode directly into Go structs. Fixed

internal/provider/configbackup.go
elinalin: Removed and stopped computing it in the controller, and now let the NX-OS provider derive the target name/path directly from the ConfigBackup resource itself.

internal/provider/configbackup.go
elinalin: Fixed

internal/provider/configbackup.go
elinalin: Removed ManagedPrefix

internal/provider/configbackup.go
elinalin: Removed the explicit backup list from DeleteConfigBackupsRequest and moved the deletion selection into the provider implementation, so the provider now derives the concrete delete operations from the ConfigBackup resource itself.

internal/provider/configbackup.go
elinalin: Keep these as pointers because nil represents “unknown / not reported by the provider”, which is distinct from a real 0 value. The controller currently relies on that distinction when computing free percent and evaluating storage thresholds.

internal/provider/configbackup.go
elinalin: Fixed

internal/controller/core/configbackup_controller.go
elinalin: Updated as suggested. I inlined the trivial schedule handling into the reconcile flow and removed the tiny helper wrappers, while keeping cron.Schedule as an interface value rather than using a pointer-to-interface

internal/provider/cisco/nxos/bgp.go
```diff
@@ -159,8 +151,8 @@ func (af *BGPDomAfItem) Key() AddressFamily { return af.Type }
 func NewInterLeakPDirect(rtMap string) *InterLeakP {
 	return &InterLeakP{
 		InterLeakPKey: InterLeakPKey{
-			Asn:   "none",
-			Inst:  "none",
+			Asn:   interLeakProtocolDefault,
```
felix-kaestner: Why was this changed? Seems unrelated to the PR.

internal/provider/cisco/nxos/name.go
```diff
@@ -7,12 +7,12 @@ import (
 	"errors"
 	"fmt"
 	"regexp"
-
-	"github.com/ironcore-dev/network-operator/internal/apistatus"
 )
 
+const managementInterfaceName = "mgmt0"
```
felix-kaestner: What is the reason for this change?

internal/provider/cisco/nxos/ospf.go
```diff
@@ -66,6 +66,8 @@ type InterLeakPKey struct {
 	Proto RtLeakProto `json:"proto"`
 }
 
+const interLeakProtocolDefault = "none"
```
felix-kaestner: Again, unrelated change to the concept of this PR.

Makefile
```diff
@@ -224,8 +224,10 @@ install-addlicense: FORCE
 
 prepare-static-check: FORCE install-goimports install-golangci-lint install-shellcheck install-typos install-go-licence-detector install-addlicense
 
+CONTROLLER_GEN_VERSION ?= v0.21.0
```
felix-kaestner: Please revert. The Makefile is currently generated by go-makefile-maker, so any manual changes not done through the Makefile.maker.yaml configuration file are overritten on the next invocation.

internal/transport/gnmiext/empty.go
```diff
@@ -9,6 +9,8 @@ import (
 	"regexp"
 )
 
+const jsonNullLiteral = "null"
```
felix-kaestner: Again, unrelated changes to the PR. 
Please don't bloat your PR with such unrelated changes.

If you want to do this. Please open a separate PR, but that shouldn't go into a "BackupConfig Controller" PR.

internal/provider/provider.go
```diff
@@ -37,29 +37,74 @@ type DeviceProvider interface {
...
+
+type ConfigBackupRequest struct {
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+}
+
+type DeleteConfigBackupsRequest struct {
+	ConfigBackup   *v1alpha1.ConfigBackup
+	ProviderConfig *ProviderConfig
+}
```
felix-kaestner: `DeleteConfigBackupsRequest` is identical to `ConfigBackupRequest`. Therefore, we can just use `ConfigBackupRequest` for the parameter of the `DeleteConfigBackups` func. No need to duplicate the same struct.

internal/provider/provider.go
```diff
@@ -37,29 +37,74 @@ type DeviceProvider interface {
...
+	CreatedAt time.Time
+	Checksum  string
+}
+
+// ConfigBackupManagedPrefix returns the filename prefix used for backups managed by a ConfigBackup.
+func ConfigBackupManagedPrefix(obj *v1alpha1.ConfigBackup) string {
+	if obj == nil {
+		return ""
+	}
+	return fmt.Sprintf("configbackup-%s-%s-", obj.Namespace, obj.Name)
```
felix-kaestner: Please just make this a method on the `*v1alpha1.ConfigBackup` type. Perhaps something like

```go
func (c *v1alpha1.ConfigBackup) Filename() string {
  return fmt.Sprintf("configbackup-%s-%s-", obj.Namespace, obj.Name)
}
```
(note: such method should have a nil check).

internal/controller/core/configbackup_controller.go
felix-kaestner: This was not change. Please don't mark as resolved, when not addressed.

internal/controller/core/configbackup_controller.go
felix-kaestner: Again, an `Operational` condition doesn't make much sense for this resource type, as there is no Operational State that is queried from the Device. So please just use a single `Ready` condition for tracking all Success/Error cases, similar to how it's done e.g. for the `AccessControlList` resource type.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,559 @@
...
+		Type:    v1alpha1.ConfiguredCondition,
+		Status:  metav1.ConditionFalse,
+		Reason:  v1alpha1.BackupPendingReason,
+		Message: "Waiting for the first scheduled backup",
+	})
+}
+
+func configBackupRequestFor(obj *v1alpha1.ConfigBackup) *provider.ConfigBackupRequest {
+	return &provider.ConfigBackupRequest{ConfigBackup: obj}
+}
```
felix-kaestner: Please avoid adding such funcs. They add very little value as a separate function and just make it harder to read the code. Just do `&provider.ConfigBackupRequest{ConfigBackup: obj}` inline. This is a oneliner.

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,559 @@
...
+
+const maxInt32 = int(^uint32(0) >> 1)
```
felix-kaestner: There is just `math.MaxInt32` in the standard library. No need to define this ourselves. Also please think about if it is realistic to have so many backups and if it is worth adding this complexity or if we should rather assume to never have more than 2147483647 backups (which I think is realistic).

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,559 @@
...
+	obj.Status.OldestBackupTimestamp = oldest
+	if obj.Spec.Type == v1alpha1.ConfigBackupTypeStartup && obj.Status.OldestBackupTimestamp == nil && obj.Status.LastBackup != nil {
+		ts := obj.Status.LastBackup.Timestamp
+		obj.Status.OldestBackupTimestamp = &ts
+	}
+
+	storage := &v1alpha1.ConfigBackupStorageStatus{
+		TotalBytes: cloneInt64Ptr(inventory.TotalBytes),
+		UsedBytes:  cloneInt64Ptr(inventory.UsedBytes),
+		FreeBytes:  cloneInt64Ptr(inventory.FreeBytes),
```
felix-kaestner: What is the reason to clone a `*in32` here? I think we can just assign it as is (no need to cloning).

internal/provider/cisco/nxos/provider.go
```diff
@@ -64,10 +63,16 @@ var (
...
 
+type nxapiClient interface {
+	Do(context.Context, nxapi.Request) ([]json.RawMessage, error)
+}
+
 type Provider struct {
-	conn   *grpc.ClientConn
-	client gnmiext.Client
-	nxapi  *nxapi.Client
+	baseConn *deviceutil.Connection
```
felix-kaestner: Why do we have this `baseConn` field? It serves no purpose.

internal/provider/provider.go
```diff
@@ -37,29 +37,74 @@ type DeviceProvider interface {
...
+	SizeBytes *int64
+	Checksum  string
+	CreatedAt time.Time
+	Duration  time.Duration
+}
+
+type ConfigBackupInventory struct {
+	Backups    []ConfigBackupFile
+	TotalBytes *int64
+	UsedBytes  *int64
```
felix-kaestner, line 111: Can't UsedBytes simply be derived from the `Backups []ConfigBackupFile`? Something like:

```go
func (c *ConfigBackupInventory) UsedBytes() int64 {
  sum := int64(0)
  for _, file := range c.Backups {
    sum += c.SizeBytes
  }
  return sum
}
```

internal/controller/core/configbackup_controller.go
```diff
@@ -0,0 +1,559 @@
...
+
+func observeConfigBackupFailure(backupType v1alpha1.ConfigBackupType, duration time.Duration) {
+	labels := prometheusLabels(backupType, "failure")
+	configBackupOperationsTotal.WithLabelValues(labels...).Inc()
+	configBackupDurationSeconds.WithLabelValues(labels...).Observe(duration.Seconds())
+}
+
+func prometheusLabels(backupType v1alpha1.ConfigBackupType, result string) []string {
+	return []string{result, string(backupType)}
+}
```
felix-kaestner, line 516: Just inline this. Having such oneliner helper functions just necessarily increases complexity for a reader to follow the code.

felix-kaestner (changes requested):

internal/provider/cisco/nxos/name.go
elinalin: Thanks for flagging this.
You are right, this change is only a refactor (extracting "mgmt0" into a constant) and does not change behavior.
It is unrelated to the ConfigBackup scope, so I reverted it to keep the PR focused.

internal/provider/cisco/nxos/bgp.go
elinalin: Reverted

internal/provider/cisco/nxos/provider.go
elinalin: Good catch, removed.

internal/provider/provider.go
elinalin: UsedBytes and Backups represent different scopes. Backups only contains files managed by this ConfigBackup policy, while UsedBytes/FreeBytes come from device storage stats and include unrelated files on the same storage.

elinalin:
@felix-kaestner Hi, I've addressed all of the review comments and updated the PR accordingly.
Could you please take another look when you have time? If everything looks good, would appreciate your approval so we can get this PR merged.
Please let me know if there's anything else I should update. Thanks!

felix-kaestner (approved):

PR #324, opened 2026-04-23
[NXOS] Enable `fabric forwarding` feature at NVE

nikatza:
NVE resources that use an Anycast Gateway require the ‘fabric forwarding’ feature on NX-OS. We now enable this feature automatically when ensuring NVE configuration.


internal/provider/cisco/nxos/provider.go
felix-kaestner, line 2821:
```suggestion
	ag := new(FabricFwd)
	ag.AdminSt = AdminStDisabled
```
@nikatza should we do this change as well for the fabric forwarding? Also the type shall be changed from `string` to `AdminSt`. Could you take care of this? 🙏

Makefile
felix-kaestner, line 1: @nikatza I guess you manually made the changes here? Just to be aware, as they will be overwritten by invoking `go-makefile-maker` next time.

felix-kaestner (approved):

PR #325, opened 2026-04-27
Add `internal/apistatus` package with typed terminal errors

felix-kaestner:
Introduces internal/apistatus, a package that gives provider implementations a structured way to return non-retryable errors to controllers without coupling directly to reconcile.TerminalError.

The package defines a Code enum (CodeInvalidArgument, CodeUnsupportedField, CodeFailedPrecondition) with String() and Valid() methods, a StatusError struct carrying a Code, an optional Message, and optional FieldViolations, and constructor functions for each code. StatusError implements Is() so that errors.Is can be used for type detection across error chains. WrapTerminalError wraps any error containing a *StatusError as a reconcile.TerminalError, leaving all other errors unchanged.

conditions.FromError is extended to detect *StatusError before the existing gRPC status path, mapping the code to the condition Reason and the formatted message to the condition Message field.

The stub internal/controllerutil package, which this supersedes, has been removed as no code imported it.

### DIFF
--- a/internal/apistatus/apistatus.go
+++ b/internal/apistatus/apistatus.go
@@ -1,7 +1,7 @@
 // SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
 // SPDX-License-Identifier: Apache-2.0
 
-// Package apistatus provides a typed terminal error for provider implementations.
+// Package apistatus provides a typed structured error for provider implementations.
 package apistatus
 
 import (
@@ -29,7 +29,10 @@ const (
 	CodeUnsupportedField
 
 	// CodeFailedPrecondition signals that a precondition on the device or
-	// environment is not met. Retrying will not resolve it.
+	// environment is not met. Unlike other codes, this is retryable — the
+	// precondition may become true on a future attempt (e.g. a BGP process
+	// configured out-of-band). [WrapTerminalError] does not promote these
+	// errors to terminal.
 	CodeFailedPrecondition
 )
 
@@ -65,9 +68,10 @@ type FieldViolation struct {
 	Description string
 }
 
-// StatusError is a terminal, non-retryable error returned by provider
-// implementations. It carries a Code, an optional top-level Message,
-// and an optional list of FieldViolations.
+// StatusError is a structured error returned by provider implementations.
+// It carries a Code, an optional top-level Message, and an optional list of
+// FieldViolations. Whether the error is terminal depends on the Code — see
+// [WrapTerminalError].
 type StatusError struct {
 	Code            Code
 	Message         string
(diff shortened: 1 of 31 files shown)


internal/apistatus/apistatus.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if len(parts) == 0 {
+		return e.Code.String()
+	}
+	return e.Code.String() + ": " + strings.Join(parts, "; ")
+}
+
+// InvalidArgument returns a [StatusError] with [CodeInvalidArgument] for
+// one or more spec field values that have an incorrect format or structure.
+func InvalidArgument(violations ...FieldViolation) *StatusError {
```
sven-rosenzweig: optional: Not sure if this violates any golang standard, but prepending a 'new' and appending "err" to the function name  e.g.``NewInvalidArgumentErr``, would make it clearer what the function is doing.

internal/apistatus/apistatus.go
```diff
@@ -0,0 +1,149 @@
...
+//
+// Example usage in a controller:
+//
+//	err := s.Provider.EnsureBGPPeer(ctx, &provider.BGPPeerRequest{...})
+//	cond := conditions.FromError(err)
+//	conditions.Set(s.BGPPeer, cond)
+//	if err != nil {
+//	    return apistatus.WrapTerminalError(err)
+//	}
+func WrapTerminalError(err error) error {
```
sven-rosenzweig, line 154: Don't we wrap grpc errors to TerminalError already? Does it make sense to combine this in one function?

internal/apistatus/apistatus.go
```diff
@@ -29,7 +29,10 @@ const (
 	CodeUnsupportedField
 
 	// CodeFailedPrecondition signals that a precondition on the device or
```
sven-rosenzweig, line 31: the idea just popped into my head, unsure if that make sense, but would it be helpful to make ``retry`` configurable? For most of the cases probably not, just wanted tho share my thoughts here.

internal/apistatus/apistatus.go
felix-kaestner: Yes, we do, but it doesn't make sense to combine this. grpc errors would be returned when we actually make a gNMI call. These apistatus errors are used for input validation, even before making any gNMI call. Hence they are also checked at different code paths.

internal/apistatus/apistatus.go
felix-kaestner: Nice idea. I think I would leave this up to future follow-ups in case we really need it. Currently, as this Error is used for the NX-OS implementation, we agreed that we are fine to keep the backup/retry behavior (making it non-terminal) so that it has the chance to eventually resolve itself.

internal/apistatus/apistatus.go
felix-kaestner: Done

sven-rosenzweig (approved):
Happy to use this package also for IOS XR. Thanks for that :).

PR #328, opened 2026-04-28
Scaffold EVPN `Fabric` Resource and Controller

felix-kaestner:
This high-level CRD will take care of creating and managing all low-level resources that are required to build a EVPN VXLAN resource.

Its controller implementation is intentionally left empty to a large extend to be extended by future changes. The setup constructs a list of sub-reconciler functions that are called in sequence to build up the full lifecycle of the `Fabric`.


api/evpn/v1alpha1/fabric_types.go
```diff
@@ -0,0 +1,296 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	corev1alpha1 "github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+)
+
+// FabricSpec defines the desired state of Fabric.
+type FabricSpec struct {
```
nikatza, line 15: I quite like this API definition, it is straightforward and self-explanatory. I think thought, that there are some assumptions being made and that are not clear/documented: e.g., one device and all sub-resources belong to just one fabric. Either if this holds of not, I think we would benefit from either adding a concept, or extending its documentation.

api/evpn/v1alpha1/fabric_types.go
```diff
@@ -0,0 +1,296 @@
...
+	// deviceSelector identifies which devices are members of this fabric.
+	// All devices whose labels match this selector will be enrolled.
+	// +required
+	DeviceSelector metav1.LabelSelector `json:"deviceSelector"`
```
nikatza, line 22: We currently wouldn't prevent a user defining two fabrics, with label selectors that intersect on some devices. This would result in a competition for the sub-resources and would potentially lead to an outage.  At the same time, I understand that if we support two fabrics, some devices might actually need to be in both. I think the question in that case is, how do we ensure that a single brick is indirectly managed by just one fabric? should we also label the sub resources?

nikatza (commented):
Looks good to me :) The API is self-contained and explanatory, conventions also (the `Unnumbered` bool had let me thinking for a bit), the  reconciler logic is also fitting for the purpose of this PR. 

Before we merge I would like to sync on this issue: I am wondering if we would want to have a more modular and extendable design for the `fabric` type. For example, by default we could have a "plain routed" fabric with just `deviceSelector`, `loopbacks`, and `underlay`. Then, we could extend it with references to a specialized type, similar to what we do with the `core` types.  Like this we could have support different fabrics, ike the current EVPN-VXLAN, EVPN MPLS, etc. Some of these fabrics would be reusing some types.

Obviously the `evpn` self-contained group is quite appealing too and eventually also the straight-forward way to go. 

Thanks for opening this PR, looking forward 🚀

api/evpn/v1alpha1/fabric_types.go
felix-kaestner: Added a godoc comment on DeviceSelector explicitly stating that a device must not match more than one Fabric, and that overlapping selectors lead to undefined behaviour.

api/evpn/v1alpha1/fabric_types.go
felix-kaestner: The controller now labels all sub-resources with `evpn.networking.metal.ironcore.dev/fabric: <name>` for discoverability (kubectl get interfaces -l evpn.networking.metal.ironcore.dev/fabric=my-fabric). Correctness is still handled via ownerReferences.

api/evpn/v1alpha1/fabric_types.go
```diff
@@ -0,0 +1,303 @@
...
+type FabricStatus struct {
+	// conditions represent the current state of the Fabric resource.
+	// Each condition has a unique type and reflects the status of a specific aspect of the resource.
+	// The status of each condition is one of True, False, or Unknown.
+	// +listType=map
+	// +listMapKey=type
+	// +patchStrategy=merge
+	// +patchMergeKey=type
+	// +optional
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
nikatza, line 248: Do you think we should already introduce the conditions flow for `fabric` resources? e.g., does `Ready` mean that all subresources are `configured` and --if applicable-- `operational`? It doesn't have to be exhaustive nor final.

nikatza (approved):

api/evpn/v1alpha1/fabric_types.go
felix-kaestner: Yes, this PR is only scaffolding. You can see the label being used in https://github.com/ironcore-dev/network-operator/pull/353/changes#diff-ac4bfa9497607cdf6476a774c8d7383eef7c6b7710fbcb78411998c522b545b2R353

api/evpn/v1alpha1/fabric_types.go
felix-kaestner: I'm not sure how well our very generic "every resource has configured/operational" condition maps to the `Fabric` resource. I would rather envision conditions such as "UnderlayEstablished" (based on the adjacencies built) or "OverlayReady" (for when bgp peering is established). So perhaps we just start with a singular "Ready" condition and add to that later?

nikatza: Me neither and that was the reason for my question. I observed that after we merge this PR we will set the `Ready` condition to the fabric, which is not really true  since we have not implemented the sub-resources yet. I was then wondering what the future-able conditions we could have. I don't have a good proposal for now, but I thought that if someone else implements the follow ups they could use some guideline.

PR #340, opened 2026-05-06
feat(iosxr): add support for 802.1Q and 802.1ad subinterfaces

sven-rosenzweig:
Add support for configuring subinterfaces on both physical and
bundle interfaces.

Refactor code by removing the unused ExtractVlan method and
renaming/refactoring ExtractBundleInterface to extract the
subinterface ID generically.

Clean up IPv4 network configuration by introducing a helper
function.

### DIFF
--- a/internal/provider/cisco/iosxr/intf.go
+++ b/internal/provider/cisco/iosxr/intf.go
@@ -13,7 +13,7 @@ import (
 )
 
 var (
-	bundleEtherRE       = regexp.MustCompile(`^Bundle-Ether(\d+)(?:\.\d+)?$`)
+	bundleEtherRE       = regexp.MustCompile(`^(Bundle-Ether|bundle-ether)(\d+)(\.\d+)?$`)
 	physicalInterfaceRE = regexp.MustCompile(`^(TenGigE|TwentyFiveGigE|FortyGigE|HundredGigE|GigabitEthernet)(\d){1}(\/\d){2}(\/\d+){1}(.\d{1,5})?$`)
 	loopbackInterfaceRE = regexp.MustCompile(`^(Loopback|Lo)\d+$`)
 	mgmtEthInterfaceRe  = regexp.MustCompile(`^MgmtEth\d+\/RP\d+\/CPU\d+\/\d+$`)
@@ -269,58 +269,48 @@ func CheckInterfaceNameTypePhysical(name string) error {
 	return nil
 }
 
-func ExtractVlanTagFromName(name string) (vlanID int32, err error) {
-	res := strings.Split(name, ".")
-	switch len(res) {
-	case 1:
-		return 0, nil
-	case 2:
-		vlan, err := strconv.ParseInt(res[1], 10, 32)
-		if err != nil {
-			return 0, fmt.Errorf("failed to parse VLAN ID from interface name %q: %w", name, err)
-		}
-		return int32(vlan), nil
+// ExtractBundleAndSubinterfaceID extracts the subinterface ID and bundle ID from an interface name.
+// If the interface is a physical interface, the bundle ID will be 0 and the subinterface ID will be extracted if present.
+// TwentyFiveGigE0/0/0/3.4095 -> (0, 4095), Bundle-Ether200 -> (200, 0), Bundle-Ether200.4095 -> (200, 4095)
+func ExtractBundleAndSubinterfaceID(name string) (int32, int32, error) {
+	beMatchGroups := bundleEtherRE.FindStringSubmatch(name)
+	physMatchGroups := physicalInterfaceRE.FindStringSubmatch(name)
+
+	var bID string
+	var sID string
+
+	var bundleID int32
+	var subIfaceID int32
+
+	switch {
+	case len(beMatchGroups) == 3:
+		bID = beMatchGroups[2]
+	case len(beMatchGroups) == 4:
+		bID = beMatchGroups[2]
+		sID = strings.ReplaceAll(beMatchGroups[3], ".", "")
+	case len(physMatchGroups) == 6:
+		sID = strings.ReplaceAll(physMatchGroups[5], ".", "")
 	default:
-		return 0, fmt.Errorf("unexpected interface name format %q, expected <interface> or <interface>.<vlan>", name)
-	}
-}
-
-func ExtractBundleAndSubinterfaceID(name string) (bundleID, subinterfaceID int32, err error) {
-	// Extract bundle ID and optional subinterface ID from Bundle-Ether<id> or Bundle-Ether<id>.<subif_id>
-	// Examples: Bundle-Ether200 -> (200, 0), Bundle-Ether200.4095 -> (200, 4095)
-
-	// Remove the "Bundle-Ether" or "BE" prefix
-	var idPart string
-
-	if !bundleEtherRE.MatchString(name) {
-		return 0, 0, fmt.Errorf("interface name %q does not start with Bundle-Ether or bundle-ether", name)
-	}
-	idPart = strings.TrimPrefix(strings.TrimPrefix(name, "Bundle-Ether"), "bundle-ether")
-	parts := strings.Split(idPart, ".")
-
-	if len(parts) == 0 || parts[0] == "" {
-		return 0, 0, fmt.Errorf("failed to extract bundle ID from interface name %q", name)
+		return 0, 0, fmt.Errorf("interface name %q does not start with Bundle-Ether or bundle-ether or match physical interface pattern", name)
 	}
 
-	// Parse bundle ID
-	id, err := strconv.ParseInt(parts[0], 10, 32)
-	if err != nil {
-		return 0, 0, fmt.Errorf("failed to parse bundle ID from interface name %q: %w", name, err)
+	if bID != "" {
+		bundleIDInt, err := strconv.ParseInt(bID, 10, 32)
+		if err != nil {
+			return 0, 0, fmt.Errorf("failed to parse bundle ID from interface name %q: %w", name, err)
+		}
+		bundleID = int32(bundleIDInt)
 	}
-	bundleID = int32(id)
 
-	// Parse subinterface ID if present
-	if len(parts) == 2 {
-		subIfaceIDInt, err := strconv.ParseInt(parts[1], 10, 32)
+	if sID != "" {
+		subIfaceIDInt, err := strconv.ParseInt(sID, 10, 32)
 		if err != nil {
 			return 0, 0, fmt.Errorf("failed to parse subinterface ID from interface name %q: %w", name, err)
 		}
-		subinterfaceID = int32(subIfaceIDInt)
-	} else if len(parts) > 2 {
-		return 0, 0, fmt.Errorf("unexpected interface name format %q, expected Bundle-Ether<id> or Bundle-Ether<id>.<subif_id>", name)
+		subIfaceID = int32(subIfaceIDInt)
 	}
 
-	return bundleID, subinterfaceID, nil
+	return bundleID, subIfaceID, nil
 }
 
 func CheckVlanRange(vlan string) error {
--- a/internal/provider/cisco/iosxr/intf_test.go
+++ b/internal/provider/cisco/iosxr/intf_test.go
@@ -83,6 +83,20 @@ func TestExtractBundleAndSubinterfaceID(t *testing.T) {
 			expectedSubIfaceID: 0,
 			wantErr:            true,
 		},
+		{
+			name:               "bundle-ether",
+			input:              "bundle-ether200",
+			expectedBundleID:   200,
+			expectedSubIfaceID: 0,
+			wantErr:            false,
+		},
+		{
+			name:               "TenGigE",
+			input:              "TenGigE0/0/0/1.100",
+			expectedBundleID:   0,
+			expectedSubIfaceID: 100,
+			wantErr:            false,
+		},
 	}
 
 	for _, tt := range tests {
@@ -90,11 +104,11 @@ func TestExtractBundleAndSubinterfaceID(t *testing.T) {
 			bundleID, subIfaceID, err := ExtractBundleAndSubinterfaceID(tt.input)
 			if tt.wantErr {
 				if err == nil {
-					t.Errorf("ExtractBundleAndSubinterfaceId(%s) expected error, got nil", tt.input)
+					t.Errorf("ExtractBundleAndSubinterfaceID(%s) expected error, got nil", tt.input)
 				}
 			} else {
 				if err != nil {
-					t.Errorf("ExtractBundleAndSubinterfaceId(%s) unexpected error: %v", tt.input, err)
+					t.Errorf("ExtractBundleAndSubinterfaceID(%s) unexpected error: %v", tt.input, err)
 				}
 				if bundleID != tt.expectedBundleID {
 					t.Errorf("ExtractBundleAndSubinterfaceId(%s) bundleID = %v, want %v", tt.input, bundleID, tt.expectedBundleID)
--- a/internal/provider/cisco/iosxr/provider.go
+++ b/internal/provider/cisco/iosxr/provider.go
@@ -155,37 +155,13 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
 		if bundleName == "" {
 			iface.Statistics.LoadInterval = uint8(30)
 
-			vlan, err := ExtractVlanTagFromName(name)
+			// TODO(sven-rosenzweig): support IPv6 addresses, IPv6 neighbor config
+
+			ipv4, err := NewIPv4(req.Interface.Spec.IPv4)
 			if err != nil {
 				return err
 			}
-
-			// Configure Subinterface
-			if vlan != 0 {
-				iface.SubInterface = NewVlanSubinterface(vlan, 0, "vlan-type-dot1q")
-				iface.ModeNoPhysical = "default"
-			}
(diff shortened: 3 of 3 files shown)


internal/provider/cisco/iosxr/provider.go
```diff
@@ -150,36 +150,14 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
-			if err != nil {
-				return err
-			}
-
-			// Configure Subinterface
-			if vlan != 0 {
-				iface.SubInterface = NewVlanSubinterface(vlan, 0, "vlan-type-dot1q")
-				iface.ModeNoPhysical = "default"
-			}
+			// (fixme): support IPv6 addresses, IPv6 neighbor config
```
nikatza: Just a note: We use `TODO` in the rest of the project. I see that for the `iosxr` we use `fixme` instead. May be we could bring this up in our weekly and decide for one (for consistency).

internal/provider/cisco/iosxr/provider.go
```diff
@@ -150,36 +150,14 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
-				iface.IPv4Network = IPv4Network{
-					Addresses: AddressesIPv4{
-						Primary: Primary{
-							Address: ip,
-							Netmask: netmask,
-						},
-					},
+				ipv4, err := NewIPv4(req.Interface.Spec.IPv4)
+				if err != nil {
+					return err
```
nikatza: With this change the provider will ignore ignore part of the specification of the interface (when there is more than one address specified) because of how `NewIPv4` is implemented. I think we can't still add a CEL rule that applies to a specific provider. So, I think this should be reverted so that the user gets notified that their pushed config is not working on IOSXR devices. What is your oppinion?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -251,51 +230,68 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+			return err
+		}
+
+		if subinterfaceID == 0 {
+			return fmt.Errorf("no subinterface ID in interfacename specified. pattern: <interface-name>.<subinterface-id>, got %q", name)
+		}
+
+		iface.SubInterface = VlanSubInterface{
+			VlanIdentifier: VlanIdentifier{
+				FirstTag: req.Interface.Spec.Encapsulation.OuterTag,
```
nikatza: This is probable just `...Spec.Encapsulation.Tag` since the [CEL rule](https://github.com/ironcore-dev/network-operator/blob/d49a0f3dd8bef50dffa2739cfa97686bb9cf2237/api/core/v1alpha1/interface_types.go#L194) prohibits the Inner and outer tags for `802.1q`.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -251,51 +230,68 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+				VlanType: "vlan-type-dot1q",
+			},
+		}
+
+		// Subinterface configures QAndQ vlan
+		if req.Interface.Spec.Encapsulation.InnerTag != 0 {
+			iface.SubInterface.VlanIdentifier.SecondTag = req.Interface.Spec.Encapsulation.InnerTag
+			iface.SubInterface.VlanIdentifier.VlanType = "vlan-type-dot1ad"
```
nikatza, line 275: I guess we need here to get the `OuterTag`?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -251,51 +230,68 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+		if req.Interface.Spec.MTU != 0 {
+			mtu, err := NewMTU(name, req.Interface.Spec.MTU)
+			if err != nil {
+				return err
+			}
+			iface.MTUs = mtu
+		}
+
+		if req.Interface.Spec.IPv4 != nil {
+			ipv4, err := NewIPv4(req.Interface.Spec.IPv4)
```
nikatza: Same question as before: can we ignore extra addresses or should we return an `error`?

internal/provider/cisco/iosxr/provider.go
```diff
@@ -319,6 +315,25 @@ func NewMTU(intName string, mtu int32) (MTUs, error) {
 	}}}, nil
 }
 
+func NewIPv4(ips *v1alpha1.InterfaceIPv4) (IPv4Network, error) {
```
nikatza, line 313: As discussed earlier, may be we could just return also an error if there is more than one address defined and just propagate it.

internal/provider/cisco/iosxr/intf_test.go
```diff
@@ -83,18 +83,32 @@ func TestExtractBundleAndSubinterfaceID(t *testing.T) {
...
+		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			bundleID, subIfaceID, err := ExtractBundleAndSubinterfaceID(tt.input)
 			if tt.wantErr {
 				if err == nil {
-					t.Errorf("ExtractBundleAndSubinterfaceId(%s) expected error, got nil", tt.input)
+					t.Errorf("ExtractSubinterfaceIdFromInterfacenName(%s) expected error, got nil", tt.input)
```
nikatza:
```suggestion
                    t.Errorf("ExtractSubinterfaceIdFromInterfacenName(%s) expected error, got nil", tt.input)
```

internal/provider/cisco/iosxr/intf_test.go
```diff
@@ -83,18 +83,32 @@ func TestExtractBundleAndSubinterfaceID(t *testing.T) {
...
 				}
 			} else {
 				if err != nil {
-					t.Errorf("ExtractBundleAndSubinterfaceId(%s) unexpected error: %v", tt.input, err)
+					t.Errorf("ExtractSubinterfaceIdFromInterfacenName(%s) unexpected error: %v", tt.input, err)
```
nikatza:
```suggestion
                    t.Errorf("ExtractSubinterfaceIdFromInterfacenName(%s) unexpected error: %v", tt.input, err)
```

internal/provider/cisco/iosxr/intf.go
```diff
@@ -269,58 +269,49 @@ func CheckInterfaceNameTypePhysical(name string) error {
...
+// Extract the subinterface ID and bundle ID from an interface name.
+// If the interface is a physical interface, the bundle ID will be 0 and the subinterface ID will be extracted if present.
+// TwentyFiveGigE0/0/0/3.4095 -> (0, 4095), Bundle-Ether200 -> (200, 0), Bundle-Ether200.4095 -> (200, 4095)
+func ExtractBundleAndSubinterfaceID(name string) (bundleID, subinterfaceID int32, err error) {
+	var idPart string
+	var typePhysical bool
+
+	switch {
+	case bundleEtherRE.MatchString(name):
+		idPart = strings.TrimPrefix(strings.TrimPrefix(name, "Bundle-Ether"), "bundle-ether")
```
nikatza: This works well, but I am thinking that perhaps we could modify and reuse the regex and use the capture group. Something like:
```
var bundleEtherRE = regexp.MustCompile('^(?:Bundle-Ether|bundle-ether)(\d+)(?:\.(\d+))?$')
```
and then something like:
```
if m := bundleEtherRE.FindStringSubmatch(name); m != nil {
    idPart = m[1]  
}
```
This way we would just need to touch the regex if we want to change something in the future.

internal/provider/cisco/iosxr/intf.go
```diff
@@ -269,58 +269,49 @@ func CheckInterfaceNameTypePhysical(name string) error {
...
-	switch len(res) {
-	case 1:
-		return 0, nil
-	case 2:
-		vlan, err := strconv.ParseInt(res[1], 10, 32)
-		if err != nil {
-			return 0, fmt.Errorf("failed to parse VLAN ID from interface name %q: %w", name, err)
-		}
-		return int32(vlan), nil
+// Extract the subinterface ID and bundle ID from an interface name.
```
nikatza: The convention is to use as first word the name of the function, i.e., `ExtractBundleAndSubinterfaceID`

internal/provider/cisco/iosxr/intf.go
```diff
@@ -269,58 +269,49 @@ func CheckInterfaceNameTypePhysical(name string) error {
...
-	// Extract bundle ID and optional subinterface ID from Bundle-Ether<id> or Bundle-Ether<id>.<subif_id>
-	// Examples: Bundle-Ether200 -> (200, 0), Bundle-Ether200.4095 -> (200, 4095)
 
-	// Remove the "Bundle-Ether" or "BE" prefix
-	var idPart string
+	parts := strings.Split(idPart, ".")
 
-	if !bundleEtherRE.MatchString(name) {
-		return 0, 0, fmt.Errorf("interface name %q does not start with Bundle-Ether or bundle-ether", name)
+	if len(parts) > 2 || len(parts) < 1 {
```
nikatza: I think that `len(parts)` can never be lower than one. The `strings.Split` signature states:
```
If s does not contain sep and sep is not empty, Split returns a slice of length 1 whose only element is s.
```

internal/provider/cisco/iosxr/intf.go
```diff
@@ -269,58 +269,49 @@ func CheckInterfaceNameTypePhysical(name string) error {
...
+		return 0, 0, fmt.Errorf("failed to extract subinterface ID from interface name %q", name)
 	}
-	idPart = strings.TrimPrefix(strings.TrimPrefix(name, "Bundle-Ether"), "bundle-ether")
-	parts := strings.Split(idPart, ".")
 
-	if len(parts) == 0 || parts[0] == "" {
-		return 0, 0, fmt.Errorf("failed to extract bundle ID from interface name %q", name)
+	if !typePhysical {
```
nikatza: In Golang we should try to avoid the use flags to decide the code path. Instead, the best practice would be to apply an early return per branch to avoid shared mutable state.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -251,51 +230,68 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+
+		if req.Interface.Spec.MTU != 0 {
+			mtu, err := NewMTU(name, req.Interface.Spec.MTU)
+			if err != nil {
+				return err
+			}
+			iface.MTUs = mtu
+		}
+
+		if req.Interface.Spec.IPv4 != nil {
```
nikatza: Since we check that `Interface.Spec.IPv4` is not `nil`, we could avoid checking for nilness in the `NewIPv4` function and change its signature to `func NewIPv4(ips v1alpha1.InterfaceIPv4) (IPv4Network, error)`, avoid double checking and leave the responsability to the caller.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -251,51 +230,68 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
 	}
 }
 
-func NewBundleInterface(req *v1alpha1.Interface) Iface {
-	bundle := Iface{
-		Name:        req.Spec.Name,
-		Description: req.Spec.Description,
-		// Set Interface mode to virtual for bundle interfaces
-		Mode: gnmiext.Empty(true),
+func NewInterface(name, descr string) Iface {
```
nikatza: Since there is no complex logic in this function we may be better off by just instantiating directly. We don't save many lines anyway and we can set the `iface.Active` field in the same instantiation.

internal/provider/cisco/iosxr/intf.go
```diff
@@ -269,58 +269,49 @@ func CheckInterfaceNameTypePhysical(name string) error {
...
-	case 2:
-		vlan, err := strconv.ParseInt(res[1], 10, 32)
-		if err != nil {
-			return 0, fmt.Errorf("failed to parse VLAN ID from interface name %q: %w", name, err)
-		}
-		return int32(vlan), nil
+// Extract the subinterface ID and bundle ID from an interface name.
+// If the interface is a physical interface, the bundle ID will be 0 and the subinterface ID will be extracted if present.
+// TwentyFiveGigE0/0/0/3.4095 -> (0, 4095), Bundle-Ether200 -> (200, 0), Bundle-Ether200.4095 -> (200, 4095)
+func ExtractBundleAndSubinterfaceID(name string) (bundleID, subinterfaceID int32, err error) {
```
nikatza: I did not catch this. Claude suggests using named returns we should mostly use bare returns. We are not doing this here, so may be we should just use:
```
func ExtractBundleAndSubinterfaceID(name string) (int32, int32, error) {
```
This might be indeed easier to follow because we have 6 return cases.

nikatza (changes requested):
Thanks for the PR. I have several small comments, many of the just nits. A couple of them seem to be bugs. i think this should be easier to fix for you :) 

🚀 🚀 🚀

internal/provider/cisco/iosxr/intf.go
nikatza: Another thing (I added this in the review but is not appearing here). A common convention is that the first word in the doc is the name of the function (here `Extract` should be `ExtractBundleAndSubinterfaceID`

internal/provider/cisco/iosxr/provider.go
```diff
@@ -150,36 +150,14 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
 			iface.Statistics.LoadInterval = uint8(30)
 
-			vlan, err := ExtractVlanTagFromName(name)
-			if err != nil {
-				return err
-			}
-
-			// Configure Subinterface
-			if vlan != 0 {
-				iface.SubInterface = NewVlanSubinterface(vlan, 0, "vlan-type-dot1q")
```
nikatza, line 165: `NewVlanSubinterface` is now dead code, we should remove it.

internal/provider/cisco/iosxr/provider.go
sven-rosenzweig: Happy to use ``TODO``

internal/provider/cisco/iosxr/provider.go
sven-rosenzweig: you are right about this.

internal/provider/cisco/iosxr/intf.go
sven-rosenzweig: Reworked the method. Also addresses R291, R295

internal/provider/cisco/iosxr/intf_test.go
nikatza: Apologies, this should be: `ExtractBundleAndSubinterfaceID`

internal/provider/cisco/iosxr/provider.go
```diff
@@ -256,51 +233,67 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
-			// conf = append(conf, &iface)
 		}
 		return updateInterface(ctx, p.client, conf...)
-	case v1alpha1.InterfaceTypeLoopback, v1alpha1.InterfaceTypeRoutedVLAN, v1alpha1.InterfaceTypeSubinterface:
+	case v1alpha1.InterfaceTypeSubinterface:
+		iface := NewInterface(name, req.Interface.Spec.Description)
+		// Set Interface mode to virtual for bundle interfaces
+		iface.Mode = gnmiext.Empty(false)
+		iface.ModeNoPhysical = "default"
+		iface.Active = "act"
```
nikatza: I still think that the NewInterface method does not bring much value to the code base. A simple instantation makes clear what `iface` is since we avoid an indirection. For example:
```suggestion
        iface := Iface{
			Name:        name,
			Description: req.Interface.Spec.Description,
			Active:      "act",
			ModeNoPhysical: "default",
			Mode: gnmiext.Empty(false),
		}
```

internal/provider/cisco/iosxr/provider.go
```diff
@@ -229,8 +206,8 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
 			return err
 		}
 
-		iface := NewBundleInterface(req.Interface)
-
+		iface := NewInterface(name, req.Interface.Spec.Description)
+		iface.Active = "act"
```
nikatza: This could be just:

```suggestion
        iface := Iface{
			Name:        name,
			Description: req.Interface.Spec.Description,
			Active:      "act",
		}
```

nikatza (commented):
Some of my comments were not correctly (in the tests). Here is another pass. Thanks for the effort to incorporate the feedback :)

internal/provider/cisco/iosxr/provider.go
```diff
@@ -256,62 +239,69 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+			Name:           name,
+			Description:    req.Interface.Spec.Description,
+			Active:         "act",
+			Mode:           gnmiext.Empty(false),
+			ModeNoPhysical: "default",
+		}
+
+		iface.Mode = gnmiext.Empty(false)
+		iface.ModeNoPhysical = "default"
+		iface.Active = "act"
```
nikatza: This is not needed anymore :)

nikatza (approved):

nikatza (approved):
Nice work! Looking forward having this working 🚀

PR #353, opened 2026-05-12
Implement EVPN Fabric loopback reconciliation

felix-kaestner:
Reconcile lo0 (Router-ID), lo1/lo2 (VTEP), and lo100 (anycast RP) loopback Claims and Interfaces for all fabric devices. The Fabric becomes the controller owner of each Claim so the Owns() watch re-enqueues correctly. Interfaces are only created once the Claim is allocated; a guard at the top of Reconcile short-circuits when the provider does not implement InterfaceProvider.

Extend provider interface with LoopbackInterfaceName so each driver can return a vendor-specific handle for a given loopback index.

Add envtest suite covering Claims, Interfaces, ownership, and the Ready condition.

### DIFF
--- a/internal/controller/evpn/fabric_controller.go
+++ b/internal/controller/evpn/fabric_controller.go
@@ -9,17 +9,23 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
 	evpnv1alpha1 "github.com/ironcore-dev/network-operator/api/evpn/v1alpha1"
+	poolv1alpha1 "github.com/ironcore-dev/network-operator/api/pool/v1alpha1"
 	"github.com/ironcore-dev/network-operator/internal/conditions"
 	"github.com/ironcore-dev/network-operator/internal/provider"
 )
@@ -36,13 +42,17 @@ type FabricReconciler struct {
 	// More info: https://book.kubebuilder.io/reference/raising-events
 	Recorder events.EventRecorder
 
-	// Provider is the driver that will be used to create & delete the interface.
+	// Provider is the driver that will be used to create interfaces.
 	Provider provider.ProviderFunc
 }
 
 // +kubebuilder:rbac:groups=evpn.networking.metal.ironcore.dev,resources=fabrics,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=evpn.networking.metal.ironcore.dev,resources=fabrics/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=evpn.networking.metal.ironcore.dev,resources=fabrics/finalizers,verbs=update
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=devices,verbs=get;list;watch
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=interfaces,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=claims,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=ipaddresspools,verbs=get;list;watch
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
@@ -62,6 +72,18 @@ func (r *FabricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 		}
 		log.Error(err, "Failed to get resource")
 		return ctrl.Result{}, err
+	}
+
+	if _, ok := r.Provider().(provider.InterfaceProvider); !ok {
+		if meta.SetStatusCondition(&fabric.Status.Conditions, metav1.Condition{
+			Type:    v1alpha1.ReadyCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  v1alpha1.NotImplementedReason,
+			Message: "Provider does not implement provider.InterfaceProvider",
+		}) {
+			return ctrl.Result{}, r.Status().Update(ctx, fabric)
+		}
+		return ctrl.Result{}, nil
 	}
 
 	if !fabric.DeletionTimestamp.IsZero() {
@@ -136,6 +158,15 @@ func (r *FabricReconciler) SetupWithManager(mgr ctrl.Manager) error {
 
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&evpnv1alpha1.Fabric{}).
+		Owns(&poolv1alpha1.Claim{}).
+		Owns(&v1alpha1.Interface{}).
+		// Re-reconcile when a Device's labels change so that devices newly
+		// matching a deviceSelector are enrolled into the fabric.
+		Watches(
+			&v1alpha1.Device{},
+			handler.EnqueueRequestsFromMapFunc(r.devicesToFabrics),
+			builder.WithPredicates(predicate.LabelChangedPredicate{}),
+		).
 		WithEventFilter(filter).
 		Named("evpn-fabric").
 		Complete(r)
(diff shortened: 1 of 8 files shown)


internal/controller/evpn/fabric_controller.go
```diff
@@ -169,3 +202,215 @@ func (r *FabricReconciler) finalize(ctx context.Context, fabric *evpnv1alpha1.Fa
 	_ = fabric
 	return nil
 }
+
+const (
+	LoopbackRouterID    = 0   // Router-ID and BGP source address, present on all fabric devices
+	LoopbackVTEP        = 1   // Primary VTEP address, present on VTEP devices
+	LoopbackVTEPAnycast = 2   // Anycast VTEP address, present on VTEP devices (deprecated in favour of ESI)
+	LoopbackAnycastRP   = 100 // PIM anycast rendezvous point address, shared across RP devices
+)
```
nikatza, line 211: This is currently not  directly visible/configurable by the end-user. Other users might have requirements where the ID is already used. I  would propose to introduce a config resource referenced by fabric resources. Would you favor that?

internal/controller/evpn/fabric_controller.go
```diff
@@ -169,3 +202,215 @@ func (r *FabricReconciler) finalize(ctx context.Context, fabric *evpnv1alpha1.Fa
...
+		intf.Spec.AdminState = v1alpha1.AdminStateUp
+		switch loopbackID {
+		case LoopbackRouterID:
+			intf.Spec.Description = "Router-ID, BGP Source"
+		case LoopbackVTEP:
+			intf.Spec.Description = "Primary VTEP"
+		case LoopbackVTEPAnycast:
+			intf.Spec.Description = "VTEP Anycast"
+		case LoopbackAnycastRP:
+			intf.Spec.Description = "Rendezvous Point"
```
nikatza, line 366: Thinking out loud: Would it make sense to prefix this with an indicator that this is managed by an operator? Probably not, since the end user probably will be better of using our CLI tool to retrieve a list objects that are operator-managed. Wondering however, if the description is something we add into a `Fabric` config resource together with the ID, if we go down that road.

internal/provider/openconfig/interface.go
```diff
@@ -193,6 +193,10 @@ func (p *Provider) InterfaceNameEqual(_ context.Context, a, b string) (bool, err
 	return a == b, nil
 }
 
+func (p *Provider) LoopbackInterfaceName(id int) (string, error) {
+	return fmt.Sprintf("lo%d", id), nil
```
swagner-de, line 197: [Is this standardized over openconfig?](https://www.youtube.com/watch?v=j4ovbmsp6p0).

Maybe a TODO for later on.

internal/controller/evpn/fabric_controller.go
```diff
@@ -169,3 +202,215 @@ func (r *FabricReconciler) finalize(ctx context.Context, fabric *evpnv1alpha1.Fa
...
+
+// reconcileSystemLoopbacks ensures lo0 (Router-ID / BGP source) exists on every fabric device.
+func (r *FabricReconciler) reconcileSystemLoopbacks(ctx context.Context, fabric *evpnv1alpha1.Fabric) (ctrl.Result, error) {
+	selector, err := metav1.LabelSelectorAsSelector(&fabric.Spec.DeviceSelector)
+	if err != nil {
+		return ctrl.Result{}, reconcile.TerminalError(fmt.Errorf("invalid deviceSelector: %w", err))
+	}
+	devices := &v1alpha1.DeviceList{}
+	if err := r.List(ctx, devices, client.InNamespace(fabric.Namespace), client.MatchingLabelsSelector{Selector: selector}); err != nil {
+		return ctrl.Result{}, fmt.Errorf("listing devices: %w", err)
```
nikatza: nit: "listing devices for system loopbacks reconciliation" (same below)

internal/controller/evpn/fabric_controller.go
```diff
@@ -169,3 +202,215 @@ func (r *FabricReconciler) finalize(ctx context.Context, fabric *evpnv1alpha1.Fa
...
+func (r *FabricReconciler) reconcileLoopbackInterface(ctx context.Context, fabric *evpnv1alpha1.Fabric, device *v1alpha1.Device, loopbackID int, claim *poolv1alpha1.Claim) error {
+	cond := conditions.Get(claim, poolv1alpha1.AllocatedCondition)
+	if cond == nil || cond.Status != metav1.ConditionTrue || claim.Status.Value == "" {
+		return nil
+	}
+
+	prefix, err := v1alpha1.ParsePrefix(claim.Status.Value + "/32")
+	if err != nil {
+		return reconcile.TerminalError(fmt.Errorf("parsing allocated address %q: %w", claim.Status.Value, err))
+	}
```
nikatza, line 335: This seems redundant / too defensive given the pool type is already defined.

nikatza (approved):
Nice! 🔥

internal/controller/evpn/fabric_controller.go
```diff
@@ -169,3 +202,215 @@ func (r *FabricReconciler) finalize(ctx context.Context, fabric *evpnv1alpha1.Fa
...
+	return claim, nil
+}
+
+// reconcileLoopbackInterface creates or updates the Interface for a given device loopback
+// once its Claim is allocated. A no-op if the claim is not yet allocated; the Owns() watch
+// on Claim will re-enqueue this Fabric when the pool controller updates the claim status.
+func (r *FabricReconciler) reconcileLoopbackInterface(ctx context.Context, fabric *evpnv1alpha1.Fabric, device *v1alpha1.Device, loopbackID int, claim *poolv1alpha1.Claim) error {
+	cond := conditions.Get(claim, poolv1alpha1.AllocatedCondition)
+	if cond == nil || cond.Status != metav1.ConditionTrue || claim.Status.Value == "" {
+		return nil
```
swagner-de, line 329: If this returns `nil` it means the Fabric Resource jumps to Ready, even if the loopback is not allocated, right?

nikatza: This is a longer discussion...

internal/controller/evpn/fabric_controller.go
```diff
@@ -147,7 +178,9 @@ type ReconcileFunc func(context.Context, *evpnv1alpha1.Fabric) (ctrl.Result, err
 
 func (r *FabricReconciler) reconcile(ctx context.Context, fabric *evpnv1alpha1.Fabric) (ctrl.Result, error) {
 	phases := []ReconcileFunc{
-		// r.reconcileNodes,
+		r.reconcileSystemLoopbacks,
+		r.reconcileVTEPLoopbacks,
+		r.reconcileAnycastRPLoopbacks,
```
nikatza, line 183: As long we only support one provider it will work. If we add support for multiple simultaneously providers the iteration will return early if one provider fails, skipping some that would otherwise work. Likewise, if an error while creating one of the subresources in k8s due to e.g., the provider raising an error, we will return early and skip some of the remaining subresources?. Is my understanding correct?

nikatza (approved):

internal/controller/evpn/fabric_controller.go
felix-kaestner: As agreed in our call, we will defer such configuration to a later follow-up and proceed with pre-selected/hardcoded IDs for now.

internal/controller/evpn/fabric_controller.go
felix-kaestner: I'm a bit worried that a prefix might just unnecessarily bloat the description. In an ideal scenario, everything on a device would be managed by the operator, in which case having the prefix wouldn't add any additional value. I would prefer if we keep it like this for now and consider a follow-up if we really see the need for either prefixed descriptions or even user-configured ones via the Fabric spec.

internal/controller/evpn/fabric_controller.go
felix-kaestner: What would you recommend instead? Returning no error?

internal/provider/openconfig/interface.go
felix-kaestner: No it isn't and will also never be. The interface handle is a YANG tree key which is intentionally left to the platform as a value. Therefore we either have to introduce platform-branching or consider an alternative approach. 

However, using `lo%d` seems to be the approach that seems to correlate with the most platforms, so I used that as a best-effort case for now. We can still iterate on this in the future, but I'd leave this out of scope of the `Fabric` implementation.

internal/controller/evpn/fabric_controller.go
felix-kaestner: Will tackle this as a follow-up PR.

internal/controller/evpn/fabric_controller.go
felix-kaestner: > If we add support for multiple simultaneously providers 

Could you explain what you mean by this? 

> we will return early and skip some of the remaining subresources

Yes, this is intentional. We return early, if we encounter an error and not proceed to create resources if previous steps fail. I think for now this is the best approach, as otherwise we would need to model dependencies between steps (e.g. I need the loopback before I can create dependent resources) which would explode the complexity for this initial implementation. Hence I would like to have a very basic and implementation to start with and then iterate on it to eventually improve on such cases.

internal/controller/evpn/fabric_controller.go
nikatza: My understanding here is that `ParsePrefix` will never return `error` because the the Claim was successful and its `Value` is already a valid IPAddress. That said, the check doesn't hurt either.

internal/controller/evpn/fabric_controller.go
nikatza: Please ignore my comment about the multiple providers, that was just wrong since `LoopbackInterfaceName` is the only Provider-specific method being called and it must be implemented. Not sure if the method should return an `error` at all for now. 

Thanks for the other clarification :)

nikatza (approved):

internal/controller/evpn/fabric_controller.go
felix-kaestner: Yes, that is true. However as the signature includes an `error` I think we should rather explicitly handle it that to just assign to a `_` despite the fact that this should™ basically never happen.

PR #361, opened 2026-05-18
Implement EVPN Fabric underlay p2p link reconciliation

felix-kaestner:
Add numbered underlay addressing support to the Fabric controller. When spec.underlay.addressing.ipPrefixPoolRef is set, the controller allocates /31 prefix Claims per fabric link and assigns host addresses to both interface endpoints.

Also refactor the test suite:
- Remove helper functions, inline all resource creation
- Harmonize GenerateName conventions (spine<N>-, leaf<N>-, etc.)
- Simplify IPv4 assignment in unnumbered/numbered phases
- Add V(1) info log for interfaces missing neighbor label
- Add TODO comments for future SSA migration on patch calls

### DIFF
--- a/internal/controller/evpn/fabric_controller.go
+++ b/internal/controller/evpn/fabric_controller.go
@@ -4,8 +4,10 @@
 package evpn
 
 import (
+	"cmp"
 	"context"
 	"fmt"
+	"net/netip"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -14,6 +16,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -53,6 +56,7 @@ type FabricReconciler struct {
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=interfaces,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=claims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=ipaddresspools,verbs=get;list;watch
+// +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=ipprefixpools,verbs=get;list;watch
 // +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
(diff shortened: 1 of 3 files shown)


internal/controller/evpn/fabric_controller.go
```diff
@@ -382,6 +394,159 @@ func (r *FabricReconciler) reconcileLoopbackInterface(ctx context.Context, fabri
...
+	intf.Spec.IPv4 = &v1alpha1.InterfaceIPv4{
+		Unnumbered: &v1alpha1.InterfaceIPv4Unnumbered{
+			InterfaceRef: v1alpha1.LocalObjectReference{
+				Name: fmt.Sprintf("%s-%s-lo0", fabric.Name, intf.Spec.DeviceRef.Name),
+			},
+		},
+	}
+	if equality.Semantic.DeepEqual(orig.Spec, intf.Spec) {
+		return nil
+	}
```
nikatza, line 459: nit: applyconfiguration -> apply configuration

internal/controller/evpn/fabric_controller.go
```diff
@@ -382,6 +394,159 @@ func (r *FabricReconciler) reconcileLoopbackInterface(ctx context.Context, fabri
...
+// assigns a host address to this interface. The peer is identified by PhysicalInterfaceNeighborLabel;
+// if absent the interface is skipped (link not yet complete). The claim name is derived from
+// the two interface names sorted alphabetically so both ends resolve to the same Claim.
+// The lexicographically first interface receives addr 0 of the /31; the second receives addr 1.
+func (r *FabricReconciler) reconcileUnderlayInterfaceNumbered(ctx context.Context, fabric *evpnv1alpha1.Fabric, intf *v1alpha1.Interface) error {
+	peerName, ok := intf.Labels[v1alpha1.PhysicalInterfaceNeighborLabel]
+	if !ok {
+		ctrl.LoggerFrom(ctx).V(1).Info("Skipping interface without neighbor label", "interface", intf.Name)
+		return nil
+	}
```
nikatza, line 477: Would it make sense to consider the value of `status.neighbors[].validation` and return an error if the interface could not be verified? Currently the label must be added by the user anyhow: I guess the user would expect that the controller takes that into consideration. It is not a must here, but may be worth considering for a follow-up. If so I would happily volunteer for the task.

internal/controller/evpn/fabric_controller.go
```diff
@@ -382,6 +394,159 @@ func (r *FabricReconciler) reconcileLoopbackInterface(ctx context.Context, fabri
...
+}
+
+// reconcileUnderlayInterfaceNumbered allocates a /31 prefix per link pair via a Claim and
+// assigns a host address to this interface. The peer is identified by PhysicalInterfaceNeighborLabel;
+// if absent the interface is skipped (link not yet complete). The claim name is derived from
+// the two interface names sorted alphabetically so both ends resolve to the same Claim.
+// The lexicographically first interface receives addr 0 of the /31; the second receives addr 1.
+func (r *FabricReconciler) reconcileUnderlayInterfaceNumbered(ctx context.Context, fabric *evpnv1alpha1.Fabric, intf *v1alpha1.Interface) error {
+	peerName, ok := intf.Labels[v1alpha1.PhysicalInterfaceNeighborLabel]
+	if !ok {
```
nikatza, line 474: If an interface was previously correctly reconciled then the controller created a `Prefix` claim. However, if later a user removes the label then the claim won't be cleaned up, right? I guess we might to take that into consideration and manage that lifecycle.

internal/controller/evpn/fabric_controller.go
felix-kaestner: That is the name of the generator tool (ref/ https://github.com/kubernetes-sigs/controller-tools/blob/main/pkg/applyconfiguration/doc.go). So I would leave it like it is.

internal/controller/evpn/fabric_controller.go
felix-kaestner: Taking the validation into account would mean that lldp would have to be enabled on these devices beforehand. Also, that means that only providers which implement/support lldp can be used with the `Fabric` resource, so this would feel like a limitation. Therefore I would leave this as a potential follow-up for later.

internal/controller/evpn/fabric_controller.go
felix-kaestner: As these claims have a deterministic name, another controller run would re-find the existing claim and should add back the label, I think.

swagner-de (approved):

PR #362, opened 2026-05-18
Pause child resources while Device phase is not Running

felix-kaestner:
Previously, computeCondition only paused child resources when the Device phase was a non-empty value other than Running. A freshly created Device starts with an empty phase before the DeviceReconciler sets it to Pending/Running. During this window, children were not paused and could attempt reconciliation against a device that is not yet ready.

Remove the empty-phase exclusion in computeCondition so that any phase other than Running (including empty) pauses child resources.

Align DevicePausedChanged with this semantic: previously it also excluded the empty phase from its change detection, meaning the transition from "" to "Running" was invisible to the watch predicate and children were never re-triggered to unpause. Rewrite the predicate using positive-form comparisons (==) for clarity.

Add a DeviceReconciler to the cisco/nx controller test suite so that test Devices automatically transition to Running when no provisioning is configured, matching production behavior.


SchwarzM (approved):
lgtm

PR #363, opened 2026-05-20
Implement BGP Peering for IOS XR

sven-rosenzweig:


api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -59,6 +59,10 @@ type BGPPeerSpec struct {
 	// Controls which address families are enabled and their specific configuration.
 	// +optional
 	AddressFamilies *BGPPeerAddressFamilies `json:"addressFamilies,omitempty"`
+
+	// LocalAS specifies a local AS number to present to the BGP peer without changing the main BGP process ASN.
```
felix-kaestner:
```suggestion
	// LocalAS specifies a local AS number to present to the BGP peer, masking the global BGP process ASN.
```

Could you also extend the current validation webhook to verify this AS Number, same as the remote as number field?

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -59,6 +59,10 @@ type BGPPeerSpec struct {
...
+	// +optional
+	LocalAS intstr.IntOrString `json:"localAS,omitempty"`
```
nikatza: We probably want this to make this a pointer as this type does not have built-in nil value. As described in [the k8s API conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required) in more detail, a pointer helps to distinguish if the this is just not set or it is set but empty. 

I guess it would be possible to keep the type as it is assuming that ASN0 is reserved and must not be used anyway. However, I think the pointer would be cleaner.

felix-kaestner: On a second thought, in BGP Process, we called the field "ASNumber". So should this naturally be "LocalASNumber"? How would the capitalization then look for the json/yaml key? "localASNumber"?

sven-rosenzweig: ``LocalASNumber`` it is.

sven-rosenzweig: Valid point, made this a pointer of type intstr.IntOrString.

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -59,6 +59,10 @@ type BGPPeerSpec struct {
 	// Controls which address families are enabled and their specific configuration.
 	// +optional
 	AddressFamilies *BGPPeerAddressFamilies `json:"addressFamilies,omitempty"`
+
+	// LocalASNumber specifies a local AS number to present to the BGP peer, masking the global BGP process ASN.
+	// +optional
+	LocalASNumber *int `json:"localAS,omitempty"`
```
felix-kaestner: `LocalASNumber` only applies to eBGP neighbors (i.e. BGPPeer.asNumber != BGP.asNumber). As this check goes across multiple resources, we should put that into the controller for the BGPPeer (`internal/controller/core/bgp_peer_controller.go#`( to check for this condition, i.e. something like:

```diff
diff --git a/internal/controller/core/bgp_peer_controller.go b/internal/controller/core/bgp_peer_controller.go
index 39645e1c..8891da8a 100644
--- a/internal/controller/core/bgp_peer_controller.go
+++ b/internal/controller/core/bgp_peer_controller.go
@@ -441,6 +441,16 @@ func (r *BGPPeerReconciler) reconcile(ctx context.Context, s *bgpPeerScope) (ret
 		sourceInterface = intf.Spec.Name
 	}
 
+	if s.BGPPeer.Spec.LocalASNumber != nil && s.BGPPeer.Spec.ASNumber.String() == bgp.Spec.ASNumber.String() {
+		conditions.Set(s.BGPPeer, metav1.Condition{
+			Type:    v1alpha1.ConfiguredCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  v1alpha1.ErrorReason,
+			Message: "local-as cannot be configured on iBGP peers",
+		})
+		return reconcile.TerminalError(fmt.Errorf("local-as cannot be configured on iBGP peers"))
+	}
+
 	if err := s.Provider.Connect(ctx, s.Connection); err != nil {
 		return fmt.Errorf("failed to connect to provider: %w", err)
 	}
```

Returning a terminal error, if that validation fails.

sven-rosenzweig: You are right. Will add the check.

felix-kaestner (approved):

PR #365, opened 2026-05-20
Rewrite OpenConfig provider without ygot/ygnmi

felix-kaestner:
Replace the ygot-generated code with hand-crafted structs and gnmiext.Client, following the same patterns used by the NX-OS provider.


internal/provider/openconfig/device.go
```diff
@@ -0,0 +1,173 @@
...
+	chassis := new(ChassisComponentState)
+	if err := p.client.GetState(ctx, sys, chassis); err != nil {
+		return nil, fmt.Errorf("failed to get device info: %w", err)
+	}
+
+	return &provider.DeviceInfo{
+		Hostname:        sys.Hostname,
+		Manufacturer:    chassis.MfgName,
+		Model:           chassis.ModelName,
+		SerialNumber:    sanitize(chassis.SerialNo),
```
nikatza: Could you add a comment on why the sanitize is needed, please? It's a bit odd to modify the value returned by the device.

It's quite unlikely that two devices would have the same `SerialNumber` due to a collision after the truncation. However, it's still a possibility. That is one of the reasons for my question.

nikatza (approved):

internal/provider/openconfig/device.go
felix-kaestner: As the serial number is written into a label on the `Device` resource, it needs to adhere to some characteristics (ref/ https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set). In my tests, on a Nokia SRLinux Node in Containerlab the Serial Number was just exposed as `"Sim Serial No."` so a placeholder, which is not a valid label value.

That said, I think this sanitization belongs better into the device controller, as it is the one that sets the serial number as a label. The provider implementation should rather just return the value as is.

PR #378, opened 2026-05-27
Add EthernetSegment resource and NX-OS provider implementation

felix-kaestner:
Introduce the EthernetSegment API type for EVPN multi-homing (RFC 7432 Section 5) with controller, CRD, and NX-OS provider.

The controller validates that the referenced interface is an aggregate with switchport configuration and watches for switchport state transitions. The NX-OS provider enables EVPN multihoming globally, configures the ESI per port-channel via a subtree patch, and rejects SingleActive mode and vPC coexistence with typed terminal errors.


api/core/v1alpha1/ethernetsegment_types.go
```diff
@@ -0,0 +1,139 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="InterfaceRef is immutable"
+	InterfaceRef LocalObjectReference `json:"interfaceRef"`
+
+	// ESI is the 10-byte Ethernet Segment Identifier in colon-separated hex notation
+	// (e.g., "00:11:22:33:44:55:66:77:88:01"). Must not be all-zeros or all-ones (reserved per RFC 7432).
+	// Immutable.
+	// +required
+	// +kubebuilder:validation:Pattern=`^([0-9a-fA-F]{2}:){9}[0-9a-fA-F]{2}$`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ESI is immutable"
+	ESI string `json:"esi"`
```
nikatza: The RFC goes a bit more in detail into the format of the ESI (type and corresponding values). I am wondering if we should be more explicit and go into that level of detail and show the full ESI value in the status. The advantage might be easier human validation. On the other hand, the current approach results in a leaner API.

OpenConfig does have an [esi type](https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-ethernet-segments.html#ethernet-segments-ethernet-segment-config-esi-type), but does not go beyond that. 

One of the interesting things is that type I values are auto-generated:  (in nxos with `identifier auto lacp`). I'm not sure if we want this to be optional, but currently the API does not support a way to allow that. A user would have to pre-fetch the values from the switch in advance.

Overall I would say we can use this approach, but may be we could consider adding the type in the status.

internal/controller/core/ethernetsegment_controller.go
```diff
@@ -0,0 +1,502 @@
...
+	err = s.Provider.EnsureEthernetSegment(ctx, &provider.EnsureEthernetSegmentRequest{
+		EthernetSegment: s.EthernetSegment,
+		Interface:       intf,
+		ProviderConfig:  s.ProviderConfig,
+	})
+
+	cond := conditions.FromError(err)
+	// As this resource is configuration only, we use the Configured condition as top-level Ready condition.
+	cond.Type = v1alpha1.ReadyCondition
+	conditions.Set(s.EthernetSegment, cond)
```
nikatza, line 338: We should set the `Configured` condition (as we discussed in our last meeting), a defer that calls the recompute of the `Ready` condition, and the initialization of the status upon creation. 

Adam is preparing already a PR to update all other affected resources.

internal/controller/core/ethernetsegment_controller.go
```diff
@@ -0,0 +1,502 @@
...
+	key := client.ObjectKey{
+		Name:      s.EthernetSegment.Spec.InterfaceRef.Name,
+		Namespace: s.EthernetSegment.Namespace,
+	}
+
+	intf := new(v1alpha1.Interface)
+	if err := r.Get(ctx, key, intf); err != nil {
+		if apierrors.IsNotFound(err) {
+			conditions.Set(s.EthernetSegment, metav1.Condition{
+				Type:    v1alpha1.ReadyCondition,
```
nikatza: Depending on the action above we would want to change this condition's type (same would apply to the other checks).

internal/controller/core/ethernetsegment_controller_test.go
```diff
@@ -0,0 +1,185 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("Cleanup the specific resource instance Device")
+			Expect(k8sClient.Delete(ctx, resource)).To(Succeed())
+
+			By("Ensuring the resource is deleted from the provider")
+			Eventually(func(g Gomega) {
+				g.Expect(testProvider.PreLoginBanner).To(BeNil(), "Provider PreLogin Banner should be nil")
+				g.Expect(testProvider.PostLoginBanner).To(BeNil(), "Provider PostLogin Banner should be nil")
+			}).Should(Succeed())
```
nikatza, line 63: I think this slipped through and all this test needs to be updated :)

internal/provider/cisco/nxos/testdata/esi_interface.json
```diff
@@ -0,0 +1,18 @@
...
+      "AggrIf-list": [
+        {
+          "id": "po10",
+          "layer": "Layer2",
+          "mode": "trunk",
+          "ethernetsegment-items": {
+            "esType": "native",
+            "esi": "0011.2233.4455.6677.8801",
+            "tag": "0"
+          }
```
nikatza, line 14: I'm a bit concerned because we have the `Interface` and the `ethernetsegment` reconciler touch the same path of the yang model (for NXOS). 

We are circumventing any issues by leveraging the per-device locking (controller), and making sure we use Update instead of Replace methods (provider). That could change in the future as we evolve the code base. 

I think we need a way to document and even better include a test to ensure this does not happen. The e2e tests could be a place (not sure if the controller tests would be also a place where add tests too since this is an NXOS only problem so far).

Do you have any thoughts on this topic we could address. May be in a follow-up?

internal/provider/cisco/nxos/provider.go
```diff
@@ -3300,6 +3272,62 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+	}
+
+	name, err := ShortName(req.Interface.Spec.Name)
+	if err != nil {
+		return err
+	}
+
+	parts := strings.Split(req.EthernetSegment.Spec.ESI, ":")
+	if len(parts) != 10 {
+		return apistatus.NewInvalidArgumentError(apistatus.FieldViolation{
```
nikatza: I think that cisco only supports type-0 (arbitrary) and type-3 (value that *can* be auto-generated):
https://www.ciscolive.com/c/dam/r/ciscolive/global-event/docs/2026/pdf/BRKDCN-2940.pdf

nikatza (changes requested):
Nice, many thanks for this PR! There are some issues that can we could resolve quickly, so I hope we can merge it soon. 

We may want to include the ethernet-segment routes in this API, however it would only mean an extension (and preferably in a follow-up).

Looking forward! 🔥

internal/controller/core/ethernetsegment_controller_test.go
felix-kaestner: Yes, this was still in copy+paste state (hence PR was still in draft).

internal/provider/cisco/nxos/testdata/esi_interface.json
felix-kaestner: I think the best would be if we have a way, to lint that when multiple controllers write into the same YANG (sub)tree, that the gnmi set must only use `Patch` calls. However, this would then definitively be a follow-up.

As we also have this situation in other places, would you be fine to defer that discussion in order to unblock this change for now? I think we need some more sophisticated solution here and put some more time in thinking about a good solution.

nikatza: Sure, I only wanted to raise attention to the matter.

api/core/v1alpha1/ethernetsegment_types.go
felix-kaestner: Where did you get `identifier auto lacp` from? Looking into [these slides](https://www.ciscolive.com/c/dam/r/ciscolive/global-event/docs/2026/pdf/BRKDCN-2940.pdf) LACP (Type-1) ESIs don't seem to be supported on Cisco NX-OS currently.

felix-kaestner: Adapted your recommendation, to support all ESI types (even the auto-generated ones) and have a status field being populated from the state of the device which holds the value of the ESI.

api/core/v1alpha1/ethernetsegment_types.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// ESI is the 10-byte Ethernet Segment Identifier in colon-separated hex notation
+	// (e.g., "00:11:22:33:44:55:66:77:88:01"). Must not be all-zeros or all-ones (reserved per RFC 7432).
+	// Required when ESIType is Arbitrary. Must be omitted when ESIType is LACP or MST.
+	// Optional for MAC, RouterID, and AS types (omit to auto-derive on the device).
+	// Immutable once set.
+	// +optional
+	// +kubebuilder:validation:Pattern=`^([0-9a-fA-F]{2}:){9}[0-9a-fA-F]{2}$`
+	// +kubebuilder:validation:XValidation:rule="size(oldSelf) == 0 || self == oldSelf",message="ESI is immutable once set"
+	ESI string `json:"esi,omitempty"`
```
swagner-de, line 62: We are missing the following exclusions:
RFC 7432, Section 5
```
   - ESI 0 denotes a single-homed site.

   - ESI {0xFF} (repeated 10 times) is known as MAX-ESI and is reserved.
```

api/core/v1alpha1/ethernetsegment_types.go
```diff
@@ -0,0 +1,232 @@
...
+
+	// ESI is the realized 10-byte Ethernet Segment Identifier on the device,
+	// in colon-separated hex notation. Populated from spec or read back from
+	// device when auto-generated.
+	// +optional
+	ESI string `json:"esi,omitempty"`
+
+	// ESIType is the ESI derivation type parsed from the first byte of ESI.
+	// +optional
+	ESIType ESIType `json:"esiType,omitempty"`
```
swagner-de, line 163: Wouldn't it make more sense to use a mutating webhook for this? After all this is immutable after creation, right?

swagner-de: Also saw [this thread](https://github.com/ironcore-dev/network-operator/pull/378#pullrequestreview-4523297453) now indicating that the device may not always support the requested state. I am wondering if we shouldn't go into configured=False then because intent!=configured.

internal/provider/cisco/nxos/provider.go
```diff
@@ -3328,6 +3330,151 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+
+	f := new(Feature)
+	f.Name = "evpn"
+	f.AdminSt = AdminStEnabled
+
+	mh := new(MultihomingItems)
+	mh.State = AdminStEnabled
+	mh.EadEviRoute = true
+	mh.DfElectionMode = DfElectModeModulo
+	mh.DfElectionTime = "3.0"
```
swagner-de, line 3381: Is this the device's default?

internal/provider/cisco/nxos/provider.go
```diff
@@ -3328,6 +3330,151 @@ func (p *Provider) GetDHCPRelayStatus(ctx context.Context, req *provider.DHCPRel
...
+			es.LocalIdentifier = uint32(d)
+		} else {
+			es.SysMacInherit = true
+			es.LocalIdentifierInherit = true
+		}
+	case v1alpha1.ESITypeLACP, v1alpha1.ESITypeMST, v1alpha1.ESITypeRouterID, v1alpha1.ESITypeAS:
+		return fmt.Errorf("ESI type %s is not supported by this provider", req.EthernetSegment.Spec.ESIType)
+	}
+
+	return p.Patch(ctx, f, mh, mm, es)
```
swagner-de, line 3414: Do you see any option to not patch all of these entities for each ES? Everything but `es` are singeltons, aren't they? I am not seeing them being cleaned up. Do you think a (vendored) global ESI Multi-Homing config would make sense?

api/core/v1alpha1/ethernetsegment_types.go
felix-kaestner: Added two CEL XValidation rules that reject `00:00:...:00` (single-homed, reserved) and `ff:ff:...:ff` (MAX-ESI, reserved) per RFC 7432 Section 5.

api/core/v1alpha1/ethernetsegment_types.go
felix-kaestner: `Status.ESIType` is populated from the device state (read-back from the operational ESI's first byte), not defaulted from user input. So a mutating webhook wouldn't apply here. The spec field `ESIType` is already immutable via CEL validation.

Regarding `Configured=False` for unsupported states: the provider returns an error for unsupported ESI types, which the controller surfaces as a terminal error and sets Configured=False. So this case is already covered.

internal/provider/cisco/nxos/provider.go
felix-kaestner: Yes, 3 seconds is the NX-OS default for the DF election hold timer. We set it explicitly so that the desired-state payload is fully declarative — otherwise a gNMI GET would return "3.0" from the device but our desired state would have no value, causing a perpetual reconcile diff.

Only `DfElectionMode=Modulo` is not device default, as it would be the Cisco specific `PerFlow`. However, the mode is contained in the spec, with a default value of `Modulo`, so we use that here (adhering to the spec).

felix-kaestner: `EadEviRoute` is also not device default, but we set it always to `true` as the cross-vendor default set by this project.

internal/provider/cisco/nxos/provider.go
felix-kaestner: You are right that some of these entities are global singletons on NX-OS and not per `ES`. However, the options like DF in openconfig are per `ES`, hence we added it also the spec.

```
/openconfig-ethernet-segments		
  ethernet-segments
    ethernet-segment[name]
      name
      config
      state
      df-election
        config
        df-election-method
        preference
        revertive
        election-wait-time
```

Things like the feature activation or the `EvpnMulticastItems` are static values, same for all ES, so setting those paths in every ES's reconcilation is not a conflict. Only the `DfElectionTime` in the `MultihomingItems` would lead to conflict between multiple ES resources.

swagner-de (approved):

PR #380, opened 2026-05-27
Implement EVPN Fabric underlay IGP reconciliation

felix-kaestner:
Add a reconcileUnderlayIGP phase that materialises one OSPF or ISIS resource per fabric device based on spec.underlay.protocol. Loopback and uplink interface references are accumulated across earlier phases via a ReconcileState struct, avoiding redundant API calls.

OSPF: instance UNDERLAY, loopbacks passive in area 0.0.0.0, uplinks active, logAdjacencyChanges enabled.

ISIS: Level2, OverloadBit=OnStartup, AddressFamilies=[IPv4Unicast], NET derived from the device lo0 IPv4 address.

### DIFF
--- a/internal/controller/evpn/fabric_controller.go
+++ b/internal/controller/evpn/fabric_controller.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"fmt"
 	"net/netip"
+	"slices"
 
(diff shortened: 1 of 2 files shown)


nikatza (approved):
Nice!

PR #381, opened 2026-05-27
Implement NX-OS reprovisioning via NX-API

felix-kaestner:
Replace the defunct GNMI-based reprovisioning with NXAPI JSON-RPC calls. The new implementation issues two requests:

1. boot poap enable + copy running-config startup-config (batched with stop-on-error rollback)
2. reload (separate request that tolerates transport errors because the device goes down before responding)

Add nxapi.IsTransportError to distinguish network-level errors (EOF, timeout, connection reset) from logical errors (RPCError, HTTPError), enabling callers of disruptive commands to tolerate expected connection drops.

Remove the now unused BootPOAP GNMI type.

### DIFF
--- a/internal/provider/cisco/nxos/provider.go
+++ b/internal/provider/cisco/nxos/provider.go
@@ -99,7 +99,7 @@ func (p *Provider) Connect(ctx context.Context, conn *deviceutil.Connection) (er
 	// NXAPI only uses the address for URI construction.
 	c := *conn
 	c.Address = netip.MustParseAddrPort(conn.Address).Addr().String()
-	p.nxapi, err = nxapi.NewClient(&c, timeout)
+	p.nxapi, err = nxapi.NewClient(&c, nxapi.WithTimeout(timeout))
 	if err != nil {
 		return fmt.Errorf("failed to create nxapi client: %w", err)
 	}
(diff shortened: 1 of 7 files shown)


internal/provider/cisco/nxos/provider.go
```diff
@@ -144,23 +142,35 @@ func (p *Provider) FactoryReset(ctx context.Context, conn *deviceutil.Connection
...
-			reterr = errors.Join(reterr, err)
+func (p *Provider) Reprovision(ctx context.Context, conn *deviceutil.Connection) error {
+	client := p.nxapi
+	if client == nil {
+		c := *conn
+		c.Address = netip.MustParseAddrPort(conn.Address).Addr().String()
+		var err error
+		client, err = nxapi.NewClient(&c, timeout)
+		if err != nil {
+			return fmt.Errorf("failed to create nxapi client: %w", err)
```
IvoGoman: This is the same as L94-L98. WDYT about a small shared func for client creation?

internal/transport/nxapi/nxapi.go
```diff
@@ -35,16 +36,44 @@ type Client struct {
...
+type options struct {
+	port string
+}
+
+// WithPort overrides the port in the connection address.
+// This is useful when NX-API is reachable on a different
+// port (e.g. 8443) than the default (80/443).
+func WithPort(port string) Option {
+	return func(o *options) { o.port = port }
+}
```
IvoGoman: This contradicts with how the option is implemented in #446 . Personally, I like the `type Option func(*Client)` from the other PR more.
But it might be a bit more involved to implement here with the uri construction.

felix-kaestner: Changed to the `type Option func(*Client)` format. Will later rebase #446.

IvoGoman (approved):

internal/provider/cisco/nxos/provider.go
felix-kaestner: I've refactored this code to do the nxapi client creation only once in the Connect/Disconnect and reuse it.

Issue #162, opened 2026-01-29
Cisco NX-OS: gNOI endpoint `LoadCertificateRequest` does not allow setting `ca_certificate`


swagner-de:
## Problem Statement

The Cisco NXOS gNOI API for the `LoadCertificateRequest` endpoint [(see here)](https://github.com/openconfig/gnoi/blob/main/cert/cert.proto#L302-L324) does not allow setting a `ca_certificate`, which prevents installing certificates from a private CA. The client cannot validate the chain of trust as it cannot be linked to the issuing CA because the field is missing. This restriction effectively limits certificate installation to self-signed certificates only.

[This script](https://gist.github.com/felix-kaestner/b423e7c9b87f6de2520c35bfdbf50475) can be run to trigger the condition.

## Vendor Acknowledgement
Cisco has labeled this issue as a feature request rather than a bug, but we argue that in conjunction with #161, the gNOI endpoint cannot handle private CA installations at all. See [CSCws33167](https://bst.cisco.com/bugsearch/bug/CSCws33167).

Additionally, this gNOI endpoint leaks private keys to the disk by running `crypto ca import $trustpoint pkcs bootflash:temp_pkcs_$hex_value cisco123` in the background, where `cisco123` is a static password used on every import.

## Workarounds
Provision and manage TLS certificates via ZTP and NX-API for better security and control over the certificate installation process.

swagner-de:
Tested on NX-OS prerelease build `10.7(0)IMG9(0.81)`, both issues now work with caveats.

We wrote a [test script](https://gist.github.com/swagner-de/0765ea2437ff88974558266ef2bad385) that exercises the full bootstrap flow end to end:

1. LoadCertificate with a CA-signed leaf cert (CN/SAN = device IP, ca_certificates field populated)
2. LoadCertificateAuthorityBundle to push the CA
3. gNMI Set on System/grpc-items to configure cert + certClientRoot + port
4. Verified mTLS gNMI Get with no InsecureSkipVerify, using a client cert signed by the same CA

**Results:**

```
❯ go run ./hack/gnoi-cert-test/ \
    -address $DEVICE:9339 \
    -username $USER \
    -password $PASSWORD \
    -trustpoint see-if-fixed
  saved ca.crt
  saved ca.key
  saved client.crt
  saved client.key
=== Step 1: LoadCertificate with CA-signed cert (CN/SAN = device IP) ===
[...]

PR #382, opened 2026-05-28
Add NX-API PKCS#12 workaround for certificate provisioning

felix-kaestner:
gNOI LoadCertificate on NX-OS does not support CA chains and creates immutable KeyPair entries that block rotation.

Work around both issues by encoding the certificate as PKCS#12 and importing it via NX-API CLI commands for older versions.

Also pass the CA chain to gNOI LoadCertificate for >= 10.7(1) which is expected to fix this issue (to be tested when released).

See https://github.com/ironcore-dev/network-operator/issues/162

### DIFF
--- a/internal/provider/cisco/nxos/cert.go
+++ b/internal/provider/cisco/nxos/cert.go
@@ -6,14 +6,15 @@ package nxos
 import (
 	"bytes"
 	"context"
-	"crypto/rand"
 	"crypto/rsa"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
 	"fmt"
 
-	"github.com/openconfig/gnoi/cert"
+	certpb "github.com/openconfig/gnoi/cert"
 	"google.golang.org/grpc"
+	pkcs12 "software.sslmate.com/src/go-pkcs12"
 
 	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
 )
@@ -21,8 +22,9 @@ import (
 // Certificate represents a X.509 certificate and its associated private key.
 // It can be used to load the certificate into a NX-OS device truspoint via gNOI.
 type Certificate struct {
-	Key  *rsa.PrivateKey
-	Cert *x509.Certificate
+	Key     *rsa.PrivateKey
+	Cert    *x509.Certificate
+	CACerts []*x509.Certificate
 }
 
 // Load loads the certificate into the specified trustpoint via the gNOI [cert service].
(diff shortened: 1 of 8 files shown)


internal/provider/cisco/nxos/version.go
```diff
@@ -6,22 +6,55 @@ package nxos
...
+	VersionNX10_4_5         // 10.4(5)
+	VersionNX10_4_6         // 10.4(6)
+	VersionNX10_5_1         // 10.5(1)
+	VersionNX10_5_2         // 10.5(2)
+	VersionNX10_5_3         // 10.5(3)
+	VersionNX10_6_1         // 10.6(1)
+	VersionNX10_6_2         // 10.6(2)
+	VersionNX10_6_3         // 10.6(3)
+
+	VersionNX10_7_1 // 10.7(1)
```
sven-rosenzweig: nit: indentation looks weird

felix-kaestner: This indentation is enforced by `go fmt` as the official/recommended style. ref/ https://go.dev/doc/effective_go#formatting

felix-kaestner: We also have a `make fmt` target which enforces that all source code is formatted correctly, which also applies here.

internal/provider/cisco/nxos/provider.go
```diff
@@ -667,27 +670,124 @@ func (p *Provider) GetPeerStatus(ctx context.Context, req *provider.BGPPeerStatu
...
+
+		logger.V(2).Info("Certificate upload completed")
+		return nil
 	}
 
-	kp := new(KeyPair)
-	kp.Name = req.ID
-	if err := p.client.GetConfig(ctx, kp); !errors.Is(err, gnmiext.ErrNil) {
-		// If the key pair already exists, we cannot update it, so we skip the rest of the process.
+	var pass [16]byte
```
sven-rosenzweig, line 719: Do we need to safe the password of the store somewhere? 
Does it make sense to move this to ints own method (probably we will not reuse it)?
What happens if we do not specify a password? is this mandatory?

internal/provider/cisco/nxos/provider.go
```diff
@@ -667,27 +670,124 @@ func (p *Provider) GetPeerStatus(ctx context.Context, req *provider.BGPPeerStatu
...
+			}
+			cert.CACerts = append(cert.CACerts, ca)
+		}
+
+		err := cert.Load(ctx, p.conn, req.ID)
+		if err != nil {
+			return fmt.Errorf("failed to upload certificate via gNOI: %w", err)
+		}
+
+		logger.V(2).Info("Certificate upload completed")
```
sven-rosenzweig, line 715: Maybe ``logger.V(2).Info("Certificate upload NX-OS version > 10.7 completed")``?

internal/provider/cisco/nxos/provider.go
```diff
@@ -667,27 +670,124 @@ func (p *Provider) GetPeerStatus(ctx context.Context, req *provider.BGPPeerStatu
...
+		return nil
 	}
```
sven-rosenzweig, line 718: nit: Adding a single comment ``configuring certificates for NX-OS Version <= 10.6.`` would help readability

internal/provider/cisco/nxos/cert.go
```diff
@@ -115,3 +122,29 @@ func (*KeyPair) IsListItem() {}
 func (k *KeyPair) XPath() string {
 	return "System/userext-items/pkiext-items/keyring-items/KeyRing-list[name=" + k.Name + "]"
 }
+
+// EncodeCertificatePKCS12 encodes a tls.Certificate (leaf + chain + private key) into
+// PKCS#12 format protected by the given passphrase. LegacyRC2 encoding is used
+// for compatibility with NX-OS.
+func EncodeCertificatePKCS12(cert *tls.Certificate, passphrase string) ([]byte, error) {
```
sven-rosenzweig, line 129: Just a felling of mine: The method placing feels a little bit weird as it has nothing to do with the nxos.Cert struct defined here. Honestly I do not know an better place for this method.

internal/provider/cisco/nxos/provider.go
felix-kaestner: > Do we need to safe the password of the store somewhere?

No. This is also not the password of the store, it's only the password of the pkcs12 file that is created on the fly and then imported. As this is a temporary file, no need to store. The complete certificate (chain) lives in kubernetes and is managed by cert-manager.

>  Does it make sense to move this to ints own method (probably we will not reuse it)?

I would say no, as long as we don't have a reason to reuse this (YAGNI)

>  What happens if we do not specify a password? is this mandatory?

Yes, this mandatory in the pkcs12 import command.

internal/provider/cisco/nxos/provider.go
felix-kaestner: I think I will rather add the precise version of the device as a tag into the logger. Then it's present in each log statement and shows the precise version.

internal/provider/cisco/nxos/provider.go
felix-kaestner: See https://github.com/ironcore-dev/network-operator/pull/382#discussion_r3622553981. Added the nx-version key to all log statements.

internal/provider/cisco/nxos/cert.go
felix-kaestner: In Go, a file might not solely be dedicated to a single type/struct. Rather to a concept/idea. Therefore I moved it into the `cert.go` which is the home for everything around certificates and their encoding.

sven-rosenzweig (approved):

PR #386, opened 2026-05-29
Fix stale error leak in EnsureBGP causing "gnmiext: nil"

felix-kaestner:
When the BGP domain does not exist on the device (first-time config) and the ASN uses plain notation, the outer err variable retained the ErrNil value from the domain GetConfig call. Neither switch case for AS format fired, so err was never reassigned and leaked to the caller.

Scope error handling into each switch case so that the shared err variable is no longer carried across unrelated operations.

Also skip the router-ID conflict check when the existing domain has an empty router-ID, avoiding a false-positive mismatch error.


SchwarzM (approved):
lgtm

PR #391, opened 2026-06-02
[NX-OS] Fix invalid asFormat value

felix-kaestner:
The l3vm_AsFormatType YANG enum expects "asdot" but the constant was set to "as-dot" (with a hyphen), causing the device to reject the Set RPC with an invalid-value error.


SchwarzM (approved):
lgtm

nikatza (approved):

PR #396, opened 2026-06-04
fix(interface): match labels against FQDN in LLDP validation

nikatza:
In this commit we change the validation of the label against the
information received via LLDP. If DNS is configured on the remote device
via a `DNS` Kubernetes resource, the interface controller fetches that
resource and uses the `domain` field together with the device `hostname`
to build an FQDN that is then compared to the LLDP systemName. This
change now captures devices that export the FQDN in LLDP.

Add DNS watch to interface controller for LLDP neighbor re-validation.
When a DNS resource is created, deleted, or has its domain changed,
interfaces with neighbor labels pointing to that device's interfaces are
re-enqueued so LLDP adjacency validation re-evaluates the expected FQDN.
Also fix SetLLDPNeighbor test helper to set ChassisIDType/PortIDType so
LLDP adjacencies are no longer silently skipped in integration tests.

Replace conditions-length heuristic in interfaceUpdatePredicate with
explicit finalizer and deletion-timestamp checks. Adding a finalizer is
a metadata-only change that no standard predicate (Generation, Label,
Annotation) catches, so it falls to interfaceUpdatePredicate. The old
len(conditions) < 4 guard happened to let these through because
conditions are empty during bootstrap, but expressed the wrong
invariant. The new checks explicitly pass the lifecycle transitions that
the controller's multi-step bootstrap (finalizer → init conditions →
reconcile) depends on.

### DIFF
--- a/internal/controller/core/interface_controller.go
+++ b/internal/controller/core/interface_controller.go
@@ -17,7 +17,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/events"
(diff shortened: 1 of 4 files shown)


internal/controller/core/interface_controller.go
```diff
@@ -444,12 +463,16 @@ func (interfaceUpdatePredicate) Update(e event.UpdateEvent) bool {
...
-	// InitializeConditions adds Ready/Configured/Operational, and paused.EnsureCondition
-	// adds Paused. Until all 4 are present, we must allow reconciles to complete setup.
-	if len(newIntf.Status.Conditions) < 4 {
+
+	// Allow lifecycle metadata transitions to pass through so the controller can
+	// continue after adding/removing finalizers and when deletion starts.
+	if !equality.Semantic.DeepEqual(oldIntf.GetFinalizers(), newIntf.GetFinalizers()) {
 		return true
 	}
+	if !equality.Semantic.DeepEqual(oldIntf.GetDeletionTimestamp(), newIntf.GetDeletionTimestamp()) {
```
felix-kaestner:
```suggestion
	if !equality.Semantic.DeepEqual(oldIntf.GetFinalizers(), newIntf.GetFinalizers()) || !equality.Semantic.DeepEqual(oldIntf.GetDeletionTimestamp(), newIntf.GetDeletionTimestamp()) {
```
nit: could combine into a single if statement.

internal/controller/core/interface_controller.go
```diff
@@ -1344,7 +1378,7 @@ func (r *InterfaceReconciler) interfacesForProviderConfig(ctx context.Context, o
 
 	list := &v1alpha1.InterfaceList{}
 	if err := r.List(ctx, list, client.InNamespace(obj.GetNamespace())); err != nil {
-		log.Error(err, "Failed to list Interfacees")
+		log.Error(err, "Failed to list Interfaces")
```
felix-kaestner, line 1378: Nice catch!

felix-kaestner (approved):

PR #397, opened 2026-06-04
Remove go-makefile-maker and continue with a custom Makefile

### LINKED ISSUE #6: Use `kubebuilder` style `Makefile`
# Suggested changes

Can we revert the `Makefile` to a `kubebuilder` style version? (an example is the one we are using here https://github.com/ironcore-dev/metal-operator/blob/main/Makefile)

felix-kaestner:
At the current point of time, we have disabled most of the features go-makefile-maker brings, like Dockerfile, REUSE, Nix, custom golangci-lint configuration and GitHub Action workflow generation due to customizations we like to perform ourselves. Additionally, go-makefile-maker unconditionally generates files like a Renovate configuration and a CodeQL workflow that are not wanted in this project: we use Dependabot for dependency updates and CodeQL is enabled as a repository-wide setting that runs automatically without a dedicated workflow file. When running the tool, a contributor would have to know to discard these unwanted files before committing, which makes the experience cumbersome.

The remaining scope of go-makefile-maker in this project has been reduced to generating lint, test and build targets for a controller-runtime based project, along with a handful of install-* targets for tooling. However, the `verbatim` section of the setup has grown to a size where the majority of the Makefile targets come from custom code anyway.

As a result, it feels like the right approach to remove go-makefile-maker and go forward with a custom Makefile, based on kubebuilders Makefile. This gives us the flexibility to adjust and extend Makefile targets in the future and removes yet another dependency used by the project that contributors will have to know/learn to be able to contribute.

Another advantage of the kubebuilder-style Makefile is that it pins all tool dependencies to explicit versions (e.g. kubebuilder v4.13.1, controller-gen v0.20.1, golangci-lint v2.11.4). The previous setup installed tools with `@latest`, which has caused CI breakages in the past when a new incompatible release of a tool was published. Pinning versions makes builds reproducible and prevents unexpected failures in CI.

This comes at the cost of manually maintaining related files.

Closes #6


cmd/main.go
```diff
@@ -19,7 +20,6 @@ import (
 	// Set runtime concurrency to match CPU limit imposed by Kubernetes
 	_ "go.uber.org/automaxprocs"
 
-	"github.com/sapcc/go-api-declarations/bininfo"
```
felix-kaestner, line 22: Removing such SAP-specific dependency from the project seems like the right path for a ApeiroRA project that has the desire to be openly and independently governed.

.goreleaser.yaml
```diff
@@ -1,5 +1,5 @@
 # yaml-language-server: $schema=https://goreleaser.com/static/schema.json
-# SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company
+# SPDX-FileCopyrightText: 2020 SAP SE or an SAP affiliate company and IronCore contributors
```
nikatza: nit: 2020 -> 2026

nikatza (approved):

Makefile
```diff
@@ -1,405 +1,424 @@
...
+
+.PHONY: test
+test: manifests generate fmt vet setup-envtest ## Run tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e | grep -v /lab) -coverprofile cover.out
+
+.PHONY: coverage
+coverage: test ## Run tests and generate coverage report.
+	go tool cover -html=cover.out -o cover.html
+
+# TODO(felix-kaestner): To use a different vendor for e2e tests, modify the setup under 'tests/e2e'.
```
nikatza: `e2e` tests don`t use providers, I think these comment can be deleted.

nikatza (approved):

Makefile
```diff
@@ -1,405 +1,424 @@
...
+# E2E test dependency versions
+E2E_PROMETHEUS_OPERATOR_VERSION ?= v0.82.2
+E2E_CERTMANAGER_VERSION ?= v1.17.2
 
-lint: FORCE bin/golangci-lint-custom ## Run golangci-lint linter
-	@printf "\e[1;36m>> golangci-lint\e[0m\n"
-	@bin/golangci-lint-custom config verify
-	@bin/golangci-lint-custom run
+.PHONY: test-e2e
+test-e2e: setup-test-e2e manifests generate fmt vet ## Run the e2e tests. Expected an isolated environment using Kind.
```
nikatza: do we really want `fmt` and `vet`?

nikatza: (same question for the main `test` target)

felix-kaestner: Removed.

Makefile
```diff
@@ -1,405 +1,422 @@
...
+test-gnmi: FORCE ## Run integration tests for gNMI.
 	@printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
 
-docker-build: FORCE
-	@printf "\e[1;36m>> $(CONTAINER_TOOL) build --tag=$(IMG) .\e[0m\n"
-	@$(CONTAINER_TOOL) build --build-arg=BININFO_BUILD_DATE=$(BININFO_BUILD_DATE) --build-arg=BININFO_COMMIT_HASH=$(BININFO_COMMIT_HASH) --build-arg=BININFO_VERSION=$(BININFO_VERSION) --tag=$(IMG) .
-
-docker-push: FORCE
-	@printf "\e[1;36m>> $(CONTAINER_TOOL) push $(IMG)\e[0m\n"
-	@$(CONTAINER_TOOL) push $(IMG)
```
adamtrizuljak-sap, line 105: I feel like `docker-push` and `docker-push-test-lab` could be useful to keep. But I'm not yet deeply familiar with all of your workflows, so you probably have better judgment here :)

adamtrizuljak-sap (approved):

Makefile
felix-kaestner: I intentionally left this out, as I feel pushing the image should be exclusively be done by our workflow we have setup for this. And the lab image we currently don't deploy/push anywhere.

IvoGoman (approved):

PR #400, opened 2026-06-10
Update `e2e` tests

nikatza:
# E2e tests update

Update tests to validate the full reconciliation pipeline from Kubernetes CRD to gNMI JSON output.

## Rationale

Unit tests verify individual functions but miss the gaps between layers: CRD validation, controller logic, provider field mapping, and gNMI path construction. For example, a controller passing the wrong value to a provider.

The current e2e tests cover this but rely on Kind clusters with a full operator installation. This PR refactors both test modes to share the same testdata-driven approach:

- **Envtest mode** — runs without infrastructure using envtest + in-process gnmi-test-server. Tests run sequentially but complete in seconds.
- **Cluster mode** — runs against Kind with full operator installation. Validates RBAC, webhooks, metrics. Each test deploys its own gnmi-test-server pod in its own namespace.

## Test Format

Each test is a txtar file in `test/e2e/testdata/<provider>/`:

```
-- state/preload --              # OPTIONAL: initial gNMI state
{"System": {"procsys-items": {"bootTime": "1700000000"}}}

-- Interface/lo0 --
kind: Interface
spec:
  deviceRef:
    name: device                 # substituted at runtime
  ...

-- state/expect --
{"System": {"intf-items": ...}}
```

Resources are created sequentially, waiting for `Configured` (or `Ready` if the former does not exist) condition before proceeding. This handles dependencies between resources (e.g., BGPPeer requires BGP).

> **Note:** The JSON stored in gnmi-test-server may differ from actual provider output. For example, NXOS `DME_UNSET_PROPERTY_MARKER` values are filtered out by the test server.

## Usage

```bash
# Fast, no cluster required
make test-e2e-envtest PROVIDER=cisco-nxos-gnmi

# Full cluster mode (requires Kind)
make test-e2e PROVIDER=cisco-nxos-gnmi
```

## Key Changes

- Unified `TestEnvironment` interface in `testutil/` — both modes implement the same contract
- Cluster mode now deploys a gnmi-test-server pod per test for isolation
- Shared testdata discovery — both modes auto-discover `test/e2e/testdata/<provider>/*.txt`
- Added `PreloadGNMIState()` for setting initial device state before reconciliation
- Added `state/preload` txtar section for tests that need pre-existing gNMI state
- Add more samples


nikatza:
Closing this PR as it is replaced by smaller ones.

PR #407, opened 2026-06-16
 e2e-tests (1/4): Repackage gnmi-test-server and `e2e` test infrastructure

nikatza:
e2e tests should only evaluate the correct deployment of the operator as defined by the kubebuilder markers. Hence, we remove all the gNMI related infrastructure from the e2e tests.

This PR effectively removes the current tests which are to be reincorporated in a follow-up PR


charts/network-operator/templates/prometheus/controller-manager-metrics-monitor.yaml
```diff
@@ -20,22 +20,7 @@ spec:
     scheme: {{ if .Values.metrics.secure }}https{{ else }}http{{ end }}
     {{- if .Values.metrics.secure }}
     tlsConfig:
-      serverName: {{ include "network-operator.resourceName" (dict "suffix" "controller-manager-metrics-service" "context" $) }}.{{ .Release.Namespace }}.svc
```
felix-kaestner, line 23: What is the reason this got removed? 🤔

config/develop/kustomization.yaml
felix-kaestner, line 1: I think this resource was also referenced in the Tiltfile, so it also has to be deleted there. https://github.com/ironcore-dev/network-operator/blob/main/Tiltfile#L23

test/e2e/e2e_test.go
```diff
@@ -33,45 +31,17 @@ const metricsRoleBindingName = "network-operator-metrics-binding"
...
-		_, err = Run(cmd)
-		Expect(err).NotTo(HaveOccurred())
-
-		cmd = exec.CommandContext(
-			ctx, "kubectl", "get", "pod", "gnmi-test-server",
-			"--output", "jsonpath='{.status.podIP}'",
-			"--namespace", "default",
-		)
+		By("creating manager namespace")
+		cmd := exec.CommandContext(ctx, "kubectl", "create", "ns", namespace, "--dry-run=client", "-o", "yaml")
```
felix-kaestner, line 40: Instead of `dry-run=client -o yaml` to then pass it to `kubectl apply`, could we simply check if the namespace exists with `kubectl get ns` and only create it, when it doesn't exist yet?

test/e2e/e2e_test.go
```diff
@@ -311,113 +280,6 @@ var _ = Describe("Manager", Ordered, func() {
...
 		// +kubebuilder:scaffold:e2e-webhooks-checks
-
-		// TODO: Customize the e2e test suite with scenarios specific to your project.
-		// Consider applying sample/CR(s) and check their status and/or verifying
-		// the reconciliation by using the metrics, i.e.:
-		// metricsOutput := getMetricsOutput()
-		// Expect(metricsOutput).To(ContainSubstring(
-		//    fmt.Sprintf(`controller_runtime_reconcile_total{controller="%s",result="success"} 1`,
-		//    strings.ToLower(<Kind>),
-		// ))
```
felix-kaestner, line 322: Shall we keep this comment from kubebuilder?

Makefile.maker.yaml
```diff
@@ -122,7 +126,19 @@ verbatim: |
...
-    @KIND_CLUSTER=$(KIND_CLUSTER) go test ./test/e2e/ -v -ginkgo.v
+    @KIND_CLUSTER=$(KIND_CLUSTER) \
+      E2E_PROMETHEUS_OPERATOR_VERSION=$(E2E_PROMETHEUS_OPERATOR_VERSION) \
+      E2E_CERTMANAGER_VERSION=$(E2E_CERTMANAGER_VERSION) \
+      go test ./test/e2e/ -v -ginkgo.v
+
+  # Run gNMI server unit tests
+  test-gnmi-unit: FORCE
+    @printf "\e[1;36m>> go test ./test/gnmi/testserver/...\e[0m\n"
+    @go test ./test/gnmi/testserver/...
```
felix-kaestner: I guess these tests would already be run by the 'make build/cover.out' target. What is the reason to run these unit tests in isolation?

test/gnmi/testserver/server.go
```diff
@@ -36,8 +36,115 @@ var _ gpb.GNMIServer = (*Server)(nil)
...
+		httpAddr:   httpLis.Addr().String(),
+	}
+
+	gpb.RegisterGNMIServer(grpcServer, server)
+
+	reflection.Register(grpcServer)
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/v1/state", server.handleState)
+	server.httpServer = &http.Server{Handler: mux}
```
felix-kaestner: Any reason to still have this as a http mux instead of just exposing a method on the returned `*Server` to obtain the state?

test/gnmi/testserver/server.go
```diff
@@ -36,8 +36,115 @@ var _ gpb.GNMIServer = (*Server)(nil)
...
+	go func() { //nolint:gosec // G118: ctx is already done, must use Background for shutdown timeout
+		<-ctx.Done()
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := server.Close(shutdownCtx); err != nil { //nolint:contextcheck // shutdownCtx is correctly derived from Background
+			log.Printf("Shutdown error: %v", err)
+		}
+	}()
+
+	return server, server.grpcAddr, server.httpAddr, nil
```
felix-kaestner: What is the reasons we return these addrs, if they call already be retrieved by calling `GRPCAddr()`?

test/e2e/e2e_test.go
nikatza: The use case for the dry-run is a stretched but someone could delete the namespace between the `get ns` and the `create` call. The current approach doesn't fail unless there is a problem with the cluster.

I also preferred this approach because having a workflow that relies on checking the state via `kubectl` doesn't seem right (if we want to go that way I would try to move the tests to use a k8s library for golang).

test/gnmi/testserver/server.go
nikatza: Indeed. I removed it and left a `GetState()` method. The tests will use call it to verify that the payload is correctly processed; meaning the method is unused in this commit.

Makefile.maker.yaml
nikatza: I found some issues with the server while emulating `nxos` behavior and I added to just run those tests. Having this target as a requirement for the `test-gnmi` target also was convenient to spot issues and have the `gnmi` tests self-contained. That said, now that this part is almost over we could certainly remove the target. Just let me know if this is still your preference and i will remove it.

charts/network-operator/templates/prometheus/controller-manager-metrics-monitor.yaml
nikatza: This change was auto-generated via `make charts` and `controller-gen@v0.21.0`

nikatza: It's due to uncommenting prometheus in `config/default/kustomization.yaml` to avoid having to programmatically uncomment it while running the e2e tests. I will revert that change.

felix-kaestner: Hmm, I didn't expect this to then change the tlsconfig of the ServiceMonitor in the generated chart..

test/e2e/e2e_test.go
felix-kaestner: Fair point. Then let's stick with the dry-run-apply.

Makefile.maker.yaml
felix-kaestner: I think we don't have unit test targets for individual packages in other places and as such, I usually when developing a package in isolation just run `go test ..` to test a specific package. So I would slightly lean towards not adding this. The command is simple enough that someone can just put it together on the fly I'd say.

test/gnmi/testserver/server.go
felix-kaestner: nit: the canonical go style would be to call it just `State()` and avoid the `Get` prefix. ref/ https://google.github.io/styleguide/go/decisions#getters https://google.github.io/styleguide/go/best-practices.html#naming-conventions

nikatza: Fixed, and also renamed the `testserver` package to just `server` since this is implied already in the path.

test/gnmi/server/server.go
```diff
@@ -33,25 +31,136 @@ import (
...
+	// state is the internal state of the server, which can be manipulated via gNMI Set requests or the HTTP API.
+	state *State
+	// grpcServer is the gRPC server instance where gNMI clients can connect to.
+	grpcServer *grpc.Server
+	// grpcAddr is the address grpcServer is listening on, e.g., 127.0.0.1:9443
+	grpcAddr string
+	// httpServer is the HTTP server instance used to inspect and manipulate the server's internal state.
+	httpServer *http.Server
+	// httpAddr is the address httpServer is listening on, e.g., 127.0.0.1:8000
+	httpAddr string
```
felix-kaestner: Any reasons we still create the http listener here?

felix-kaestner (approved):

test/gnmi/server/server.go
felix-kaestner: Can also be cleaned up in a follow-up.

nikatza: Removed.

test/e2e/e2e_suite_test.go
```diff
@@ -107,3 +97,50 @@ var _ = AfterSuite(func(ctx SpecContext) {
 		UninstallCertManager(ctx)
 	}
 })
+
+// UncommentCode searches for target in the file and remove the comment prefix
```
felix-kaestner: As we kept this function, should we also keep it in the utils_test.go file it was originally?

nikatza: Done.

PR #409, opened 2026-06-16
testing (3/4): Add  gNMI integration tests with `envtest`

nikatza:
This PR introduces envtest-based gNMI integration tests for the OpenConfig provider, enabling faster validation of the Kubernetes resources to gNMI sequence without requiring a full cluster deployment.

The test architecture uses a per-process gNMI server design that enables parallel test execution with Ginkgo's --procs flag. Each process creates its own suite-level server in BeforeSuite and shares it across tests running serially within that process.


.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,30 @@
...
+
+  name: Test gNMI integration
+
+  on:
+    pull_request:
+      branches:
+        - main
+      paths-ignore:
+        - 'docs/**'
+        - '**/*.md'
```
IvoGoman: Proposal for a bit more aggressive filtering. This should only trigger e2e on code or dependency changes.
```suggestion
      paths:
        - '**.go'
        - 'hack/**'
        - 'config/**'
        - 'go.mod'
        - 'go.sum'
        - 'Makefile'
```

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	envProvider := os.Getenv(ProviderEnvVar)
+	supportedProviders := provider.Providers()
+
+	// If provider is invalid, create a failing test with clear message
+	if envProvider == "" {
+		It("requires PROVIDER to be set", func() {
+			Fail(ProviderEnvVar + " not set. Please set " + ProviderEnvVar + " to one of: " + strings.Join(supportedProviders, ", "))
+		})
+		return
+	}
```
IvoGoman: This can be rewritten using only gomega matchers:

```suggestion
	Expect(envProvider).NotTo(BeEmpty(), "PROVIDER environment variable must be set to one of: %s", strings.Join(supportedProviders, ", "))
```

Same for some other if checks in this file

test/gnmi/json_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt"
+	"reflect"
+	"sort"
+)
+
+// CompareJSON compares two JSON strings and returns an error if they are not equal.
+// For comparison, it unmarshals both into interface{} and uses reflect.DeepEqual
+// after sorting any arrays and removing empty arrays/objects to ignore ordering
+// and cleanup artifacts.
+func CompareJSON(got, want string) error {
```
IvoGoman: Have you tried the gomega build-in matcher for JSON? https://onsi.github.io/gomega/#matchjsonjson-any

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,30 @@
+  # SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+  # SPDX-License-Identifier: Apache-2.0
+
+  name: Test gNMI integration
```
felix-kaestner:
```suggestion
name: Test gNMI integration
```
why does this file have a two space indentation by default?

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,30 @@
...
+    test-gnmi:
+      name: Run gNMI Integration Tests (OpenConfig)
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v7
+        - uses: actions/setup-go@v6
+          with:
+            go-version-file: 'go.mod'
+        - name: Install ginkgo
+          run: go install github.com/onsi/ginkgo/v2/ginkgo@latest
```
felix-kaestner: Is this install step required if the 'make test-gnmi' already takes care of installing ginkgo if it's not present?

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,30 @@
...
+      paths-ignore:
+        - 'docs/**'
+        - '**/*.md'
+
+  permissions:
+    contents: read
+
+  jobs:
+    test-gnmi:
+      name: Run gNMI Integration Tests (OpenConfig)
```
felix-kaestner:
```suggestion
      name: gNMI Integration (${{ matrix.provider }})
      strategy:
        fail-fast: false
        matrix:
          provider:
            - openconfig
```
Should we already integrate a matrix build, such that it is easier to integrate more providers later?

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,30 @@
...
+      runs-on: ubuntu-latest
+      steps:
+        - uses: actions/checkout@v7
+        - uses: actions/setup-go@v6
+          with:
+            go-version-file: 'go.mod'
+        - name: Install ginkgo
+          run: go install github.com/onsi/ginkgo/v2/ginkgo@latest
+        - name: Running gNMI tests
+          run: make test-gnmi TESTGNMI_PROVIDER=openconfig GINKGO_PROCS=4
```
felix-kaestner:
```suggestion
          run: make test-gnmi TESTGNMI_PROVIDER=openconfig GINKGO_PROCS=$(nproc)
```
why not let the platform use all available?

test/gnmi/server/server.go
```diff
@@ -105,24 +112,19 @@ func (s *Server) State() *State {
...
 // and from multiple goroutines; only the first call performs shutdown.
-func (s *Server) Close(ctx context.Context) error {
-	var closeErr error
+func (s *Server) Close() {
 	s.closeOnce.Do(func() {
 		log.Printf("Shutting down gNMI test server")
-
 		if s.grpcServer != nil {
 			s.grpcServer.GracefulStop()
 		}
```
felix-kaestner:
```suggestion
		if s.grpcServer != nil {
			const timeout = 5 * time.Second
			t := time.AfterFunc(timeout, func() {
				log.Printf("Graceful shutdown timed out, forcing stop")
				s.grpcServer.Stop()
			})
			s.grpcServer.GracefulStop()
			t.Stop()
		}
```
What about this? This completely removes the need to create a `done chan` and spawing a gorountine in a goroutine. So the go func in the `NewTestServer` can just become:

```go
	go func() {
		<-ctx.Done()
		server.Close()
	}()
```

test/gnmi/server/server.go
```diff
@@ -105,24 +112,19 @@ func (s *Server) State() *State {
...
 	})
-	return closeErr
 }
 
-// Capabilities returns the capabilities of the gNMI server
```
felix-kaestner, line 120: Why remove these comments on all the exported/public methods?

Makefile.maker.yaml
```diff
@@ -127,13 +127,19 @@ verbatim: |
...
-      E2E_CERTMANAGER_VERSION=$(E2E_CERTMANAGER_VERSION) \
+      E2E_PROMETHEUS_OPERATOR_VERSION=$(TESTE2E_PROMETHEUS_OPERATOR_VERSION) \
+      E2E_CERTMANAGER_VERSION=$(TESTE2E_CERTMANAGER_VERSION) \
       go test ./test/e2e/ -v -ginkgo.v
 
-  # Run integration tests for gNMI.
-  test-gnmi: FORCE
-    @printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
+  # Provider for gNMI integration tests
+  TESTGNMI_PROVIDER ?= openconfig
```
felix-kaestner: Why not use the `PROVIDER` environment variable that is already used by the Tiltfile to start the operator on? Then this would become just the same provider that is already used for deploying/starting the operator (default: openconfig).

Makefile.maker.yaml
```diff
@@ -127,13 +127,19 @@ verbatim: |
...
+  # Number of parallel Ginkgo processes
+  GINKGO_PROCS ?= 4
```
felix-kaestner:
```suggestion
GINKGO_PROCS ?= $(shell nproc 2>/dev/null || sysctl -n hw.logicalcpu) 
```
Why should we default to `4`?

test/gnmi/gnmi_suite_test.go
```diff
@@ -0,0 +1,37 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package gnmi_test
```
felix-kaestner: > Go package names should not have underscores.
ref/ https://google.github.io/styleguide/go/decisions.html#package-names

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/events"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/envtest"
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/yaml"
```
felix-kaestner, line 25: These import blocks should be combined. It should be 3 import blocks:
1. std
2. third party
3. local

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+
+// Provider — resolved from PROVIDER env var during test tree construction
+var (
+	providerName string
+	providerFunc provider.ProviderFunc
+)
+
+// Resources are the API resources with finalizers.
+// During cleanup, these are deleted FIRST so their finalizers can
+// complete while Device and config resources still exist.
```
felix-kaestner: You can also just use foreground cascading deletion on the device: https://kubernetes.io/docs/tasks/administer-cluster/use-cascading-deletion/#use-foreground-cascading-deletion. Then this is already implied by the owner reference.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	v1alpha1.GroupVersion.WithKind("PrefixSet"),
+	v1alpha1.GroupVersion.WithKind("BGP"),
+	v1alpha1.GroupVersion.WithKind("BGPPeer"),
+	v1alpha1.GroupVersion.WithKind("Syslog"),
+	v1alpha1.GroupVersion.WithKind("SNMP"),
+	v1alpha1.GroupVersion.WithKind("ManagementAccess"),
+	v1alpha1.GroupVersion.WithKind("AccessControlList"),
+	v1alpha1.GroupVersion.WithKind("DHCPRelay"),
+	v1alpha1.GroupVersion.WithKind("ISIS"),
+	nxv1alpha1.GroupVersion.WithKind("VPCDomain"),
```
felix-kaestner: I worry that this list will get outdated quite quickly. In fact, it is already (doesn't include AAA).

Could we follow an approach like [here](https://github.com/ironcore-dev/network-operator/blob/cli/kubectl-net/tools/generate-resources/main.go) were we generate these from source code, such that these are generated automatically?

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+var ConfigResources = []schema.GroupVersionKind{
+	nxv1alpha1.GroupVersion.WithKind("InterfaceConfig"),
+	nxv1alpha1.GroupVersion.WithKind("LLDPConfig"),
+	nxv1alpha1.GroupVersion.WithKind("BGPConfig"),
+	nxv1alpha1.GroupVersion.WithKind("NVEConfig"),
+	nxv1alpha1.GroupVersion.WithKind("ManagementAccessConfig"),
+}
+
+func init() {
+	_, _ = fmt.Fprintf(GinkgoWriter, "Starting network-operator gNMI tests\n")
```
felix-kaestner: Why put this in an init-func instead of just putting it in the Test func itself. Feels a bit unnecessarily complex.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+}
+
+// initTestEnv initializes the envtest environment and starts the controller manager.
+func initTestEnv(ctx SpecContext) {
```
felix-kaestner: I would honestly just inline the `initTestEnv` and the `cleanupTestEnv` into the `BeforeSuite` and `AfterSuite` respectively. There are only called from a single place so I feel like moving them to another func just unnecessarily creates complexity.


Same basically also applies for `startControllerManager`.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+		Metrics: metricsserver.Options{BindAddress: "0"}, // Disable metrics server
+	})
+	Expect(err).ToNot(HaveOccurred())
+
+	// Ignore events during tests
+	recorder := events.NewFakeRecorder(0)
+	go func() {
+		for range recorder.Events { //nolint:revive // intentionally drain events
+		}
+	}()
```
felix-kaestner: Why do we need this? With `0` the channel is unbuffered and as such we don't need to drain it (I think).

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+		},
+	}
+	if err := c.Create(ctx, device); err != nil {
+		return nil, err
+	}
+
+	device.Status.Phase = v1alpha1.DevicePhaseRunning
+	if err := c.Status().Update(ctx, device); err != nil {
+		return nil, err
+	}
```
felix-kaestner: I would expect that we don't even need that, as the Device (without provisioning set) should transition from Pending → Running on first reconcile. https://github.com/ironcore-dev/network-operator/blob/main/internal/controller/core/device_controller.go#L126-L130

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+}
+
+var _ = Describe("gNMI requests tests", func() {
+	// Resolve provider during tree construction so we can generate individual It nodes
+	envProvider := os.Getenv(ProviderEnvVar)
+	supportedProviders := provider.Providers()
+
+	// If provider is invalid, create a failing test with clear message
+	if envProvider == "" {
+		It("requires PROVIDER to be set", func() {
```
felix-kaestner: Is it required to be wrapped in an `It`? This feels rather wrong, as the `It` has no real purpose here. Same applies to other cases.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+				},
+			}
+			Expect(k8sClient.Create(ctx, ns)).To(Succeed())
+			testNamespace = ns.Name
+		})
+
+		AfterEach(func(ctx SpecContext) {
+			By("cleaning up resources")
+			cleanupAllResources(k8sClient, device, testNamespace)
+			device = nil
```
felix-kaestner: It feels somewhat wrong that the `AfterEach` resets the device variable to `nil` but the device is not created in the `BeforeEach`.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	// Tests run in parallel - each test gets its own namespace
+	Describe("Provider: "+providerName, func() {
+		var device *v1alpha1.Device
+		var testNamespace string
+
+		BeforeEach(func(ctx SpecContext) {
+			By("creating dedicated test namespace")
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: reconcileTestNamespacePrefix,
```
felix-kaestner: nit: I think I would just inline the string constant here.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+
+	// Set the namespace
+	obj.SetNamespace(namespace)
+
+	// Update deviceRef.name to use the actual device name
+	if spec, ok := obj.Object["spec"].(map[string]any); ok {
+		if deviceRef, ok := spec["deviceRef"].(map[string]any); ok {
+			deviceRef["name"] = deviceName
+		}
+	}
```
felix-kaestner:
```suggestion
	unstructured.SetNestedField(obj.Object, deviceName, "spec", "deviceRef")
```
I believe this is an easier way to achieve this.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	// Also update the device label
+	labels := obj.GetLabels()
+	if labels != nil {
+		if _, ok := labels[v1alpha1.DeviceLabel]; ok {
+			labels[v1alpha1.DeviceLabel] = deviceName
+			obj.SetLabels(labels)
+		}
+	}
```
felix-kaestner:
```suggestion
```
Not needed, the controller will take care of that.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+func waitForResource(ctx SpecContext, c client.Client, obj client.Object) {
+	key := client.ObjectKeyFromObject(obj)
+	gvk := obj.GetObjectKind().GroupVersionKind()
+
+	// Config resources don't have status conditions
+	if slices.ContainsFunc(ConfigResources, func(cfg schema.GroupVersionKind) bool {
+		return cfg.Kind == gvk.Kind
+	}) {
+		return
+	}
```
felix-kaestner: I guess we would also be able to clasify these by simply checking if a resource has a status subresource/field. This way, we could also support creating things like secrets (think for the trustpoint configuration).

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+
+	Eventually(func(g Gomega) {
+		r := &unstructured.Unstructured{}
+		r.SetGroupVersionKind(gvk)
+		g.Expect(c.Get(ctx, key, r)).To(Succeed())
+
+		conditions, err := ExtractConditions(r)
+		g.Expect(err).NotTo(HaveOccurred())
+
+		conditionToCheck := string(v1alpha1.ReadyCondition)
```
felix-kaestner, line 209: Should we also check for having status conditions? Because when the resource is created it won't initially have any conditions, those are then initialized on the first reconcile and set to "Unknown".

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+
+		if len(list.Items) > 0 {
+			Eventually(func(g Gomega) {
+				var remaining unstructured.UnstructuredList
+				remaining.SetGroupVersionKind(list.GroupVersionKind())
+				g.Expect(c.List(cleanupCtx, &remaining, client.InNamespace(namespace))).To(Succeed())
+				g.Expect(remaining.Items).To(BeEmpty())
+			}).WithContext(cleanupCtx).Should(Succeed())
+		}
+	}
```
felix-kaestner: I guess a 

```	
c.Delete(cleanupCtx, device, client.PropagationPolicy(metav1.DeletePropagationForeground))
```

should already work for deleting in the right order. 

If that is not the case, then we clearly have some internal errors elsewhere.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+	startControllerManager(controllerCtx)
+}
+
+// cleanupTestEnv stops the controller manager.
+func cleanupTestEnv(_ SpecContext) {
+	if controllerCancel != nil {
+		controllerCancel()
+	}
+
+	if gnmiServerCancel != nil {
```
felix-kaestner: I think this is a race, as  ggnmiServerCancel()` triggers a context cancelation, which means the the goroutine inside of `NewTestServer` will continue to call `GracefulStop()`, and `Close()` later also calls `GracefulStop()`. I believe this is also the reason why `sync.Once` was introduced as a quickfix, but I feel like that shows that the implementation is a bit muddy here.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,688 @@
...
+
+// Test environment — initialized in BeforeSuite
+var (
+	testEnv    *envtest.Environment
+	restConfig *rest.Config
+	k8sClient  client.Client
+	gnmiServer *testserver.Server
+)
+
+// Lifecycle — cancel functions called in AfterSuite
```
felix-kaestner: Why two separate contexts? Isn't this just one from `t.Context()`?

test/gnmi/server/server.go
```diff
@@ -294,9 +303,14 @@ func (s *State) Get(path *gpb.Path) []byte {
 	return []byte(res.Raw)
 }
 
-func (s *State) Set(path *gpb.Path, raw []byte) {
+func (s *State) Set(path *gpb.Path, raw []byte) error {
```
felix-kaestner:
```suggestion
func (s *State) Set(path *gpb.Path, raw []byte) (err error) {
```
nit: avoid having multiple `var err error` later.

test/gnmi/server/server.go
```diff
@@ -322,17 +336,38 @@ func (s *State) Set(path *gpb.Path, raw []byte) {
...
+		s.Buf = buf
+		return nil
+	}
+
+	for k, v := range elems[len(elems)-1].GetKey() {
+		buf, err = sjson.SetBytes(buf, sb.String()+"."+k, v)
+		if err != nil {
+			return fmt.Errorf("sjson.SetBytes final key %s: %w", k, err)
+		}
+	}
```
felix-kaestner:
```suggestion
	if elems := path.GetElem(); len(elems) > 0 {
	    for k, v := range elems[len(elems)-1].GetKey() {
		    buf, err = sjson.SetBytes(buf, sb.String()+"."+k, v)
		    if err != nil {
			    return fmt.Errorf("sjson.SetBytes final key %s: %w", k, err)
		    }
	    }
    }
```

test/gnmi/json_test.go
nikatza: We need to check unordered; in the link you mentioned they suggest `https://github.com/Benjamintf1/unmarshalledmatchers` for this case. It fits nicely, thanks!

.github/workflows/test-gnmi.yaml
nikatza: I planned to do it the follow-up, but I will incorporate it here then.

test/gnmi/gnmi_test.go
nikatza: Unfortunately this does not work with `envtest` as there is no garbage collection. Ref: https://book.kubebuilder.io/reference/envtest#testing-considerations

felix-kaestner: I didn't know. Thanks for the pointer!

test/gnmi/gnmi_test.go
nikatza: gingko's `t.Context()` is scoped to `BeforeSuite()` and the gNMI server would terminate before each test runs. So we need to use `context.Background()`. That said, we can just use one single context for it. Will update this.

test/gnmi/gnmi_test.go
nikatza: I think you are right. I have removed the goroutine from `NewTestServer()` and removed the 
`sync.Once`. The `Close()` no longer needs `sync.Once` since now there's only one shutdown path triggered by `AfterSuite()`

test/gnmi/gnmi_test.go
nikatza: Moved device creation to `BeforeEach`, removed the nil assignment, and updated the `cleanupAllResources` to delete any `device` resource.

test/gnmi/gnmi_test.go
nikatza: That's correct, if I had the registered the device reconciler. I have added it, removed this part regarding the `Phase`, and inlined the device creation into `BeforeEach()`

test/gnmi/gnmi_test.go
nikatza: Indeed. This is not necessary: we can simply check if the resource has finalizers during the cleanup in `AfterEach`.

test/gnmi/gnmi_test.go
nikatza: By doing so we have to add functionality to the tests to `preload` state into the device, i.e., the last reboot time, otherwise we can't transition.  This was added in a a follow up PR with the cisco quirks, but I will already introduce it here.

test/gnmi/gnmi_test.go
nikatza: I would argue that this is part of the gNMI integration tests and hence we don't care much about the controller flow, but rather we focus on the server state after the resource has been "ensured". For this reason wrapped the check in an `Eventually`: to wait until the resource is configured/ready and then check the payload in the server.

test/gnmi/gnmi_test.go
nikatza: Due to the changes applied I had to revert the location where the device is created. This is because now we also check the state of the servver after the k8s resources are deleted. So the device not anymore created inside the `BeforeEach` and the `AfterEach` blocks.

test/gnmi/gnmi_test.go
nikatza: As indicated earlier we can't rely on that with `envtest`

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,39 @@
...
+      matrix:
+        provider:
+          - openconfig
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+      - name: Install ginkgo
+        run: go install github.com/onsi/ginkgo/v2/ginkgo@latest
```
felix-kaestner:
```suggestion
```
The Makefile target for `test-gnmi` should already take care of installing ginkgo, if even required.

Makefile
```diff
@@ -99,9 +99,14 @@ test-e2e: setup-test-e2e manifests generate ## Run the e2e tests. Expected an is
...
 	@$(KIND) delete cluster --name $(KIND_CLUSTER)
 
+# Number of parallel Ginkgo processes
+GINKGO_PROCS ?= $(shell nproc 2>/dev/null || sysctl -n hw.logicalcpu)
+
 .PHONY: test-gnmi
-test-gnmi: FORCE ## Run integration tests for gNMI.
-	@printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
+test-gnmi: ## Run gNMI integration tests
+	@if ! hash setup-envtest 2>/dev/null; then printf "\e[1;36m>> Installing setup-envtest...\e[0m\n"; go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest; fi
```
felix-kaestner: envtest is already integrated and installed with the local bin setup.

```diff
diff --git a/Makefile b/Makefile
index 28ae0041..aae2346e 100644
--- a/Makefile
+++ b/Makefile
@@ -102,11 +102,12 @@ cleanup-test-e2e: ## Tear down the Kind cluster used for e2e tests
 # Number of parallel Ginkgo processes
 GINKGO_PROCS ?= $(shell nproc 2>/dev/null || sysctl -n hw.logicalcpu)
 
+# Provider for gNMI integration tests
+PROVIDER ?= openconfig
+
 .PHONY: test-gnmi
-test-gnmi: ## Run gNMI integration tests
-	@if ! hash setup-envtest 2>/dev/null; then printf "\e[1;36m>> Installing setup-envtest...\e[0m\n"; go install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest; fi
-	@printf "\e[1;36m>> ginkgo run --procs=$(GINKGO_PROCS) ./test/gnmi/ (PROVIDER=$(PROVIDER))\e[0m\n"
-	@KUBEBUILDER_ASSETS=$$(setup-envtest use 1.32 -p path) PROVIDER=$(PROVIDER) ginkgo run --tags=gnmi --procs=$(GINKGO_PROCS) ./test/gnmi/
+test-gnmi: setup-envtest ## Run gNMI integration tests
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" PROVIDER=$(PROVIDER) go test ./test/gnmi/ -v -ginkgo.v
 
 .PHONY: test-lab
 test-lab: ## Run lab tests against a real network device.
```
Also note that just running with `go test` works fine (as we do with the regular `test` target). If you need to use the ginkgo cli instead, it should be installed like all the other go tools at the bottom of the Makefile into a local bin/ directory.

On another note, I think you also want to exclude the gnmi test directory from the normal 'make test' target, as we do it with the e2e tests too.

```diff
diff --git a/Makefile b/Makefile
index 28ae0041..6b580d67 100644
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate setup-envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e | grep -v /lab) -coverprofile cover.out
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e | grep -v /lab | grep -v /gnmi/) -coverprofile cover.out
 
 .PHONY: coverage
 coverage: test ## Run tests and generate coverage report.
```

Note the trailing `/` at the gnmi exclusion to not exclude the `gnmiext` package.

test/gnmi/gnmi_suite_test.go
```diff
@@ -0,0 +1,358 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+//go:build gnmi
```
felix-kaestner: What is the specific reason to use a build tag here? E.g. the e2e tests don't need one, so I wonder what the difference is here.

test/gnmi/gnmi_suite_test.go
```diff
@@ -0,0 +1,358 @@
...
+		_ = testEnv.Stop() //nolint:errcheck // best-effort cleanup in AfterSuite
+	}
+})
+
+// registerControllers registers all controllers with the manager.
+// Add more controllers here as needed for testing.
+func registerControllers(ctx context.Context, mgr ctrl.Manager, recorder *events.FakeRecorder, providerFn provider.ProviderFunc, locker *resourcelock.ResourceLocker) {
+	var err error
+
+	err = (&core.PrefixSetReconciler{
```
felix-kaestner, line 160: I feel like this list might have gotten outdated, e.g. there is no EthernetSegmentReconciler registered here.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,312 @@
...
+				Expect(resources).NotTo(BeEmpty(), "Expected at least one resource in testdata")
+				Expect(stateDelete).NotTo(BeEmpty(), "Expected '-- state/delete --' section in testdata")
+
+				By("preloading gNMI state from testdata")
+				serverState := gnmiServer.State()
+				if len(statePre) != 0 {
+					serverState.Lock()
+					serverState.Buf = make([]byte, len(statePre))
+					copy(serverState.Buf, statePre)
+					serverState.Unlock()
```
felix-kaestner: I see that we copy multiple times to the serverState buffer. Should we perhaps refactor this into a dedicated `SetBuf` method?

test/gnmi/gnmi_test.go
felix-kaestner: The problem I see is that we check for the wrong condition and the eventually might return early therefore. The logic is that we only check for "Configured" when that is found on the object. However, we don't wait for the object to report any condition before we decide that.

So generally, if a resource has no conditions, it hasn't been reconciled yet. But if it contains any conditions, then it would also contain the "Configured" condition, as we then initialize that to "Unknown".

test/gnmi/gnmi_suite_test.go
nikatza: Removed. The tag was used when the e2e and gNMI tests were living together.

test/gnmi/gnmi_suite_test.go
nikatza: That is true, but there is also no test for that. Doesn't seem necessary to register a controller that is not used.

.github/workflows/test-gnmi.yaml
nikatza: Fixed. But I think we still need gingko to run the tests in parallel.  With ginkgo we don't have to worry about the gNMI server state being shared across the tests as  each process runs one server and the tests run sequentially there. We could refactor (again) the tests to have a gNMI server per test, or extend the `server` to support multiple `States`, but this seems too much for now.

Makefile
nikatza: Thanks for the suggestion. Updated accordingly.

test/gnmi/gnmi_test.go
nikatza: I see. Added an `g.Expect(conditions).NotTo(BeEmpty()`  before checking the condition.  The switch between `Ready/Configured` needs to stay until we merge the PR that adds the `Configured` condition to each resource.

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,37 @@
...
+name: Test gNMI integration
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '**.go'
+      - 'hack/**'
+      - 'config/**'
```
rgildein, line 13: You are missing test/gnmi here. What if I simply add new test file?

.github/workflows/test-gnmi.yaml
```diff
@@ -0,0 +1,37 @@
...
+    name: gNMI Integration (${{ matrix.provider }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
+      matrix:
+        provider:
+          - openconfig
+    steps:
+      - uses: actions/checkout@v7
+      - uses: actions/setup-go@v6
```
rgildein, line 32: nitpick, maybe we want to use later and greater version v7?

Makefile
```diff
@@ -99,10 +98,15 @@ test-e2e: setup-test-e2e manifests generate ## Run the e2e tests. Expected an is
 cleanup-test-e2e: ## Tear down the Kind cluster used for e2e tests
 	@$(KIND) delete cluster --name $(KIND_CLUSTER)
 
-.PHONY: test-gnmi
-test-gnmi: FORCE ## Run integration tests for gNMI.
-	@printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
+# Provider for gNMI integration tests
+PROVIDER ?= openconfig
 
+# Number of parallel Ginkgo processes
```
rgildein, line 104: is this just for test-gnmi? if yes can we add short not, e.g. "used by test-gnmi"

rgildein (commented):
LGTM, only blocker I see is the missing `test/gnmi` path.

test/gnmi/gnmi_suite_test.go
```diff
@@ -0,0 +1,356 @@
...
+	RegisterFailHandler(Fail)
+	_, _ = fmt.Fprintf(GinkgoWriter, "Starting network-operator gNMI integration tests\n")
+	RunSpecs(t, "gNMI Integration Suite")
+}
+
+// BeforeSuite initializes the test environment.
+// It starts the gNMI test server, sets up the Kubernetes client, and starts the controller manager.
+var _ = BeforeSuite(func(ctx SpecContext) {
+	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
+	SetDefaultEventuallyTimeout(30 * time.Second)
```
adamtrizuljak-sap, line 70: I would keep it at 60s to stay consistent with the unit tests. In my experience, even 60s was sometimes not enough.
```suggestion
	SetDefaultEventuallyTimeout(60 * time.Second)
```

test/gnmi/gnmi_suite_test.go
adamtrizuljak-sap: EthernetSegment is used in the unit tests and I had to fix a missing `RequeueInterval`  https://github.com/ironcore-dev/network-operator/pull/472/changes#diff-2bdc9c1ea98eea66e2b78315eebf14dc7141592ad2e71eaad8b2e721e12be5ae. So maybe it should also be added here.

nikatza: Yes, you are correct in that we have tests for them in the controller test suite. However, the scope of this PR is  to refactor the existing e2e tests such that we have a framework to test from resource definition until the gNMI payload. We are lacking many tests for this. I was planning a follow-up to increase our coverage and only focus on migrating the existing tests in this PR.

rgildein (approved):

Makefile
```diff
@@ -68,8 +68,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate setup-envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e | grep -v /lab) -coverprofile cover.out
-
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e | grep -v /lab | grep -v /gnmi/) -coverprofile cover.out
 .PHONY: coverage
```
felix-kaestner, line 73: I guess there is a missing empty line after this

Makefile
```diff
@@ -99,10 +98,15 @@ test-e2e: setup-test-e2e manifests generate ## Run the e2e tests. Expected an is
...
-	@printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
+# Provider used in test-gnmi
+PROVIDER ?= openconfig
 
+# Number of parallel Ginkgo processes
+GINKGO_PROCS ?= $(shell nproc 2>/dev/null || sysctl -n hw.logicalcpu)
+
+.PHONY: test-gnmi
+test-gnmi: setup-envtest ## Run gNMI tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" PROVIDER=$(PROVIDER) go test ./test/gnmi/ -v -ginkgo.v
```
felix-kaestner, line 110: Seems like there should be an empty line before next statement.

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,303 @@
...
+		if err := c.Get(ctx, key, r); err == nil && len(r.GetFinalizers()) > 0 {
+			hasFinalizer = true
+			break
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	if !hasFinalizer {
+		return
+	}
```
felix-kaestner, line 197: Should this be an Eventually block?

test/gnmi/gnmi_test.go
```diff
@@ -0,0 +1,303 @@
...
+			continue
+		}
+		if gvk.Group != v1alpha1.GroupVersion.Group && gvk.Group != nxv1alpha1.GroupVersion.Group {
+			continue
+		}
+
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(gvk.GroupVersion().WithKind(gvk.Kind + "List"))
+		if err := c.List(ctx, list, client.InNamespace(namespace)); err != nil {
+			continue
```
felix-kaestner, line 299: Should we Expect this to not error instead of continuing?

test/gnmi/gnmi_test.go
nikatza: An `Eventually` block result in a test error if it time outs and this is not exactly what we need here. The purpose of this section is to handle controller-managed resources as well as configuration-only resources. The assumption is that controller-managed resources have finalizers where as config-only resources don't.

Once we find a finalizer or we time-out without finding one, we: 
- If the resource is a config-only resource we exit the function (nothing to do, the resource was created).
- If the resource is a controller-managed resource we wait for the conditions to be set by the controller.

This is not the most elegant solution for the problem but works nicely in our case.

test/gnmi/gnmi_test.go
nikatza: When we call `AddToScheme` for a group-version we register several kubernetes meta types (DeleteOptions, CreateOptions, UpdateOptions, ListOptions, etc.) If we do something like:
```
		list := &unstructured.UnstructuredList{}
		list.SetGroupVersionKind(gvk.GroupVersion().WithKind(gvk.Kind + "List"))
		err := c.List(ctx, list, client.InNamespace(namespace))
		Expect(err).NotTo(HaveOccurred())
```

we run in a test-error caused by this other types:

```
• [FAILED] [2.169 seconds]
gNMI requests tests Provider: openconfig [It] It should handle the configuration in test file interface_physical_unnumbered
...github.com/ironcore-dev/network-operator/test/gnmi/gnmi_test.go:74

  [FAILED] Unexpected error:
      <*meta.NoKindMatchError | 0x69bdb3719b40>:
      no matches for kind "DeleteOptions" in version "networking.metal.ironcore.dev/v1alpha1"
      {
          GroupKind: {
              Group: "networking.metal.ironcore.dev",
              Kind: "DeleteOptions",
          },
          SearchedVersions: ["v1alpha1"],
      }
  occurred
  In [It] at:...github.com/ironcore-dev/network-operator/test/gnmi/gnmi_test.go:298 @ 08/19/26 19:18:36.663
 ```
 
 So `continue` allows us to ignore this errors and get a list of CRDs.

Comment (felix-kaestner) [reply test/gnmi/gnmi_test.go:299]: Could we at least check the error with `Errors.Is`/`Errors.As` for an `*meta.NoKindMatchError`? So we only continue on this precise error kind and have all other errors being asserted to not happen?

File: Makefile [hunk 45]
```diff
@@ -99,11 +99,17 @@ test-e2e: setup-test-e2e manifests generate ## Run the e2e tests. Expected an is
...
+GINKGO_PROCS ?= $(shell nproc 2>/dev/null || sysctl -n hw.logicalcpu)
+
 .PHONY: test-gnmi
-test-gnmi: FORCE ## Run integration tests for gNMI.
-	@printf "\e[1;33m>> gNMI integration tests not yet implemented\e[0m\n"
+test-gnmi: setup-envtest ## Run gNMI tests.
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" PROVIDER=$(PROVIDER) go test ./test/gnmi/ -v -ginkgo.v
 
-.PHONY: test-lab
+PHONY: test-lab
```
felix-kaestner:
```suggestion
.PHONY: test-lab
```
I think this was changed by mistake.

felix-kaestner (approved):

PR #411, opened 2026-06-18
Add kubectl-net plugin for managing network-operator resources

felix-kaestner:
Implements a kubectl plugin with get, pause, unpause, and shell
completion subcommands. The get command supports server-side table
printing and all standard kubectl output formats. Label shorthand
flags (--device/-d, --aggregate, --vrf, --routed-vlan, --evi)
provide convenient filtering. Resource definitions are code-generated
from the API types via kubebuilder markers.


felix-kaestner:
```bash
$ kind get clusters
network-operator
network-operator-target

$ kubectl --context kind-network-operator get deploy -n network-operator-system
NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
network-operator-controller-manager   1/1     1            1           112m

$ kubectl --context kind-network-operator-target get deploy -n network-operator-system
NAME                                  READY   UP-TO-DATE   AVAILABLE   AGE
network-operator-controller-manager   1/1     1            1           23m

$ kubectl net move --context kind-network-operator --to-context kind-network-operator-target --to-kubeconfig ~/.kube/config --dry-run
Checking target cluster CRDs...
Discovering resources in namespace "default"...
  Device/leaf1
  Interface/lo0

Total: 2 resources would be moved.

$ kubectl net move --context kind-network-operator --to-context kind-network-operator-target --to-kubeconfig ~/.kube/config
Checking target cluster CRDs...
Discovering resources in namespace "default"...
Pausing Devices on source...
Creating resources on target...
Deleting resources from source...
Unpausing Devices on target...
Move complete.
```

Makefile
SchwarzM, line 1: I do not see any target building the new kubectl-net plugin ?
Am I reading this correctly?

kubectl-net/cmd/kubectl-net.go
```diff
@@ -0,0 +1,24 @@
...
+	"github.com/ironcore-dev/kubectl-net/pkg/cmd"
+)
+
+func main() {
+	flags := pflag.NewFlagSet("kubectl-net", pflag.ExitOnError)
+	pflag.CommandLine = flags
+
+	root := cmd.NewCmdNet(genericiooptions.IOStreams{In: os.Stdin, Out: os.Stdout, ErrOut: os.Stderr})
+	if err := root.Execute(); err != nil {
+		os.Exit(1)
```
SchwarzM, line 24: this seems we just exit without bubbling the error to the user ?

kubectl-net/pkg/cmd/get.go
```diff
@@ -0,0 +1,207 @@
...
+		return err
+	}
+	o.Namespace = namespace
+	return nil
+}
+
+// Run fetches the requested resources and prints them to the configured output.
+func (o *GetOptions) Run(args []string) error {
+	f := ""
+	if o.PrintFlags.OutputFormat != nil {
```
SchwarzM: I think the output format is already initialized and cannot be nil.

kubectl-net/pkg/cmd/get.go
```diff
@@ -0,0 +1,207 @@
...
+	if len(infos) == 1 && infos[0].Object != nil {
+		return printer.PrintObj(infos[0].Object, o.Root.IOStreams.Out)
+	}
+
+	list := &unstructured.UnstructuredList{}
+	list.SetGroupVersionKind(schema.GroupVersionKind{Version: "v1", Kind: "List"})
+	list.SetResourceVersion("")
+	for _, info := range infos {
+		if info.Object != nil {
+			list.Items = append(list.Items, *info.Object.(*unstructured.Unstructured))
```
SchwarzM, line 171: you've made sure the conversion to unstructured is safe above why not here ?

kubectl-net/pkg/cmd/move.go
```diff
@@ -0,0 +1,363 @@
...
+	if err != nil {
+		return fmt.Errorf("target cluster: %w", err)
+	}
+
+	// Pre-flight: verify CRDs exist on target.
+	fmt.Fprintf(o.Root.IOStreams.Out, "Checking target cluster CRDs...\n")
+	crd := schema.GroupVersionResource{Group: "apiextensions.k8s.io", Version: "v1", Resource: "customresourcedefinitions"}
+	for _, def := range resourceDefs {
+		name := def.Name + "." + def.Group
+		_, err := tgtDyn.Resource(crd).Get(ctx, name, metav1.GetOptions{})
```
SchwarzM, line 107: maybe we should check the version ? It might not be in use now but could be in the future ...

kubectl-net/pkg/cmd/pause.go
```diff
@@ -0,0 +1,202 @@
...
+// Run patches the matched resources to set or remove the paused annotation.
+func (o *PauseOptions) Run(args []string) error {
+	result, err := buildResourceResult(o.Root.ConfigFlags, o.Namespace, o.Root.AllNamespaces, o.Resource.QualifiedName(), args, o.Labels.BuildSelector())
+	if err != nil {
+		return err
+	}
+
+	infos, err := result.Infos()
+	if err != nil {
+		return err
```
SchwarzM, line 124: we should at least print the error

kubectl-net/pkg/cmd/pause.go
```diff
@@ -0,0 +1,202 @@
...
+		resourceClient := dyn.Resource(info.Mapping.Resource)
+		var client dynamic.ResourceInterface = resourceClient
+		if info.Mapping.Scope.Name() == meta.RESTScopeNameNamespace && info.Namespace != "" {
+			client = resourceClient.Namespace(info.Namespace)
+		}
+		obj, err := client.Patch(context.Background(), info.Name, types.MergePatchType, patchBytes, metav1.PatchOptions{})
+		if err != nil {
+			return err
+		}
+		if obj.GetObjectKind().GroupVersionKind().Empty() {
```
SchwarzM: I am unsure if the GVK can be empty after a successful patch.

kubectl-net/pkg/cmd/resources.go
```diff
@@ -0,0 +1,78 @@
...
+	{Name: "accesscontrollists", Aliases: []string{"accesscontrollist", "acl"}, Kind: "AccessControlList", Group: "networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "banners", Aliases: []string{"banner"}, Kind: "Banner", Group: "networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "bgp", Aliases: []string{}, Kind: "BGP", Group: "networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "bgpconfigs", Aliases: []string{"bgpconfig", "nxbgp"}, Kind: "BGPConfig", Group: "nx.cisco.networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "bgppeers", Aliases: []string{"bgppeer", "peer", "bgpneighbor"}, Kind: "BGPPeer", Group: "networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "bordergateways", Aliases: []string{"bordergateway", "bgw"}, Kind: "BorderGateway", Group: "nx.cisco.networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "certificates", Aliases: []string{"certificate", "cert", "netcert"}, Kind: "Certificate", Group: "networking.metal.ironcore.dev", Version: "v1alpha1"},
+	{Name: "claims", Aliases: []string{"claim", "claim"}, Kind: "Claim", Group: "pool.networking.metal.ironcore.dev", Version: "v1alpha1"},
```
SchwarzM: the aliases look weird ... double "claim" ?

kubectl-net/pkg/cmd/resources.go
```diff
@@ -0,0 +1,78 @@
...
+// group and avoiding clashes with same-named types from other controllers
+// (e.g. Calico BGPPeer, cert-manager Certificate).
+func (r ResourceDef) QualifiedName() string {
+	if r.Group == "" {
+		return r.Name
+	}
+	return r.Name + "." + r.Group
+}
+
+func allResourceNames() []string {
```
SchwarzM: is this actually used somewhere ?

kubectl-net/pkg/cmd/move.go
```diff
@@ -0,0 +1,363 @@
...
+			fmt.Fprintf(o.Root.IOStreams.Out, "  %s/%s\n", obj.GetKind(), obj.GetName())
+		}
+		fmt.Fprintf(o.Root.IOStreams.Out, "\nTotal: %d resources would be moved.\n", len(objects))
+		return nil
+	}
+
+	// Pause devices on source.
+	fmt.Fprintf(o.Root.IOStreams.Out, "Pausing Devices on source...\n")
+	device := schema.GroupVersionResource{Group: "networking.metal.ironcore.dev", Version: "v1alpha1", Resource: "devices"}
+	patch, _ := json.Marshal(map[string]any{
```
SchwarzM: I would not ignore Marshaling errors and at least have some kind of must if you do not want to deal with the error explicitly

SchwarzM (changes requested):
Very nice ! I have some minor points.

Also we might want to consider adding a test or 2 for the non-trivial logic which is now very flat but might get more complicated over time. Thinking of patch generation and flags.

Makefile
felix-kaestner: Added a build target to the Makefile.

kubectl-net/cmd/kubectl-net.go
felix-kaestner: Cobra prints errors by default unless `SilenceErrors` is set. ref/ https://github.com/spf13/cobra/blob/v1.10.2/command.go#L1157-L1161

So based on this the error is already shown to the user.

kubectl-net/pkg/cmd/get.go
felix-kaestner: Changed.

kubectl-net/pkg/cmd/move.go
felix-kaestner: The `resourceDefs` we are looping over here are generated from code and therefore already contain the api version which we check on the target cluster. So this is already implicitly the case.

kubectl-net/pkg/cmd/get.go
felix-kaestner: Changed to not safely check in both places. It can't ever not be unstructured, as this is explicitly requested in the `buildResourceResult`.

kubectl-net/pkg/cmd/pause.go
felix-kaestner: Error will automatically be printed, as mentioned in https://github.com/ironcore-dev/network-operator/pull/411#discussion_r3843951774

kubectl-net/pkg/cmd/resources.go
felix-kaestner: This was related to the fact, that the `claim_types.go` used both `singular=claim` and also `shortName=claim`, where of course there is no need for the later. Fixed this.

kubectl-net/pkg/cmd/resources.go
felix-kaestner: No, removed.

kubectl-net/pkg/cmd/pause.go
felix-kaestner: Can't reproduce where I had this case. Therefore removed.

kubectl-net/pkg/cmd/move.go
felix-kaestner: Done.

SchwarzM (approved):
lgtm

PR #414, opened 2026-06-19
Extend BGPPeer and enable LocalAS Propagation

nikatza:
**Note**: this is a breaking change in the API.

The field `LocalASNumber` enables a router to appear as a member of another AS in addition to its real AS. Most vendors support mechanisms to specify how the device should use this number. This is, however, not specified in `OpenConfig`[1].

This commit removes `LocalASNumber` and adds `LocalAS`, which specifies the local AS number and how this number factors into eBGP announcements.
Specifically, we add three independent boolean flags:
- `PrependLocalAS`: prepend local-as to inbound routes (default: true)
- `PrependGlobalAS`: prepend global-as to outbound routes (default: true)

We opt to follow the Nokia YANG model [2], as it allows us to cover most of the vendors we checked, i.e., Cisco, Juniper, and Arista. This implies that the provider must ensure that the parameter combinations are valid, and reject them with an apistatus error that the controller can wrap into a TerminalError.

### Vendor mappings:

`Cisco NX-OS/IOS-XR` [3,4]:
- PrependLocalAS=false → no-prepend
- PrependLocalAS=false, PrependGlobalAS=false → no-prepend replace-as

`Juniper` [5]:
- PrependLocalAS=false → no-prepend-global-as
- PrependGlobalAS=false → private

`Arista` [6]:
- PrependLocalAS=false → no-prepend
- PrependLocalAS=false, PrependGlobalAS=false → no-prepend replace-as

`Nokia SR` [7]:
- PrependLocalAS → prepend-local-as
- PrependGlobalAS → prepend-global-as

Notice that Cisco, Juniper and Arista offer also the dual-as, alias, and fallback options which could be mapped to an AcceptEitherAS in our case. This is left here intentionally for future work.

This PR focuses on the NXOS implementation, leaving IOSXR for a follow up.
    
[1] https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-network-instance.html#network-instances-network-instance-protocols-protocol-bgp-neighbors-neighbor-config-local-as 
[2] https://yangbrowser.nokia.com/srlinux/24.10.7/tree?path=%2Fnetworkinstance%5Bname%3D*%5D%2Fprotocols%2Fbgp%2Fneighbor%5Bpeer-address%3D*%5D%2Flocal-as 
[3] https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/106x/configuration/unicast-routing-configuration/cisco-nexus-9000-series-nx-os-unicast-routing-configuration-guide/configuring-bgp.html
[4] https://www.cisco.com/c/en/us/td/docs/routers/asr9000/software/routing/configuration/guide/rcg-asr9000/rcg-bgp.html
[5] https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/local-as-edit-protocols-bgp.html
[6] https://www.arista.com/en/um-eos/eos-border-gateway-protocol-bgp 
[7] https://documentation.nokia.com/srlinux/24-10/books/routing-protocols/bgp.html

BREAKING CHANGE: LocalASNumber field removed, replaced with LocalAS
struct containing ASNumber, PrependLocalAS, PrependGlobalAS, and
AcceptEitherAS fields.


api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -61,9 +61,31 @@ type BGPPeerSpec struct {
...
+
+	// PrependGlobalAS specifies whether to prepend the global AS number to updates sent to this eBGP neighbor.
+	// +optional
+	// +kubebuilder:default=true
+	PrependGlobalAS *bool `json:"prependGlobalAS,omitempty"`
+
+	// AcceptEitherAS specifies whether to accept BGP connections using either the local-as or the global AS number.
+	// +optional
+	// +kubebuilder:default=false
+	AcceptEitherAS *bool `json:"acceptEitherAS,omitempty"`
```
felix-kaestner: Given that this option is currently not supported on Nokia SRLinux and that we currently don't have need for it internally, should we maybe keep this out of scope for now? Keeping the API Surface as minimal as possible (while fulfilling our current needs)?

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -61,9 +61,31 @@ type BGPPeerSpec struct {
...
 	AddressFamilies *BGPPeerAddressFamilies `json:"addressFamilies,omitempty"`
 
-	// LocalASNumber specifies a local AS number to present to the BGP peer, masking the global BGP process ASN.
+	// LocalAS configures the local AS number and how it factors into eBGP announcements for this peer.
 	// +optional
-	LocalASNumber *intstr.IntOrString `json:"localASNumber,omitempty"`
+	LocalAS *LocalAS `json:"localAS,omitempty"`
+}
+
+// LocalAS defines the local AS configuration and how it factors in eBGP announcements.
```
felix-kaestner: I think this also qualifies for iBGP on certain platforms, see https://www.cisco.com/c/en/us/td/docs/ios-xml/ios/iproute_bgp/configuration/xe-16/irg-xe-16-book/bgp-support-for-ibgp-local-as.html. 
> The neighbor local-as command can now be used to do the same on an iBGP session.

So maybe we shouldn't say that it is eBGP-only? Also applies to other comments on this type.

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -61,9 +61,31 @@ type BGPPeerSpec struct {
...
+type LocalAS struct {
+	// ASNumber specifies a local AS number to present in eBGP sessions with this peer.
+	// +required
+	ASNumber *intstr.IntOrString `json:"asNumber"`
```
felix-kaestner:
```suggestion
	ASNumber intstr.IntOrString `json:"asNumber"`
```
As this is required, I don't think it needs to be a pointer.

api/core/v1alpha1/bgp_peer_types.go
nikatza: Agree and removed.

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -61,9 +61,26 @@ type BGPPeerSpec struct {
 	// +optional
 	AddressFamilies *BGPPeerAddressFamilies `json:"addressFamilies,omitempty"`
 
-	// LocalASNumber specifies a local AS number to present to the BGP peer, masking the global BGP process ASN.
+	// LocalAS configures the local AS number and how it factors into eBGP announcements for this peer.
```
felix-kaestner: This still speaks of eBGP. https://github.com/ironcore-dev/network-operator/pull/414#discussion_r3450982928

felix-kaestner (approved):

api/core/v1alpha1/bgp_peer_types.go
nikatza: Fixed

felix-kaestner (approved):

PR #426, opened 2026-06-30
Add `Configured` condition to all controllers

adamtrizuljak-sap:
Previously, some controllers did not set the Configured condition after performing the configuration. They only set the Ready condition, which was ambiguous. This PR explicitly sets the Configured condition based on the provider success and ensures that the Ready condition is set correctly at the end of the reconcile loop.

- Initialize the `ConfiguredCondition` to `ConditionFalse` with `ReconcilePendingReason` to ensure consistency if the controller exits early due to an error
- Add deferred call of `conditions.RecomputeReady()` to ensure the Ready condition is evaluated and set at the end of the reconcile loop
- `cond := conditions.FromError(err)` already returns the Configured condition, so we just remove the next line that was overriding it with the Ready condition
- Update the associated tests to check that the Configured condition has been set
- Add Kubebuilder printcolumn for the Configured condition

# Todo
- [x] In BGP controller, EVPNInstance (and other) controllers check if it's correct to just replace the `ReadyCondition` to `ConfiguredCondition` - I suspect both conditions should be set
- [x] Fix tests
- [ ] Squash commits before merging

# Test reliability
While working on this PR I discovered issues with tests that cause typically 1-2 failures out of 10 runs of the test suite. See [my comment below](https://github.com/ironcore-dev/network-operator/pull/426#issuecomment-5068620382). I've spent lots of time trying to chase complex race conditions and managed to fix some of the failure modes, including:
- Adding waiting steps for resources to be truly created / deleted - helps avoid deadlocks when referenced resources were never getting deleted
- Adding missing `RequeueInterval` to `EthernetSegmentReconciler` because it was not getting re-queued for reconciliation 

I've separated the test fixes into a separate PR https://github.com/ironcore-dev/network-operator/pull/472 which has already been merged.

However the effort has reached diminishing returns while significantly delaying the completion of this PR. Therefore we've decided to merge this PR and continue to work on improving the tests in follow-up work.

### DIFF
--- a/internal/controller/cisco/nx/bordergateway_controller.go
+++ b/internal/controller/cisco/nx/bordergateway_controller.go
@@ -331,13 +331,6 @@ func (r *BorderGatewayReconciler) reconcile(ctx context.Context, s *borderGatewa
 		conditions.RecomputeReady(s.BorderGateway)
 	}()
 
-	conditions.Set(s.BorderGateway, metav1.Condition{
-		Type:    v1alpha1.ConfiguredCondition,
-		Status:  metav1.ConditionFalse,
-		Reason:  v1alpha1.ReconcilePendingReason,
-		Message: "Reconciliation is in progress",
-	})
-
 	// Ensure the BorderGateway is owned by the Device.
 	if !controllerutil.HasControllerReference(s.BorderGateway) {
 		if err := controllerutil.SetOwnerReference(s.Device, s.BorderGateway, r.Scheme, controllerutil.WithBlockOwnerDeletion(true)); err != nil {
--- a/internal/controller/cisco/nx/bordergateway_controller.go
+++ b/internal/controller/cisco/nx/bordergateway_controller.go
@@ -331,6 +331,13 @@ func (r *BorderGatewayReconciler) reconcile(ctx context.Context, s *borderGatewa
 		conditions.RecomputeReady(s.BorderGateway)
 	}()
 
+	conditions.Set(s.BorderGateway, metav1.Condition{
+		Type:    v1alpha1.ConfiguredCondition,
+		Status:  metav1.ConditionFalse,
+		Reason:  v1alpha1.ReconcilePendingReason,
+		Message: "Reconciliation is in progress",
+	})
+
 	// Ensure the BorderGateway is owned by the Device.
 	if !controllerutil.HasControllerReference(s.BorderGateway) {
 		if err := controllerutil.SetOwnerReference(s.Device, s.BorderGateway, r.Scheme, controllerutil.WithBlockOwnerDeletion(true)); err != nil {
--- a/internal/controller/cisco/nx/bordergateway_controller.go
+++ b/internal/controller/cisco/nx/bordergateway_controller.go
@@ -165,7 +165,7 @@ func (r *BorderGatewayReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 	}
 
 	orig := obj.DeepCopy()
-	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition) {
+	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition, v1alpha1.ConfiguredCondition) {
 		log.V(1).Info("Initializing status conditions")
 		return ctrl.Result{}, r.Status().Update(ctx, obj)
 	}
@@ -327,6 +327,10 @@ func (r *BorderGatewayReconciler) reconcile(ctx context.Context, s *borderGatewa
 
 	s.BorderGateway.Labels[v1alpha1.DeviceLabel] = s.Device.Name
 
+	defer func() {
+		conditions.RecomputeReady(s.BorderGateway)
+	}()
+
 	// Ensure the BorderGateway is owned by the Device.
 	if !controllerutil.HasControllerReference(s.BorderGateway) {
 		if err := controllerutil.SetOwnerReference(s.Device, s.BorderGateway, r.Scheme, controllerutil.WithBlockOwnerDeletion(true)); err != nil {
@@ -367,8 +371,6 @@ func (r *BorderGatewayReconciler) reconcile(ctx context.Context, s *borderGatewa
 	})
 
 	cond := conditions.FromError(err)
-	// As this resource is configuration only, we use the Configured condition as top-level Ready condition.
-	cond.Type = v1alpha1.ReadyCondition
 	conditions.Set(s.BorderGateway, cond)
 
 	return err
@@ -386,7 +388,7 @@ func (r *BorderGatewayReconciler) reconcileSourceInterface(ctx context.Context, 
 	if err := r.Get(ctx, key, intf); err != nil {
 		if apierrors.IsNotFound(err) {
 			conditions.Set(s.BorderGateway, metav1.Condition{
-				Type:    v1alpha1.ReadyCondition,
+				Type:    v1alpha1.ConfiguredCondition,
 				Status:  metav1.ConditionFalse,
 				Reason:  v1alpha1.InterfaceNotFoundReason,
 				Message: fmt.Sprintf("source interface %q not found", key),
@@ -398,7 +400,7 @@ func (r *BorderGatewayReconciler) reconcileSourceInterface(ctx context.Context, 
 
 	if intf.Spec.DeviceRef.Name != s.Device.Name {
 		conditions.Set(s.BorderGateway, metav1.Condition{
-			Type:    v1alpha1.ReadyCondition,
+			Type:    v1alpha1.ConfiguredCondition,
 			Status:  metav1.ConditionFalse,
 			Reason:  v1alpha1.CrossDeviceReferenceReason,
 			Message: fmt.Sprintf("source interface %q does not belong to device %q", intf.Name, s.Device.Name),
@@ -408,7 +410,7 @@ func (r *BorderGatewayReconciler) reconcileSourceInterface(ctx context.Context, 
 
 	if intf.Spec.Type != v1alpha1.InterfaceTypeLoopback {
 		conditions.Set(s.BorderGateway, metav1.Condition{
-			Type:    v1alpha1.ReadyCondition,
+			Type:    v1alpha1.ConfiguredCondition,
 			Status:  metav1.ConditionFalse,
 			Reason:  v1alpha1.InvalidInterfaceTypeReason,
 			Message: fmt.Sprintf("source interface %q is not of type Loopback, got %q", intf.Name, intf.Spec.Type),
@@ -427,7 +429,7 @@ func (r *BorderGatewayReconciler) reconcileInterconnectInterfaces(ctx context.Co
 		if err := r.Get(ctx, client.ObjectKey{Name: ref.Name, Namespace: s.BorderGateway.Namespace}, intf); err != nil {
 			if apierrors.IsNotFound(err) {
 				conditions.Set(s.BorderGateway, metav1.Condition{
-					Type:    v1alpha1.ReadyCondition,
+					Type:    v1alpha1.ConfiguredCondition,
 					Status:  metav1.ConditionFalse,
 					Reason:  v1alpha1.InterfaceNotFoundReason,
 					Message: fmt.Sprintf("interconnect interface %q not found", ref.Name),
@@ -439,7 +441,7 @@ func (r *BorderGatewayReconciler) reconcileInterconnectInterfaces(ctx context.Co
 
 		if intf.Spec.DeviceRef.Name != s.Device.Name {
 			conditions.Set(s.BorderGateway, metav1.Condition{
-				Type:    v1alpha1.ReadyCondition,
+				Type:    v1alpha1.ConfiguredCondition,
 				Status:  metav1.ConditionFalse,
 				Reason:  v1alpha1.CrossDeviceReferenceReason,
 				Message: fmt.Sprintf("interconnect interface %q does not belong to device %q", intf.Name, s.Device.Name),
@@ -449,7 +451,7 @@ func (r *BorderGatewayReconciler) reconcileInterconnectInterfaces(ctx context.Co
 
 		if intf.Spec.Type != v1alpha1.InterfaceTypePhysical {
 			conditions.Set(s.BorderGateway, metav1.Condition{
-				Type:    v1alpha1.ReadyCondition,
+				Type:    v1alpha1.ConfiguredCondition,
 				Status:  metav1.ConditionFalse,
 				Reason:  v1alpha1.InvalidInterfaceTypeReason,
 				Message: fmt.Sprintf("interconnect interface %q is not of type Physical, got %q", intf.Name, intf.Spec.Type),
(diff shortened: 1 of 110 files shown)


internal/controller/core/bgp_controller.go
```diff
@@ -443,7 +445,7 @@ func (r *BGPReconciler) reconcileVRF(ctx context.Context, bgp *v1alpha1.BGP, dev
 	}, vrf); err != nil {
 		if apierrors.IsNotFound(err) {
 			conditions.Set(bgp, metav1.Condition{
-				Type:    v1alpha1.ReadyCondition,
+				Type:    v1alpha1.ConfiguredCondition,
```
adamtrizuljak-sap, line 455: @nikatza I'm not sure if this is entirely correct and I'd like to discuss it with you. It seems like we should set both conditions here, not just replace the Ready condition with Configured condition. Same goes for the other files where this was done.

felix-kaestner: I think this is correct with only setting the "Configured" Condition. The `reconcileVRF` func is called by `reconcile`. If the former sets the "Configured" condition, the later still calls `RecomputeReady` in it's `defer` call when returning. So the Ready condition would be set according on the "Configured" condition, which is what we want.

adamtrizuljak-sap: Yes that seems to be right. We set the Configured condition explicitly here and setting the Ready condition is taken care off by the deferred call to `conditions.RecomputeReady()`.

felix-kaestner (commented):
As we are adding a new condition to these resources, should we also add a printcolumn like so?

```diff
diff --git a/api/core/v1alpha1/acl_types.go b/api/core/v1alpha1/acl_types.go
index 6501f362..a7a01b70 100644
--- a/api/core/v1alpha1/acl_types.go
+++ b/api/core/v1alpha1/acl_types.go
@@ -122,6 +122,7 @@ type AccessControlListStatus struct {
 // +kubebuilder:printcolumn:name="Device",type=string,JSONPath=`.spec.deviceRef.name`
 // +kubebuilder:printcolumn:name="Entries",type=string,JSONPath=`.status.entriesSummary`,priority=1
 // +kubebuilder:printcolumn:name="Ready",type=string,JSONPath=`.status.conditions[?(@.type=="Ready")].status`
+// +kubebuilder:printcolumn:name="Configured",type=string,JSONPath=`.status.conditions[?(@.type=="Configured")].status`,priority=1
 // +kubebuilder:printcolumn:name="Paused",type=string,JSONPath=`.status.conditions[?(@.type=="Paused")].status`,priority=1
 // +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
```

internal/controller/core/dhcprelay_controller.go
```diff
@@ -178,7 +178,7 @@ func (r *DHCPRelayReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	orig := obj.DeepCopy()
-	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition) {
+	if conditions.InitializeConditions(obj, v1alpha1.ReadyCondition, v1alpha1.ConfiguredCondition) {
```
nikatza, line 181: Do you think it would make sense to include/refactor the `dhcprelay_controller_test.go ` and align with the other tests style?

nikatza:
@adamtrizuljak-sap We merged the `aaa` resource in the mean time, which unfortunately has the same issue that you fix in this PR. Could you please update your PR and include a fix for it? 

Likewise, in this PR we only tackle `core` resources. There are some other resources like `bordergateway` and `system` in `cisco/nx` that could use this fix too. 

Thanks a lot for your contribution! 🔥

adamtrizuljak-sap:
> @adamtrizuljak-sap We merged the `aaa` resource in the mean time, which unfortunately has the same issue that you fix in this PR. Could you please update your PR and include a fix for it?
> 
> Likewise, in this PR we only tackle `core` resources. There are some other resources like `bordergateway` and `system` in `cisco/nx` that could use this fix too.
> 
> Thanks a lot for your contribution! 🔥

Done. Are there any other resources that need this change?

felix-kaestner (commented):
We currently have some places in the codebase that use the check for `IsReady` on resources, that previously only had the `Ready` condition, e.g. https://github.com/ironcore-dev/network-operator/blob/046350c49f301394940a955cdff3c194e53a3061/internal/controller/core/bgp_controller.go#L465-L466

This is combined with a watch trigger to do reconcilation once this condition changes, e.g. https://github.com/ironcore-dev/network-operator/blob/046350c49f301394940a955cdff3c194e53a3061/internal/controller/core/bgp_controller.go#L311

Now that resources like the `VRF` get a configured condition, I think those checks should also be adjusted to use the `IsConfigured` check instead of the `IsReady` check, https://github.com/ironcore-dev/network-operator/blob/046350c49f301394940a955cdff3c194e53a3061/internal/conditions/conditions.go#L85-L96

So I think one additional task would be to go through the codebase and find usages of `IsReady` which were used for resources that only had the `Ready` condition previously and change those to now check for `IsConfigured` instead.

adamtrizuljak-sap:
> We currently have some places in the codebase that use the check for `IsReady` on resources, that previously only had the `Ready` condition, e.g.
> 
> https://github.com/ironcore-dev/network-operator/blob/046350c49f301394940a955cdff3c194e53a3061/internal/controller/core/bgp_controller.go#L465-L466
> ...

@felix-kaestner I _think_ that this should not be an issue or a change in behavior? At the end of reconciliation, `RecomputeReady()` loops over all conditions of the resource and if any of the conditions is False, it sets the `Ready` condition to False

https://github.com/ironcore-dev/network-operator/blob/main/internal/conditions/conditions.go#L119

Therefore if the `Configured` condition is False, the `Ready` condition will also be False and the following check should still have the same behavior. Plus it will trigger the reconciliation if any other condition becomes False, not just the Configured condition

> This is combined with a watch trigger to do reconcilation once this condition changes, e.g.
> 
> https://github.com/ironcore-dev/network-operator/blob/046350c49f301394940a955cdff3c194e53a3061/internal/controller/core/bgp_controller.go#L311
> 

Do you think that we should also explicitly evaluate the `Configured` condition?

felix-kaestner:
> Do you think that we should also explicitly evaluate the `Configured` condition?

Yes, I think that would be more semantically correct here, as the intention is "I can't configure X before configuring Y". The reason this currently uses the "Ready" condition was simply that these resources didn't have a "Configured" condition, as is also commented in those places. Now that we introduce the "Configured" condition I think we should change that.

adamtrizuljak-sap:
> > Do you think that we should also explicitly evaluate the `Configured` condition?
> 
> Yes, I think that would be more semantically correct here, as the intention is "I can't configure X before configuring Y". The reason this currently uses the "Ready" condition was simply that these resources didn't have a "Configured" condition, as is also commented in those places. Now that we introduce the "Configured" condition I think we should change that.

Got it. @nikatza helped me understand the problem - you were talking about only about the controllers that depend on other referenced resources.

For example the `BGPPeer` controller watches referenced `BGP` and `VRF` resources and triggers a self-reconciliation when their conditions change. Since these resources now expose a proper Configured condition, we update the checking logic to use it instead of the Ready condition (which now has too broad meaning).

adamtrizuljak-sap:
Update after long time, as I've been (and still am) trying to fix a regression in test reliability...

The `Configured` condition is being updated at the end of `reconcile()`. If an error happens earlier in `reconcile()`, this part is never reached and the condition stays outdated. This caused tests to fail randomly, where one error in one resource's controller would cause a cascade of subsequent test case failures due to conditions not being updated. Especially tests and other logic that uses [`conditions.IsConfigured()`](https://github.com/ironcore-dev/network-operator/blob/eb056e0673565422e964a65f40cc2e4245601b01/internal/conditions/conditions.go#L87), which checks if `condition.ObservedGeneration` matches the object's generation. Resources that reference other resources were more likely to fail, because the probabilities of failure in the referenced resources add up. For example `BGPPeer` which references `VRF`, `BGP` and `RoutingPolicy`. Overall failure rate was approx. 3-4 fails out of 10 runs of the full test suite.

My idea to fix this is to explicitly set the `Configured` condition to False or Unknown early in the `reconcile()` function. Therefore if it exits early due to an error, the condition will be left in a consistent state and its `ObservedGeneration` will be up-to-date. Only if all actions in `reconcile()` succeed, we get to the end and actually set the condition based on the result of `Provider.EnsureXXX()`
```go
conditions.Set(s.BGPPeer, metav1.Condition{
	Type:   v1alpha1.ConfiguredCondition,
	Status: metav1.ConditionFalse,
})

// Controller code which may return early due to error...

err = s.Provider.EnsureBGPPeer(ctx, &provider.EnsureBGPPeerRequest{...})
cond := conditions.FromError(err)
conditions.Set(s.BGPPeer, cond)
```

This method has improved the test suite reliability to less than 1 failure out of 10 runs, but it's still not quite there. I will likely have to add this fix to all controllers that have the Configured condition.

internal/controller/cisco/nx/bordergateway_controller.go
```diff
@@ -327,6 +327,17 @@ func (r *BorderGatewayReconciler) reconcile(ctx context.Context, s *borderGatewa
...
+	defer func() {
+		conditions.RecomputeReady(s.BorderGateway)
+	}()
+
+	conditions.Set(s.BorderGateway, metav1.Condition{
+		Type:    v1alpha1.ConfiguredCondition,
+		Status:  metav1.ConditionFalse,
+		Reason:  v1alpha1.ReconcilePendingReason,
+		Message: "Reconciliation is in progress",
+	})
```
felix-kaestner, line 339: General Question: Did we agree that we want to reset the status condition like this at the start of the reconcilation?

adamtrizuljak-sap: I think we discussed this on our sync last week and we agreed on this, but correct me if I'm wrong. We can discuss it again today if needed.

IvoGoman: Have we validated this outside of the tests? I would expect the resetting on the beginning of the reconciliation to cause a reconciliation loop since the objects status is changed during each iteration. As soon as the status is flipped the condition will get a new `LastTransitionedAt` timestamp, so even for an otherwise noop each reconciliation will end up requeueing the object.

felix-kaestner: This is a very valid concern, if we always set the condition at the start of the reconcilation we alter the status even if we go from `Success → Success`, which cause an infinite loop of reconcilation (as also status updates trigger another reconcilation) and it also breaks the `LastTransitionedAt` value.

adamtrizuljak-sap:
@IvoGoman correctly noted that overwriting the conditions in every loop updates the `LastTransitionedAt` timestamp, causing infinite reconciliation. Therefore this fix is not valid. I will have to remove the this code section and see if the controllers still work and the tests still pass.

The proper way will be to explicitly handle all `reconcile()` failure cases in all controllers with the Configured condition. For example, the [`DHCPRelay`](https://github.com/ironcore-dev/network-operator/blob/main/internal/controller/core/dhcprelay_controller.go#L233) controller already handles the error cases and sets the Configured condition appropriately. I need to check if all controllers do this and implement it if not.

internal/controller/cisco/nx/bordergateway_controller.go
adamtrizuljak-sap: I removed this as we discussed. Some failing tests popped up again, which I tried to fix.

internal/controller/core/bgp_peer_controller.go
```diff
@@ -426,7 +453,7 @@ func (r *BGPPeerReconciler) reconcile(ctx context.Context, s *bgpPeerScope) (ret
 					Reason:  v1alpha1.InterfaceNotFoundReason,
 					Message: fmt.Sprintf("source interface %q not found", addr.InterfaceRef.Name),
 				})
-				return reconcile.TerminalError(fmt.Errorf("source interface %q not found", addr.InterfaceRef.Name))
+				return fmt.Errorf("source interface %q not found", addr.InterfaceRef.Name)
```
felix-kaestner, line 456: This should stay a `reconcile.TerminalError`. The watcher will make sure to retrigger reconcilation when the referenced resource is created. Returning a non-terminal error will otherwise retry over and over again in which there is no point.

internal/controller/core/bgp_peer_controller_test.go
```diff
@@ -297,6 +303,35 @@ var _ = Describe("BGPPeer Controller", func() {
...
+			}
+			Expect(k8sClient.Create(ctx, differentDevice)).To(Succeed())
+			DeferCleanup(func() {
+				intfList := &v1alpha1.InterfaceList{}
+				Expect(k8sClient.List(ctx, intfList, client.InNamespace(metav1.NamespaceDefault), client.MatchingLabels{v1alpha1.DeviceLabel: differentDevice.Name})).To(Succeed())
+				for i := range intfList.Items {
+					Expect(client.IgnoreNotFound(k8sClient.Delete(ctx, &intfList.Items[i]))).To(Succeed())
+				}
+				Expect(client.IgnoreNotFound(k8sClient.Delete(ctx, differentDevice))).To(Succeed())
+			})
```
felix-kaestner: Why are we deleting interfaces here for a freshly created device? (This shouldn't be neeed, the device uses a GenerateName, so it's most likely unique anyways and doesn't have interface resources linked against it)

internal/controller/core/bgp_peer_controller_test.go
```diff
@@ -462,7 +497,7 @@ var _ = Describe("BGPPeer Controller", func() {
 		})
 
 		It("Should not reconcile iBGP peer if local-as is set", func() {
-			By("Creating a BGP resource for the Device")
+			// By("Creating a BGP resource for the Device")
```
felix-kaestner: Leftover commented out code?

felix-kaestner (changes requested):

internal/controller/core/bgp_peer_controller_test.go
adamtrizuljak-sap: The `By()` statement is duplicated in main, I'll delete it

internal/controller/core/bgp_peer_controller.go
adamtrizuljak-sap: I found that after TerminalError the controller would never be re-enqueued for reconciliation. Maybe this is a deeper issue with envtest, as similar problems are happening all over the test suite.

IvoGoman:
In the case that the device gets paused, the `Configured` condition might not reflect the current `.spec` of a resource but an older version.
If the `ObserveredGeneration` of the current object is less than the value in the condition then `Configured` can be set to false as well. This shows that the current `.spec` was not yet reconciled onto the device.

internal/controller/core/bgp_peer_controller_test.go
adamtrizuljak-sap: You are right. And envtest should even create a unique namespace for every test, so cleanup shouldn't really be required here. It was added by Claude while trying to fix the tests. I've removed it and it doesn't seem to have an impact on test failures.

adamtrizuljak-sap:
> In the case that the device gets paused, the `Configured` condition might not reflect the current `.spec` of a resource but an older version. If the `ObserveredGeneration` of the current object is less than the value in the condition then `Configured` can be set to false as well. This shows that the current `.spec` was not yet reconciled onto the device.

Reconciliation doesn't happen when the device is paused. So I think that it's semantically correct that the current condition might not be reflecting the current state, it will get updated once the device is resumed.

Are you suggesting to always set Configured to False (or Unknown) when the device is paused?  https://github.com/ironcore-dev/network-operator/blob/main/internal/paused/paused.go#L48

PR #435, opened 2026-07-07
feat: Static Route Controller

sven-rosenzweig:
Implement IOS-XR and NXOS provider.


api/core/v1alpha1/groupversion_info.go
```diff
@@ -72,6 +72,10 @@ const L2VNILabel = "networking.metal.ironcore.dev/evi-name"
 // the name of the VRF they belong to.
 const VRFLabel = "networking.metal.ironcore.dev/vrf-name"
 
+// StaticRouteLabel is a label applied to VRFs to indicate
+// the name of the StaticRoute that references them.
+const StaticRouteLabel = "networking.metal.ironcore.dev/static-route-name"
```
felix-kaestner: How will this work when there are multiple `StaticRoute`s in a VRF? Based on the way we structure the `StaticRoute` CRD this would be possible, no?

api/core/v1alpha1/staticroute_types.go
```diff
@@ -0,0 +1,145 @@
...
+	ProviderConfigRef *TypedLocalObjectReference `json:"providerConfigRef,omitempty"`
+
+	// Name is the name of the static route.
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=255
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Name is immutable"
+	Name string `json:"name"`
+
+	// +optional
```
felix-kaestner, line 40: Should have a very brief comment, as all other fields.

api/core/v1alpha1/staticroute_types.go
```diff
@@ -0,0 +1,145 @@
...
+	Address string `json:"address,omitempty"`
+
+	// Metric assigns a priority to the static route. Lower values indicate higher priority.
+	// +optional
+	Metric *int32 `json:"metric,omitempty"`
+}
+
+// StaticRouteStatus defines the observed state of StaticRoute.
+type StaticRouteStatus struct {
+	// The conditions are a list of status objects that describe the state of the Interface.
```
felix-kaestner:
```suggestion
	// The conditions are a list of status objects that describe the state of the StaticRoute.
```
copy+pasta

api/core/v1alpha1/staticroute_types.go
```diff
@@ -0,0 +1,145 @@
...
+
+	// VrfRef is a reference to the VRF resource that this static route belongs to.
+	// If not specified, the static route will be part of the default VRF.
+	// The referenced VRF must exist in the same namespace.
+	// +optional
+	VrfRef *LocalObjectReference `json:"vrfRef,omitempty"`
+
+	// IPPrefix is the destination IP prefix for the static route.
+	// +required
+	Prefix IPPrefix `json:"prefix,omitempty"`
```
felix-kaestner:
```suggestion
	Prefix IPPrefix `json:"prefix"`
```
omitempty is unused if the field is required. Also applies to the NextHops.

config/rbac/staticroute_admin_role.yaml
```diff
@@ -0,0 +1,27 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app.kubernetes.io/name: network-operator
+    app.kubernetes.io/managed-by: kustomize
+  name: networking-staticroute-admin-role
+rules:
+- apiGroups:
+  - networking.networking.metal.ironcore.dev
```
felix-kaestner:
```suggestion
  - networking.metal.ironcore.dev
```
This is wrong. Probably a `grep` over the codebase reveals other places that are also effected.

config/samples/v1alpha1_staticroute.yaml
felix-kaestner, line 1: Can you add this sample also to the `Tiltfile`?

config/samples/v1alpha1_staticroute.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  labels:
+  name: staticroute-sample
+spec:
+  name: sample-static-route
+  description: This is a sample static route.
+  deviceRef:
+    name: sample-device
+  IPPrefix: 
+    Prefix: "192.168.1.0/24"
```
felix-kaestner: I think this is wrong.

internal/controller/core/interface_controller.go
```diff
@@ -90,7 +90,7 @@ func (r *InterfaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		if apierrors.IsNotFound(err) {
 			// If the custom resource is not found then it usually means that it was deleted or not created
 			// In this way, we will stop the reconciliation
-			log.V(3).Info("Resource not found. Ignoring since object must be deleted")
+			log.V(3).Info("Resource not foufinnd. Ignoring since object must be deleted")
```
felix-kaestner: The hell?

api/core/v1alpha1/staticroute_types.go
```diff
@@ -0,0 +1,145 @@
+// SPDX-FileCopyrightText: 2025 SAP SE or an SAP affiliate company and IronCore contributors
```
felix-kaestner: Wrong Year. Also applies to other newly created files.

internal/controller/core/suite_test.go
```diff
@@ -998,6 +1010,21 @@ func (p *Provider) GetEthernetSegment(name string) (string, bool) {
 	return esi, ok
 }
 
+func (p *Provider) EnsureStaticRoute(_ context.Context, req *provider.StaticRouteRequest) error {
+	p.Lock()
+	defer p.Unlock()
+
```
felix-kaestner:
```suggestion
```

internal/controller/core/vrf_controller.go
```diff
@@ -313,6 +313,11 @@ func (r *VRFReconciler) finalize(ctx context.Context, s *vrfScope) (reterr error
 		}
 	}()
 
+	// Check if the VRF is still referenced by any StaticRoute.
+	if _, found := s.VRF.Labels[v1alpha1.StaticRouteLabel]; found {
+		return fmt.Errorf("cannot delete VRF %q because it is still referenced by StaticRoute %q", s.VRF.Name, s.VRF.Labels[v1alpha1.StaticRouteLabel])
```
felix-kaestner: I don't think we should do this. I think the normal approach here would be to just continue with the deletion and remove the whole vrf from the device. Then, when the `StaticRoute` reconciles for the next time, it tries to load the referenced VRF and goes into a terminal error state that the "vrf is not found". 
But a StaticRoute CR blocking deletion of the VRF which doesn't reference the StaticRoute (only the other way around) doesn't feel right.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -574,6 +573,65 @@ func (p *Provider) GetPeerStatus(ctx context.Context, req *provider.BGPPeerStatu
...
+			IsIpv4:         false,
+		}
+	}
+
+	if req.VRF != nil && req.VRF.Spec.Name != "" {
+		vrf := req.VRF.Spec.Name
+		prefix.VRFName = vrf
+		return p.client.Update(ctx, &prefix)
+	}
+	return p.client.Update(ctx, &prefix)
```
felix-kaestner, line 647:
```suggestion
	if req.VRF != nil && req.VRF.Spec.Name != "" {
		prefix.VRFName = req.VRF.Spec.Name
	}
	
	return p.client.Update(ctx, &prefix)
```
Doesn't need to make it so complicated.

internal/provider/cisco/iosxr/provider.go
```diff
@@ -574,6 +573,65 @@ func (p *Provider) GetPeerStatus(ctx context.Context, req *provider.BGPPeerStatu
...
+			IsIpv4:         true,
+		}
+	} else if prefixIP.Addr().Is6() {
+		prefix = Prefix{
+			PrefixAddress:  prefixIP.Addr().String(),
+			PrefixLength:   prefixIP.Bits(),
+			NextHopAddress: &nexthopAddress,
+			IsIpv4:         false,
+		}
+	}
```
felix-kaestner:
```suggestion
	prefix = Prefix{
		PrefixAddress:  prefixIP.Addr().String(),
		PrefixLength:   prefixIP.Bits(),
		NextHopAddress: &nexthopAddress,
		IsIpv4:         prefixIP.Addr().Is4(),
	}
```

internal/provider/cisco/iosxr/static_route.go
```diff
@@ -0,0 +1,64 @@
...
+
+package iosxr
+
+import "strconv"
+
+func (s *Prefix) XPath() string {
+	basePath := "Cisco-IOS-XR-um-router-static-cfg:router/static/vrfs/vrf[vrf-name=" + s.VRFName + "]/address-family/"
+	if s.VRFName == "" {
+		basePath = "Cisco-IOS-XR-um-router-static-cfg:router/static/address-family/"
+	}
```
felix-kaestner:
```suggestion
	basePath := "Cisco-IOS-XR-um-router-static-cfg:router/static/address-family/"
	if s.VRFName != "" {
		basePath = "Cisco-IOS-XR-um-router-static-cfg:router/static/vrfs/vrf[vrf-name=" + s.VRFName + "]/address-family/"
	}
```
nit: the other way around feels more logical to me.

internal/provider/cisco/nxos/provider.go
```diff
@@ -3804,6 +3805,16 @@ func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest
 	return p.Update(ctx, &Feature{Name: "tacacsplus", AdminSt: AdminStDisabled})
 }
 
+func (p *Provider) EnsureStaticRoute(ctx context.Context, req *provider.StaticRouteRequest) error {
+	// TODO(sven-rosenzweig): implement static route configuration
```
felix-kaestner: Is there an open Ticket for this? Otherwise I think leaving these methods out would be better, as the controller already checks if the provider implements this interface (which it now does, because it has these methods). Not having these functions means the controller sees that the provider doesn't implement the contract and correctly updates the kubernetes resource status.

PROJECT
```diff
@@ -290,6 +290,14 @@ resources:
   kind: DHCPRelay
   path: github.com/ironcore-dev/network-operator/api/core/v1alpha1
   version: v1alpha1
+- api:
+    crdVersion: v1
+    namespaced: true
+  domain: networking.metal.ironcore.dev
+  group: core
```
felix-kaestner:
```suggestion
```
not correct.

felix-kaestner (changes requested):
Nice stuff! 🚀

internal/controller/core/staticroute_controller.go
```diff
@@ -0,0 +1,534 @@
...
+			handler.EnqueueRequestsFromMapFunc(r.staticRoutesForProviderConfig),
+			builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),
+		)
+	}
+
+	return bldr.
+		Watches(
+			&v1alpha1.VRF{},
+			handler.EnqueueRequestsFromMapFunc(r.vrfToStaticRoute),
+			builder.WithPredicates(predicate.Funcs{
```
felix-kaestner, line 256: I don't think we want to trigger reconcilation on updates to the vrf, as we are basically only interested in the `.spec.name` and that is immutable. So I would expect that we have a 

```go
				return false
			},
```

internal/controller/core/staticroute_controller_test.go
```diff
@@ -0,0 +1,368 @@
...
+				g.Expect(k8sClient.Get(ctx, staticRouteKey2, resource)).To(Succeed())
+
+				ready := false
+				for _, cond := range resource.Status.Conditions {
+					if cond.Type == v1alpha1.ReadyCondition {
+						ready = true
+						g.Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+					}
+				}
+				g.Expect(ready).To(BeTrue(), "ReadyCondition should be present")
```
felix-kaestner: I would prefer to adhere to the same pattern for condition checks as in the other controller tests (e.g. to use the `metav1.FindStatusCondition`).

api/core/v1alpha1/groupversion_info.go
sven-rosenzweig: Yes, this does not work as a VRF will have multiple static routes routes (tuple of vrf, prefix and nexthop).

sven-rosenzweig: I guess we do not need this anymore, as I tried to model a 1-1 relationship, which did change compared to the first draft of this controller.

internal/controller/core/vrf_controller.go
sven-rosenzweig: Agree with you here. Let me remove it.

internal/provider/cisco/nxos/provider.go
sven-rosenzweig: https://github.wdf.sap.corp/sap-cloud-infrastructure/neutron-issues/issues/463

PR #436, opened 2026-07-08
Add VRFRef to EVPNInstance and refactor L3VNI ownership

felix-kaestner:
Move L3VNI configuration from the VRF resource to the EVPNInstance (Routed type). This aligns with how NX-OS models the relationship: the EVI owns the VNI-to-VRF binding, not the VRF itself.

  - Deprecate spec.vni on VRF in favor of the EVPNInstance VNI field
  - Add spec.vrfRef to EVPNInstance for Routed type (mirrors vlanRef for Bridged)
  - Refactor NX-OS provider: VRF uses Patch (preserves EVI-managed fields), EVI patches L3VNI/Encap separately
  - Restrict routeDistinguisher on EVPNInstance to Bridged type only
  - Support "Auto" value for route distinguisher (rd auto)
  - Reuse RouteDistinguisher/RouteTarget helpers across VRF and EVI providers


internal/controller/core/evpninstance_controller.go
```diff
@@ -266,6 +279,20 @@ func (r *EVPNInstanceReconciler) SetupWithManager(ctx context.Context, mgr ctrl.
 				},
 			}),
 		).
+		// Watches enqueues EVPNInstances for updates in referenced VRF resources.
+		// Only triggers on create and delete events since VRF names are immutable.
```
IvoGoman: Was wondering why we are not interested in changes to the referenced object:
```suggestion
		// Only triggers on create and delete events since VRF names and deviceRef are immutable.
```

internal/controller/core/evpninstance_controller.go
```diff
@@ -315,6 +342,15 @@ func (r *EVPNInstanceReconciler) reconcile(ctx context.Context, s *eviScope) (re
 		}
 	}
 
+	var vrf *v1alpha1.VRF
+	if s.EVPNInstance.Spec.Type == v1alpha1.EVPNInstanceTypeRouted && s.EVPNInstance.Spec.VRFRef != nil {
```
IvoGoman, line 346: CRD Validiation should ensure that VRFRef is set. Still should a misconfigured EVPNInstance be reflected via status conditions?

Since VRFRef is required when `Type == Route` the VRFRef != nil could be handled in `reconcileVRF(..)` to set the conditions

felix-kaestner: Given the CEL XValidation rules already enforce that, is there ever a case where this would be the useful?

internal/controller/core/evpninstance_controller.go
felix-kaestner: Updated.

internal/controller/core/evpninstance_controller.go
IvoGoman: Don't see any.

IvoGoman (approved):

PR #437, opened 2026-07-09
Update setup instructions

adamtrizuljak-sap:
I am updating the setup instructions based on my recent experience with installing the dev environment on MacOS.

- Add prerequisites into README. I'm just listing the packages and not including exact shell commands, as these will differ per OS and package manager
- Add a new ~`prepare`~ `setup` target that wraps all `make` targets that install dependencies. ~Suggestions for a better name are welcome~ :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,10 @@ Network-operator is a project built using Kubebuilder and controller-runtime to 
 - kubectl version v1.33.1+.
 - Access to a Kubernetes v1.33.0+ cluster.
 - [Git LFS](https://git-lfs.com) installed (`git lfs install`)
+- kind version 0.32.0+
+- Tilt version v0.37.3+
+- gh version 2.93.0+
+- coreutils 9.11+
 
 ### To Deploy on the cluster
 


felix-kaestner (approved):
As for naming suggestions, I could think of "setup" as you already named the branch and PR that way. So I think that would be quite intuitive. But also no strong opinion here.

nikatza (approved):
Nice! Thx!

rgildein:
IMO I will remove all "install-*" command and define list of required tool in README. It should be up to user how to install those tools (brew, macports, nix, apt, snap, source, ...) and on what machine (this does not work on fresh VM with ubuntu).

felix-kaestner:
> IMO I will remove all "install-*" command and define list of required tool in README. It should be up to user how to install those tools (brew, macports, nix, apt, snap, source, ...) and on what machine (this does not work on fresh VM with ubuntu).

On that note, https://github.com/ironcore-dev/network-operator/pull/397 might be a welcome change because it follows the kubebuilder convention of installing required tooling local within the project only (not system-wide) into a `/bin` directory, completely removing the need for these install-* commands.

At the same time, this eliminates the problems of different people having different versions of the same tool installed on their machine, leading to differences in e.g. code generation or formatting, as the versions used by this repo would be pinned.

PR #438, opened 2026-07-10
feat: Implement BFD for BGPPeers

sven-rosenzweig:


api/core/v1alpha1/bfd_types.go
```diff
@@ -0,0 +1,45 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+// Package v1alpha1 contains API Schema definitions for the networking.metal.ironcore.dev v1alpha1 API group.
+// +kubebuilder:validation:Required
+// +kubebuilder:object:generate=true
+// +groupName=networking.metal.ironcore.dev
```
felix-kaestner: These comments shouldn't be here. They are only meant to be there once for the entire package, which `doc.go` already does.

api/core/v1alpha1/bgp_peer_types.go
```diff
@@ -64,6 +64,11 @@ type BGPPeerSpec struct {
 	// LocalAS configures the local AS number and how it factors into BGP announcements for this peer.
 	// +optional
 	LocalAS *LocalAS `json:"localAS,omitempty"`
+
+	// BFD defines the Bidirectional Forwarding Detection configuration for the interface.
+	// BFD is only applicable for Layer 3 interfaces.
```
felix-kaestner, line 69: Interface vs. BGPPeer. Probably just copy+paste.

internal/provider/cisco/nxos/bgp.go
```diff
@@ -187,22 +188,74 @@ func (af *BGPDomAfItem) SetMultipath(m *v1alpha1.BGPMultipath) error {
...
 	} `json:"af-items,omitzero"`
 }
 
+type BGPNeighborBfd struct {
+	DetectMult   uint32 `json:"multiplier"`
+	MinRxIntvlMs uint32 `json:"minRxMs"`
+	MinTxIntvlMs uint32 `json:"minTxMs"`
+}
+
+func NewBGPNeighborBfd(peer *v1alpha1.BGPPeerSpec) (*BGPNeighborBfd, error) {
```
felix-kaestner: It is weird to only pass the spec here. Either I would pass the whole `*v1alpha1.BGPPeer` object, or otherwise, just a `*v1alpha.BFD` (.spec.bfd).

internal/provider/cisco/nxos/bgp.go
```diff
@@ -187,22 +188,74 @@ func (af *BGPDomAfItem) SetMultipath(m *v1alpha1.BGPMultipath) error {
...
+	VRFName string      `json:"-"`
+	Addr    string      `json:"addr"`
+	Name    string      `json:"name,omitempty"`
+	AdminSt AdminSt     `json:"adminSt"`
+	Asn     string      `json:"asn"`
+	AsnType PeerAsnType `json:"asnType"`
+	SrcIf   string      `json:"srcIf,omitempty"`
+
+	// BFD enablement
+	PeerControl string `json:"ctrl"`
```
felix-kaestner: I don't see this field at all in the YANG model. Should we delete it?

internal/provider/cisco/nxos/bgp.go
```diff
@@ -187,22 +188,74 @@ func (af *BGPDomAfItem) SetMultipath(m *v1alpha1.BGPMultipath) error {
...
+	// Bfd Type: none (default), single-hop, multi-hop
```
felix-kaestner, line 203: I would create a proper enum for these string values instead of documenting supported values in a comment.

internal/provider/cisco/nxos/bgp.go
```diff
@@ -187,22 +188,74 @@ func (af *BGPDomAfItem) SetMultipath(m *v1alpha1.BGPMultipath) error {
...
+			return nil, apistatus.NewInvalidArgumentError(apistatus.FieldViolation{
+				Field:       "spec.bfd.detectionMultiplier",
+				Description: "BFD detection multiplier must be non-negative",
+			})
+		}
+		bfd.DetectMult = uint32(multiplier)
+	}
+	if peer.BFD.RequiredMinimumReceive != nil {
+		ms := peer.BFD.RequiredMinimumReceive.Duration
+
```
felix-kaestner:
```suggestion
```
weird formatting.

internal/provider/cisco/nxos/bgp.go
```diff
@@ -362,6 +415,13 @@ const (
 
 const RouteReflectorClient = "rr-client"
 
+const (
+	// PeerControlBFD enables BFD on BGP peer
+	PeerControlBFD = "bfd"
+	// BfdTypeNone relies on default BFD session type selection
+	BfdTypeNone = "none"
+)
+
```
felix-kaestner, line 429: I would make this a proper enum instead of constants:

```go
type PeerControlType string

const (
	PeerControlTypeBFD PeerControlType = "bfd"
	PeerControlTypeNone PeerControlType = "none"
)
```

internal/provider/cisco/nxos/provider.go
```diff
@@ -539,6 +539,17 @@ func (p *Provider) EnsureBGPPeer(ctx context.Context, req *provider.EnsureBGPPee
...
+	if req.BGPPeer.Spec.BFD != nil && req.BGPPeer.Spec.BFD.Enabled {
+		bfd, err := NewBGPNeighborBfd(&req.BGPPeer.Spec)
+		if err != nil {
+			return err
+		}
+		pe.PeerControl = PeerControlBFD
+		pe.BfdType = BfdTypeNone
+		pe.BfdMultihop = bfd
+	}
+
```
felix-kaestner, line 796:
```suggestion
    pe.BfdType = BfdTypeNone

	if req.BGPPeer.Spec.BFD != nil && req.BGPPeer.Spec.BFD.Enabled {
		bfd, err := NewBGPNeighborBfd(&req.BGPPeer.Spec)
		if err != nil {
			return err
		}
		pe.PeerControl = PeerControlBFD
		pe.BfdType = "multihop"
		pe.BfdMultihop = bfd
	}
```
we have to set the default values unconditionally, otherwise we risk trigger a diff → update cycle on every reconcile.

internal/provider/cisco/nxos/bgp.go
sven-rosenzweig: ``*v1alpha.BFD`` it is.

internal/provider/cisco/nxos/bgp.go
sven-rosenzweig: I somehow have in my had, this was needed to activate the BFD feature. 

This should be the field:
https://github.com/YangModels/yang/blob/559418adbd5f05860c36b1d2ffa590ef8ee0309e/vendor/cisco/nx/10.6-2/Cisco-NX-OS-device.yang#L48686-L48690

internal/provider/cisco/nxos/bgp.go
felix-kaestner: Still valid.

internal/provider/cisco/nxos/bgp.go
```diff
@@ -189,22 +190,74 @@ func (af *BGPDomAfItem) SetMultipath(m *v1alpha1.BGPMultipath) error {
...
+	VRFName string      `json:"-"`
+	Addr    string      `json:"addr"`
+	Name    string      `json:"name,omitempty"`
+	AdminSt AdminSt     `json:"adminSt"`
+	Asn     string      `json:"asn"`
+	AsnType PeerAsnType `json:"asnType"`
+	SrcIf   string      `json:"srcIf,omitempty"`
+
+	// BFD enablement
+	PeerControl PeerControlType `json:"ctrl,omitempty"`
```
felix-kaestner: Do we really want this to be `omitempty`? Does this not have a platform default value that is always present? If not and it's completely omitted from the json payload, than we should rather use the `Option[T]` type we have in this `nxos` package.

internal/provider/cisco/nxos/provider.go
```diff
@@ -548,6 +548,17 @@ func (p *Provider) EnsureBGPPeer(ctx context.Context, req *provider.EnsureBGPPee
...
 	pe.VRFName = bgp.Name
 	pe.Addr = req.BGPPeer.Spec.Address
+
+	pe.BfdType = BfdTypeNone
+	if req.BGPPeer.Spec.BFD != nil && req.BGPPeer.Spec.BFD.Enabled {
+		bfd, err := NewBGPNeighborBfd(req.BGPPeer.Spec.BFD)
+		if err != nil {
+			return err
+		}
+		pe.PeerControl = PeerControlTypeBFD
```
felix-kaestner: If this field is only conditionally present, it should rather use the `Option` type than be `omitempty`.

internal/provider/cisco/nxos/provider.go
```diff
@@ -548,6 +548,17 @@ func (p *Provider) EnsureBGPPeer(ctx context.Context, req *provider.EnsureBGPPee
...
+		pe.BfdMultihop = bfd
```
felix-kaestner: As this field is only set, when bfd is in the spec and never set or reset otherwise, please add a comment that this will not work with a `gnmiext.Client.Patch` call. The reason is that bfd might have been configured outside of the network operator and thus, the "mhbfdintvl-items" entry is present. However, as we don't set it (because the spec doesn't specify it), we would unmarshal the response and detect a drift/diff on every reconcile, making gnmi set calls over and over again to Patch the configuration. Currently this is not the case, as we are Updating/Replacing the configuration, however with a Patch we would have a problem here. So I would prefer if we document this in a little comment.

internal/provider/cisco/nxos/bgp.go
felix-kaestner: I don't see this field being present in the payload at all:

```sh
leaf1(config-router-neighbor)# sh run | sec '^router bgp'
router bgp 65000
  neighbor 10.1.1.1
    bfd multihop interval 999 min_rx 999 multiplier 10
```

```sh
$ gnmic --skip-verify get --type config --values-only --path 'System/bgp-items/inst-items/dom-items/Dom-list[name=default]/peer-items/Peer-list[addr=10.1.1.1]'
[
  [
    {
      "addr": "10.1.1.1",
      "adminSt": "enabled",
      "asnType": "none",
      "bfdType": "none",
      "bmpSrvId1St": "disabled",
      "bmpSrvId2St": "disabled",
      "capSuppr4ByteAsn": "disabled",
      "dscp": "cs6",
      "epe": "disabled",
      "holdIntvl": 180,
      "internalVpnClient": "disabled",
      "kaIntvl": 60,
      "logNbrChgs": "none",
      "lowMemExempt": "disabled",
      "mhbfdintvl-items": {
        "minRxMs": 999,
        "minTxMs": 999,
        "multiplier": 10
      },
      "passwdType": "LINE",
      "peerType": "fabric-internal",
      "privateASctrl": "none"
    }
  ]
]
```

internal/provider/cisco/nxos/provider.go
```diff
@@ -548,6 +548,17 @@ func (p *Provider) EnsureBGPPeer(ctx context.Context, req *provider.EnsureBGPPee
 	pe := new(BGPPeer)
 	pe.VRFName = bgp.Name
 	pe.Addr = req.BGPPeer.Spec.Address
+
+	pe.BfdType = BfdTypeNone
+	if req.BGPPeer.Spec.BFD != nil && req.BGPPeer.Spec.BFD.Enabled {
+		bfd, err := NewBGPNeighborBfd(req.BGPPeer.Spec.BFD)
```
felix-kaestner: You have to make sure that the bfd feature is also enabled here, otherwise the gnmi call will fail. I assume bfd was already enabled elsewhere on the device you were testing against? 

You can just take the logic for enabling bfd from the `EnsureInterface` func.

felix-kaestner (changes requested):

internal/provider/cisco/nxos/provider.go
sven-rosenzweig: Honestly, personally I could not remember activating it, but anyhow you are right with this: 
``error:operation-failed msg:BFD feature is not enabled``

internal/provider/cisco/nxos/bgp.go
sven-rosenzweig: ``
show running-config bfd
router bgp 65000
  vrf vrf-cc-admin
    neighbor 10.0.0.1
      bfd
      bfd multihop interval 250 min_rx 250 multiplier 7
``
adding ``bfd`` sets th ``ctrl`` field. according to the documentation, it is responsible for activating bfd. 
At least from my understanding, the bfd session parameters are configured under a separate path (corresponding to the second line). So I think we need both.


[bfd config](https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus3000/sw/unicast/503_u3_1/l3_nx-os/l3_bfd_for_bgp.pdf)

sven-rosenzweig: To my knowledge this does not have a platform default value which is always present. Once somebody configures bfd for the bgppeer, this should be set in order to activate bfd. 

I will take a look at the ``Option[T]`` type.

internal/provider/cisco/nxos/provider.go
sven-rosenzweig: using the ``Option`` type now.

PR #439, opened 2026-07-10
[NX-OS] Add EVPN multihoming core-tracking to InterfaceConfig

felix-kaestner:
Introduce EVPNMultihoming struct in the NX-OS InterfaceConfig API with a CoreTracking field. When enabled on uplink interfaces, the switch shuts down ESI-attached access links if all tracked core-links go down, preventing traffic blackholing.

Wire the field into EnsureInterface for physical interfaces and add a golden test case for the YANG payload.


internal/provider/cisco/nxos/provider.go
```diff
@@ -984,6 +984,10 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
 			p.PhysExtdItems.BufferBoost = AdminStDisable
 		}
 
+		if cfg.Spec.EVPNMultihoming != nil && cfg.Spec.EVPNMultihoming.CoreTracking {
```
sven-rosenzweig, line 993: Is it enough for us to enable this option ``AdminStEnabled`` or do we benefit from setting it to ``AdminStDisEnabled``?

sven-rosenzweig (approved):

internal/provider/cisco/nxos/provider.go
felix-kaestner: No, when disabled, there is simply no `esimhcoretracking-items` subtree in the physical interface YANG tree.

PR #445, opened 2026-07-15
Add SetBuilder for batched gNMI Set operations

felix-kaestner:
Introduce a SetBuilder type that accumulates update, patch, and delete operations to be executed as a single gNMI Set RPC via the new Client.Do method. This allows atomic multi-operation commits without multiple round-trips to the device.

An optional Limit method controls the maximum number of operations per SetRequest, splitting into multiple RPCs when set.


internal/transport/gnmiext/client.go
```diff
@@ -10,11 +10,12 @@ import (
 	"fmt"
 	"log/slog"
 	"reflect"
+	"slices"
 	"strings"
 
 	cp "github.com/felix-kaestner/copy"
 	"github.com/go-logr/logr"
-	gpb "github.com/openconfig/gnmi/proto/gnmi"
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
```
felix-kaestner, line 19: "pb" is the canonical suffix for packages exposing protobuf generated grpc client/server code. ref/ https://pkg.go.dev/google.golang.org/protobuf/types/known

internal/transport/gnmiext/client.go
```diff
@@ -216,10 +298,23 @@ func (c *client) get(ctx context.Context, dt gpb.GetRequest_DataType, el ...Data
...
+	op := "get_config"
+	if dt == gnmipb.GetRequest_STATE {
+		op = "get_state"
+	}
+	start := time.Now()
 	res, err := c.gnmi.Get(ctx, r)
 	if err != nil {
+		rpcDurationSeconds.WithLabelValues(c.target, "Get", "error").Observe(time.Since(start).Seconds())
+		for _, e := range el {
+			operationsTotal.WithLabelValues(c.target, "Get", op, e.XPath(), "error").Inc()
```
felix-kaestner: TODO: Take out dynamic parts, like the yang key.

felix-kaestner: Done.

internal/transport/gnmiext/client.go
```diff
@@ -61,10 +63,66 @@ type Capabilities struct {
 	SupportedModels []Model
 }
 
+// mode represents the type of gNMI Set operation.
+type mode uint8
+
+const (
+	replace mode = iota + 1
+	update
+	del
```
IvoGoman: NIT: `replace` & `update` are not abbreviated only `delete`, which is the same length as update.

internal/transport/gnmiext/client.go
```diff
@@ -61,10 +63,66 @@ type Capabilities struct {
...
+	limit int
+}
+
+// Update queues a full replacement of the given elements (gNMI replace).
+func (b *SetBuilder) Update(el ...DataElement) *SetBuilder {
+	for _, e := range el {
+		b.ops = append(b.ops, operation{el: e, mode: replace})
+	}
+	return b
+}
```
IvoGoman, line 94: NIT: L77 uses `el` to abbreviate `DataElement`. Here `el` is used as "DataElement**s**" and `e` is the DataElement.
Ranging over `el` to assign the loopvar `e` to field `el` in the `operation` struct looks off to me.

internal/transport/gnmiext/metrics.go
```diff
@@ -0,0 +1,60 @@
...
+
+	gnmipb "github.com/openconfig/gnmi/proto/gnmi"
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	operationsTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "gnmi_operations_total",
```
IvoGoman: Could we add a configurable Namespace that defaults to `network_operator`? This way it is easier to find all metrics emitted by the network_operator when browsing Prometheus.
Metrics emitted via kube-state-metrics already use the `network_operator` as a prefix.

internal/transport/gnmiext/client.go
```diff
@@ -291,14 +385,15 @@ func (c *client) set(ctx context.Context, patch bool, el ...DataElement) error {
...
 		if err == nil && reflect.DeepEqual(e, got) {
 			c.logger.V(2).Info("Configuration is already up-to-date", "path", e.XPath())
+			operationsSkippedTotal.WithLabelValues(c.target, op, metricPath(path)).Inc()
 			continue
 		}
 		b, err := c.Marshal(e)
 		if err != nil {
 			return err
 		}
 		c.logger.V(1).Info("Updating", "path", e.XPath(), "payload", string(b), "patch", patch)
```
IvoGoman: WDYT about replacing the patch flag with the operation string value?
```suggestion
		c.logger.V(1).Info("Updating", "path", e.XPath(), "payload", string(b), "operation", op)
```

internal/transport/gnmiext/client.go
```diff
@@ -308,11 +403,59 @@ func (c *client) set(ctx context.Context, patch bool, el ...DataElement) error {
...
 	}
-	if len(r.GetUpdate()) == 0 && len(r.GetReplace()) == 0 {
-		// All configurations are already up-to-date.
+	return nil
+}
+
+// delete appends delete operations to the given SetRequest.
+// If an element implements [Defaultable], a replace-with-default is
+// appended instead of a path deletion.
+func (c *client) delete(_ context.Context, r *gnmipb.SetRequest, el ...DataElement) error {
```
IvoGoman:
```suggestion
func (c *client) delete(r *gnmipb.SetRequest, el ...DataElement) error {
```

internal/transport/gnmiext/client.go
```diff
@@ -61,10 +63,66 @@ type Capabilities struct {
...
+	return b
+}
+
+// Patch queues a partial merge of the given elements (gNMI update).
+func (b *SetBuilder) Patch(el ...DataElement) *SetBuilder {
+	for _, e := range el {
+		b.ops = append(b.ops, operation{el: e, mode: update})
+	}
+	return b
+}
```
IvoGoman, line 102: The `mode` is a bit confusing here. `Update(..)` uses mode _replace_ and `Patch(..)` uses mode _update_.

Then later on the Client uses `patch=true` when processing the  Update from the SetBuilder.

https://github.com/ironcore-dev/network-operator/pull/445/changes#diff-3075c8571e0e3bd6b36569701251db80196adc3d94f202f23f871685f97c73caR266-R269

Update -> mode: replace -> patch: false

Patch -> mode: update -> patch: true

I realise this comes from the gNMI operation. Still it took me a bit to understand why I was thinking something is off.

IvoGoman (commented):
Metric namespace would be nice to have. Comments about readability are up to you.

internal/transport/gnmiext/client.go
felix-kaestner: The reason `delete` was abbreviated is because it's a keyword in Go and therefore we can't use that. Do you have a better suggestion? I think abbreviating everything would suffer in readability, so I opted for the asymmetry here.

internal/transport/gnmiext/client.go
felix-kaestner: Changed.

internal/transport/gnmiext/metrics.go
felix-kaestner: Hardcoded `network_operator` as the namespace for now.

internal/transport/gnmiext/client.go
felix-kaestner: Adopted.

internal/transport/gnmiext/client.go
felix-kaestner: Changed `el` → `elements`. However, I would keep the `ctx` argument just for symmetry with the other calls like `get` and `set`.

internal/transport/gnmiext/client.go
felix-kaestner: Yes, this is unfortunately the naming that the gNMI operations have. However, for the `gnmiext.Client` we opted to have the same methods like `Update` and `Patch` as the Kubernetes client has. Therefore we ended of with that naming for the client methods. But I see how this might be confusing with the gnmi operations. I've added some comments to the operation constants to make this clear.

IvoGoman: Thanks for the details.

IvoGoman (approved):

PR #446, opened 2026-07-16
Add Prometheus metrics for NX-API client

felix-kaestner:
Register two metrics with controller-runtime's metrics registry:
- nxapi_rpc_duration_seconds: histogram of HTTP round-trip latency with target and status labels
- nxapi_rpc_commands_total: per-command counter with target, command, and status labels

Add WithTarget option to associate a device address with metrics for correlating slow-responding devices.

### DIFF
--- a/internal/transport/nxapi/nxapi.go
+++ b/internal/transport/nxapi/nxapi.go
@@ -34,6 +34,7 @@ func (f RoundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
 type Client struct {
 	client *http.Client
 	url    url.URL
+	target string
 }
 
 // Option configures a [Client].
@@ -58,6 +59,14 @@ func WithPort(port string) Option {
 func WithTimeout(d time.Duration) Option {
 	return func(c *Client) error {
 		c.client.Timeout = d
+		return nil
+	}
+}
+
+// WithTarget sets the device target label used in metrics.
+func WithTarget(target string) Option {
+	return func(c *Client) error {
+		c.target = target
 		return nil
 	}
 }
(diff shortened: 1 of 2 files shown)


internal/transport/nxapi/nxapi.go
```diff
@@ -33,13 +33,14 @@ func (f RoundTripFunc) RoundTrip(r *http.Request) (*http.Response, error) {
...
 	uri    string
+	target string
 }
 
 // NewClient creates a new [Client] for the given connection.
 // If the connection has a TLS configuration set, HTTPS is used; otherwise HTTP.
 // The underlying HTTP client uses timeout for all requests; a value of 0
 // means no timeout.
-func NewClient(c *deviceutil.Connection, timeout time.Duration) (*Client, error) {
+func NewClient(c *deviceutil.Connection, timeout time.Duration, opts ...Option) (*Client, error) {
```
IvoGoman: Wondering if `timeout` should also be treated as an option.

internal/transport/nxapi/nxapi.go
```diff
@@ -63,7 +64,21 @@ func NewClient(c *deviceutil.Connection, timeout time.Duration) (*Client, error)
...
+}
+
+// Option configures a [Client].
+type Option func(*Client)
+
+// WithTarget sets the device target label used in metrics.
+func WithTarget(target string) Option {
+	return func(c *Client) {
+		c.target = target
+	}
```
IvoGoman: Will the usage come with a follow-up PR?

felix-kaestner: Yes, that is the plan.

internal/transport/nxapi/nxapi.go
felix-kaestner: Implemented in https://github.com/ironcore-dev/network-operator/pull/381/changes/533f1568f9dd5a903a3365735fde0dd3b25a3381 as part of #381.

IvoGoman (approved):

PR #452, opened 2026-07-23
Implement EVPN Fabric overlay BGP reconciliation

felix-kaestner:
Add a reconcileOverlayBGP phase that creates BGP instances on all fabric devices and BGPPeer resources per route reflector group.

BGP instances are configured with the shared iBGP ASN, router-ID from lo0, and L2VPN EVPN with retainAll enabled. For RR-to-client groups, spine→leaf peers are marked as route reflector clients. When deviceSelector equals clientDeviceSelector, a full mesh is created instead.

Remove the scaffolded stub test that fails CRD validation and suppress the nilnil lint on the not-ready-yet return path.

### DIFF
--- a/internal/controller/evpn/fabric_controller.go
+++ b/internal/controller/evpn/fabric_controller.go
@@ -57,6 +57,8 @@ type FabricReconciler struct {
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=interfaces,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=ospf,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=isis,verbs=get;list;watch;create;update;patch;delete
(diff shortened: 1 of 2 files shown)


internal/controller/evpn/fabric_controller.go
```diff
@@ -739,6 +744,207 @@ func isisNETFromIPv4(addr string) (string, error) {
...
+			},
+		}
+		return controllerutil.SetControllerReference(fabric, peer, r.Scheme)
+	})
+	if err != nil {
+		return fmt.Errorf("reconciling BGPPeer %s: %w", name, err)
+	}
+	if res == controllerutil.OperationResultCreated {
+		r.Recorder.Eventf(fabric, nil, "Normal", "BGPPeerCreated", "Reconcile", "Created overlay BGPPeer %s", name)
+	}
```
IvoGoman, line 944: Adding an Event for the update path could be helpful for the enduser. In case that manual changes are done to these Fabric-owned BGPPeer resources it brings visibility to why these changes were removed.

Same could be done after the BGP CreateOrPatch

IvoGoman (approved):

internal/controller/evpn/fabric_controller.go
felix-kaestner: I guess this applies to all resources that are created, as all of them follow this pattern currently. Will do this as a follow up.

PR #453, opened 2026-07-23
Implement EVPN Fabric multicast PIM reconciliation

felix-kaestner:
Add a reconcileMulticastPIM phase that creates one PIM resource per device participating in any anycast rendezvous point group.

RP devices get anycastAddresses populated with peer RP lo0 IPs for redundancy. Client devices get a static RP address pointing to the shared anycast IP (lo100). PIM sparse mode is enabled on lo0, lo100 or lo1, and all fabric uplinks.

### DIFF
--- a/internal/controller/evpn/fabric_controller.go
+++ b/internal/controller/evpn/fabric_controller.go
@@ -59,6 +59,7 @@ type FabricReconciler struct {
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=isis,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=bgp,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=bgppeers,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=networking.metal.ironcore.dev,resources=pim,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=claims,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=ipaddresspools,verbs=get;list;watch
 // +kubebuilder:rbac:groups=pool.networking.metal.ironcore.dev,resources=ipprefixpools,verbs=get;list;watch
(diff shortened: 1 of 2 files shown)


internal/controller/evpn/fabric_controller.go
```diff
@@ -947,6 +950,197 @@ func (r *FabricReconciler) reconcileBGPPeer(ctx context.Context, local, remote *
...
+		}
+		clientDevices := &v1alpha1.DeviceList{}
+		if err := r.List(ctx, clientDevices, client.InNamespace(fabric.Namespace), client.MatchingLabelsSelector{Selector: clientSelector}); err != nil {
+			return ctrl.Result{}, fmt.Errorf("listing client devices for group %q: %w", rpGroup.Name, err)
+		}
+
+		// Resolve the anycast RP IP from the first RP device's lo100.
+		if len(rpDevices.Items) == 0 {
+			continue
+		}
```
IvoGoman: This would silently ignore empty `deviceSelectors`, which could be due to a misconfiguration.
Should we emit an event to make this visible to the enduser?

internal/controller/evpn/fabric_controller.go
```diff
@@ -947,6 +950,197 @@ func (r *FabricReconciler) reconcileBGPPeer(ctx context.Context, local, remote *
...
+
+			lo0Idx := slices.IndexFunc(loopbacks, func(intf *v1alpha1.Interface) bool { return intf.Name == lo0Name })
+			if lo0Idx < 0 {
+				continue
+			}
+
+			lo0 := loopbacks[lo0Idx]
+			if lo0.Spec.IPv4 == nil || len(lo0.Spec.IPv4.Addresses) == 0 {
+				continue
+			}
```
IvoGoman: This should not happen, as https://github.com/ironcore-dev/network-operator/blob/69211b51f4042d3ddc1eee2280acace245d57848/internal/controller/evpn/fabric_controller.go#L268-L271 will early-exit if the lo0 could not be created. So the reconcileState should always have a valid lo0 for the device.
Since this is unexpected I think it could be worth to at least log it for faster troubleshooting.

internal/controller/evpn/fabric_controller.go
IvoGoman: Same for the clientDeviceSelector, if there are no client devices matching it won't be noticed. Logging/events should make troubleshooting a Fabric resource regarding bad selectors easier.

IvoGoman, line 1004: This silently skips configuration if no devices are found. It could be that the devices were not yet created or that the selector is misconfigured.
Should it emit a log/event here to make it visible to an enduser that the PIM configuration was skipped?

If clientDevices is empty it will only reconcile the rpDevices. Same question if we should emit a log/event here for the enduser?

IvoGoman (approved):

internal/controller/evpn/fabric_controller.go
felix-kaestner: Emitted a warning in this case.

IvoGoman (approved):

PR #467, opened 2026-07-27
feat(iosxr): Implement Route Policies

sven-rosenzweig:
Compared with OpenConfig and Cisco NX-OS, IOS-XR-based route policies are more powerful and feature-rich. However, we currently do not need most of these advanced features. Instead, we adapt to the simpler route-map configuration model (sequence, condition, action, and disposition).

Route maps work by iterating over each entry in order of their sequence number. If a condition matches, the configured action is executed and processing stops after a match is found. The default action when no match is found is to pass.

We assume the same behavior with route policies: after each condition check, we set the policy result to "done" (which stops processing and accepts the route) or "drop" (which stops processing and rejects the route). At the end, if no match is found, the route is accepted by default.

Some key aspects that make Route Policies richer than route maps:
* Support for if/else control flow
* Referencing other RPLs within each other, creating nested blocks of policy logic
* Boolean operators (and, not, or)
* Parameterized route policy configuration
* Wildcard support in setting values

Yang support for Cisco IOS XR route policy support also differs from other OS models. It simply accepts one single policy string, being correctly formatted in the way mentioned above:
```
route-policy <name>
  if <sample statement> then
     set <sample set action>
     done
  endif
end-policy
```

https://github.com/YangModels/yang/blob/ad7aa9d483c592661c1df3d38bdfc99dc01cba56/vendor/cisco/xr/2522/Cisco-IOS-XR-um-route-policy-cfg.yang#L81-L94

### DIFF
--- a/internal/provider/cisco/iosxr/route_policies.go
+++ b/internal/provider/cisco/iosxr/route_policies.go
@@ -4,13 +4,22 @@
 package iosxr
 
 import (
+	"errors"
 	"fmt"
+	"slices"
+	"strconv"
+	"strings"
 
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/provider"
 	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
 )
 
 var _ gnmiext.DataElement = (*RoutePolicy)(nil)
 
+// RoutePolicy represents an IOS-XR route policy configuration element.
 type RoutePolicy struct {
 	Name string `json:"route-policy-name"`
 	Body string `json:"rpl-route-policy"`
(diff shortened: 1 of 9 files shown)


internal/provider/cisco/iosxr/testdata/routepolicy.json
```diff
@@ -0,0 +1,10 @@
+ {
+    "routing-policy": {
+      "route-policies": {
+        "route-policy": {
+      "route-policy-name": "TEST-POLICY",
+      "rpl-route-policy": "route-policy TEST-POLICY\n  if destination in (10.0.0.0/8) then\n    set community (65000:100)\n    done\n  endif\nend-policy\n"
+    }
```
felix-kaestner, line 7: nit: could benefit from a quick pass through `jq` for formatting.

internal/provider/cisco/iosxr/route_policies.go
```diff
@@ -4,13 +4,22 @@
...
+	"k8s.io/apimachinery/pkg/util/intstr"
+
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/provider"
 	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
 )
 
 var _ gnmiext.DataElement = (*RoutePolicy)(nil)
 
+// RoutePolicy represents an IOS-XR route policy configuration element.
```
felix-kaestner, line 22:
```suggestion
// RoutePolicy represents an route policy configuration element.
```
nit: IOS-XR is already clear from the package.

internal/provider/cisco/iosxr/route_policies.go
```diff
@@ -20,10 +29,317 @@ func (rp *RoutePolicy) XPath() string {
...
+	Statement PolicyStatement
+	Name      string
+}
+
+// NewPolicyString creates a PolicyString from provider-agnostic policy statements.
+// Statements are sorted by sequence number and linked in evaluation order.
+func NewPolicyString(name string, providerStatements []provider.PolicyStatement) (*PolicyString, error) {
+	// Sort statements by sequence number (lowest first)
+	statements := make([]provider.PolicyStatement, len(providerStatements))
+	copy(statements, providerStatements)
```
felix-kaestner, line 53: I think there is a `slices.Clone` in the `slices` stdlib package.

internal/provider/cisco/iosxr/route_policies_test.go
```diff
@@ -0,0 +1,272 @@
...
+
+	return string(content)
+}
+
+// ptr returns a pointer to the given value
+//
+//nolint:modernize // ptr helper is needed for value-returning functions like intstr.FromInt32
+func ptr[T any](v T) *T {
+	return &v
+}
```
felix-kaestner, line 76: Go 1.26 just supports `new(int32(6))` so such helper is not needed in any case.

PR #470, opened 2026-07-30
feat/multi provider support

sven-rosenzweig:
Migrate to Multi Provider Support
    
With this commit, we do not use the provider flag anymore.
The provider func now is loaded dynamically on every reconciliation.
Requires the `device.spec.provider` field to be set on every device.


sven-rosenzweig:
Do not merge until the ``device.spec.provider`` field is introduced everywhere.

api/core/v1alpha1/device_types.go
```diff
@@ -24,6 +24,11 @@ type DeviceSpec struct {
 	// +required
 	Endpoint Endpoint `json:"endpoint"`
 
+	// Provider identifies the network provider plugin used to reconcile this Device and all CRDs that reference it.
```
felix-kaestner:
```suggestion
	// Provider identifies the provider implementation used to reconcile this Device and all CRDs that reference it.
```

internal/controller/cisco/nx/bordergateway_controller.go
```diff
@@ -88,22 +85,39 @@ func (r *BorderGatewayReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
-	prov, ok := r.Provider().(Provider)
-	if !ok {
+	device, err := deviceutil.GetDeviceByName(ctx, r, obj.Namespace, obj.Spec.DeviceRef.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if device.Spec.Provider == "" {
+		return ctrl.Result{}, fmt.Errorf("device %q has no provider set", device.Name)
+	}
```
felix-kaestner: This is not possible as this is a required field. Therefore I wouldn't check this.

Applies to all controllers.

api/core/v1alpha1/device_types.go
```diff
@@ -24,6 +24,11 @@ type DeviceSpec struct {
...
+	// +required
+	// +immutable
```
felix-kaestner:
```suggestion
	// Immutable.
	// +required
	// +kubebuilder:validation:MinLength=1
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Provider is immutable"
```
There is no `+immutable` marker.

internal/controller/core/device_controller.go
```diff
@@ -50,7 +50,7 @@ type DeviceReconciler struct {
 	Recorder events.EventRecorder
 
 	// Provider is the driver that will be used to create & delete the interface.
-	Provider provider.ProviderFunc
+	// Provider provider.ProviderFunc
```
felix-kaestner: Why commented out?

internal/provider/provider.go
```diff
@@ -899,3 +901,45 @@ type ProviderConfig struct { //nolint:revive // stutter is intentional; Provider
...
+}
+
+func (e NotImplementedError) Is(target error) bool {
+	_, ok := target.(NotImplementedError)
+	return ok
+}
+
+// LoadProvider returns a provider instance cast to the requested interface type T.
+// Returns NotFoundError if the provider is not registered, NotImplementedError if it does not implement T.
+func LoadProvider[T Provider](providerName string) (T, error) {
```
felix-kaestner:
```suggestion
func LoadProvider[T Provider](providerName string) (zero T, _ error) {
```
Safes you the `var zero T` statements.

internal/provider/provider.go
```diff
@@ -899,3 +901,45 @@ type ProviderConfig struct { //nolint:revive // stutter is intentional; Provider
...
+	prov, err := Get(providerName)
+	if err != nil {
+		var zero T
+		return zero, NotFoundError{Message: fmt.Sprintf("Provider %q is not registered", providerName)}
+	}
+	genericProvider, ok := prov().(T)
```
felix-kaestner:
```suggestion
	provider, ok := prov().(T)
```
That name doesn't seem fitting.

internal/provider/provider.go
```diff
@@ -899,3 +901,45 @@ type ProviderConfig struct { //nolint:revive // stutter is intentional; Provider
...
+
+func (e NotFoundError) Is(target error) bool {
+	_, ok := target.(NotFoundError)
+	return ok
+}
+
+func (e NotImplementedError) Is(target error) bool {
+	_, ok := target.(NotImplementedError)
+	return ok
+}
```
felix-kaestner, line 929:
```suggestion
type NotFoundError struct
func (e NotFoundError) Error() string 
func (e NotFoundError) Is(target error) bool

type NotImplementedError struct 
func (e NotImplementedError) Error() string
func (e NotImplementedError) Is(target error) bool
```
This order is a bit better.

internal/provisioning/http.go
```diff
@@ -91,8 +91,8 @@ type HTTPServer struct {
 	Mux              *http.ServeMux
 	Recorder         events.EventRecorder
 	ValidateSourceIP bool
-	Provider         provider.ProvisioningProvider
-	Port             int
+	// Provider         provider.ProvisioningProvider
```
felix-kaestner: Why commented out?

Tiltfile
```diff
@@ -21,10 +21,8 @@ docker_build('controller:latest', '.', only=[
 local_resource('controller-gen', 'make generate', deps=['api/', 'hack/boilerplate.go.txt'])
 local_resource('crds', 'make install', deps=['api/'])
 
-provider = os.getenv('PROVIDER', 'openconfig')
```
felix-kaestner, line 24: We wanted to still keep supporting this env variable for local development and set it on the device sample via the `device_yaml` function.

```diff
diff --git a/Tiltfile b/Tiltfile
index 65ac2499..0828a89e 100644
--- a/Tiltfile
+++ b/Tiltfile
@@ -21,6 +21,8 @@ docker_build('controller:latest', '.', only=[
 local_resource('controller-gen', 'make generate', deps=['api/', 'hack/boilerplate.go.txt'])
 local_resource('crds', 'make install', deps=['api/'])
 
+provider = os.getenv('PROVIDER', 'openconfig')
+
 manager = kustomize('config/develop')
 manager = str(manager)
 
@@ -30,6 +32,7 @@ k8s_resource('network-operator-controller-manager', resource_deps=['controller-g
 # Sample resources with manual trigger mode
 def device_yaml():
     decoded = read_yaml_stream('./config/samples/v1alpha1_device.yaml')
+    decoded[0]['spec']['provider'] = provider
     ip = str(local("docker run --rm busybox:1.37.0 nslookup -type=a host.docker.internal 2>/dev/null | grep 'Address:' | tail -n 1 | awk '{print $2}' || echo ''", quiet=True)).rstrip('\n')
     if len(ip) > 0:
         decoded[0]['spec']['endpoint']['address'] = ip+':9339'
```

api/core/v1alpha1/device_types.go
```diff
@@ -24,6 +24,11 @@ type DeviceSpec struct {
...
+	Provider string `json:"provider,omitempty"`
```
felix-kaestner:
```suggestion
	Provider string `json:"provider"`
```
No omitempty when the field is required.

internal/controller/cisco/nx/bordergateway_controller.go
sven-rosenzweig: True, I guess this is an leftover as I had the field optional for the first version. thanks for paying attention.

internal/controller/core/device_controller.go
sven-rosenzweig: Leftover of my development process. No sepcific reason for this.

internal/provisioning/http.go
sven-rosenzweig: No specific reason.

Tiltfile
sven-rosenzweig: make sense.

api/core/v1alpha1/device_types.go
sven-rosenzweig: fixed.

internal/controller/cisco/nx/bordergateway_controller.go
```diff
@@ -88,22 +85,35 @@ func (r *BorderGatewayReconciler) Reconcile(ctx context.Context, req ctrl.Reques
...
 	}
 
-	prov, ok := r.Provider().(Provider)
-	if !ok {
+	device, err := deviceutil.GetDeviceByName(ctx, r, obj.Namespace, obj.Spec.DeviceRef.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	p, err := provider.Get(device.Spec.Provider)
```
felix-kaestner: Any reason we are not using the `provider.LoadProvider` in the `internal/controller/cisco/nx/` package?

internal/provider/provider.go
```diff
@@ -985,3 +987,43 @@ type ProviderConfig struct { //nolint:revive // stutter is intentional; Provider
...
+	prov, err := Get(providerName)
+	if err != nil {
+		return zero, NotFoundError{Message: fmt.Sprintf("Provider %q is not registered", providerName)}
+	}
+	provider, ok := prov().(T)
+	if !ok {
+		return zero, NotImplementedError{Message: fmt.Sprintf("provider %q does not implement %T", providerName, zero)}
+	}
+	return provider, nil
+}
```
felix-kaestner, line 1025: nit: I think having only a `Message` field in an error type is rather uncommon, you would rather have real struct fields and then formulate the error string in the `Error()` func, like:

```go

type NotFoundError struct {
	ProviderName string
}

func (e NotFoundError) Error() string {
	return fmt.Sprintf("provider %q is not registered", e.ProviderName)
}

type NotImplementedError[T any] struct {
	ProviderName string
}

func (e NotImplementedError[T]) Error() string {
	var zero T
	return fmt.Sprintf("provider %q does not implement %T", e.ProviderName, zero)
}

// LoadProvider returns a provider instance cast to the requested interface type T.
// Returns NotFoundError if the provider is not registered, NotImplementedError if it does not implement T.
func LoadProvider[T Provider](providerName string) (zero T, _ error) {
	prov, err := Get(providerName)
	if err != nil {
		return zero, NotFoundError{ProviderName: providerName}
	}
	provider, ok := prov().(T)
	if !ok {
		return zero, NotImplementedError[T]{ProviderName: providerName}
	}
	return provider, nil
}
```

For your usecase you also don't need to implement the `Is()` method, as a simple comparison is already sufficient, see https://cs.opensource.google/go/go/+/refs/tags/go1.27.1:src/errors/wrap.go;l=56

And in the controllers, you would also prefer the `errors.AsType()` over the  `errors.Is`.

```diff
diff --git a/internal/controller/core/interface_controller.go b/internal/controller/core/interface_controller.go
index 0abe9493..0f945428 100644
--- a/internal/controller/core/interface_controller.go
+++ b/internal/controller/core/interface_controller.go
@@ -102,7 +102,7 @@ func (r *InterfaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	prov, err := provider.LoadProvider[provider.InterfaceProvider](device.Spec.Provider)
 	if err != nil {
 		reason := v1alpha1.NotImplementedReason
-		if errors.Is(err, provider.NotFoundError{}) {
+		if _, ok := errors.AsType[provider.NotFoundError](err); ok {
 			reason = v1alpha1.ProviderNotFoundReason
 		}
 		if meta.SetStatusCondition(&obj.Status.Conditions, metav1.Condition{
```

internal/controller/evpn/fabric_controller.go
```diff
@@ -85,16 +83,33 @@ func (r *FabricReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
...
-	if _, ok := r.Provider().(provider.InterfaceProvider); !ok {
-		if meta.SetStatusCondition(&fabric.Status.Conditions, metav1.Condition{
-			Type:    v1alpha1.ReadyCondition,
-			Status:  metav1.ConditionFalse,
-			Reason:  v1alpha1.NotImplementedReason,
-			Message: "Provider does not implement provider.InterfaceProvider",
-		}) {
-			return ctrl.Result{}, r.Status().Update(ctx, fabric)
+	selector, err := metav1.LabelSelectorAsSelector(&fabric.Spec.DeviceSelector)
+	if err != nil {
```
felix-kaestner, line 87: Do we need to even do that here if the `reconcileLoopbackInterface` func already does it?

sven-rosenzweig: My intention was to fail early, if for one of the devices we do not implement the linked provider.

internal/controller/cisco/nx/bordergateway_controller.go
sven-rosenzweig: No specific reason. Fixed with next force push

internal/provider/provider.go
sven-rosenzweig: Thanks for the explanation. Will change this.

internal/controller/evpn/fabric_controller.go
sven-rosenzweig: Anyhow, I will remove it.

felix-kaestner (approved):

PR #471, opened 2026-07-30
Add OpenConfig DNS provider implementation

rgildein:
Implement EnsureDNS and DeleteDNS for the OpenConfig provider targeting SR Linux via gNMI. Maps `DNSSpec.
Domain` to `/system/dns/config/search` and `DNSSpec.Servers` to `/system/dns/servers/server` list.
Rejects unsupported fields (adminState: Down, sourceInterfaceName, vrfName) with explicit UnsupportedFieldError.

## How I tested it
```bash
$ kubectl apply -f - <<EOF
  kind: DNS
  metadata:
    name: dns
    namespace: default
  spec:
    deviceRef:
      name: leaf1
    domain: example.com
    servers:
      - address: 8.8.8.8
      - address: 1.1.1.1
EOF
dns.networking.metal.ironcore.dev/dns configured
$ gnmic -a 172.20.20.2 --port 57400 -u admin -p 'NokiaSrl1!' --skip-verify --encoding JSON_IETF get --path "openconfig-system:system/dns"
[
  {
    "source": "172.20.20.2",
    "timestamp": 1785331298950342192,
    "time": "2026-07-29T15:21:38.950342192+02:00",
    "updates": [
      {
        "Path": "openconfig-system:system/dns",
        "values": {
          "openconfig-system:system/dns": {
            "config": {
              "search": [
                "example.com"
              ]
            },
            "host-entries": {},
            "servers": {
              "server": [
                {
                  "address": "8.8.8.8",
                  "config": {
                    "address": "8.8.8.8"
                  },
                  "state": {
                    "address": "8.8.8.8"
                  }
                },
                {
                  "address": "1.1.1.1",
                  "config": {
                    "address": "1.1.1.1"
                  },
                  "state": {
                    "address": "1.1.1.1"
                  }
                }
              ]
            },
            "state": {
              "search": [
                "example.com"
              ]
            }
          }
        }
      }
    ]
  }
]
```


internal/provider/openconfig/dns_test.go
felix-kaestner, line 1: As we already validate the expected payload through the testdata file in `test/gnmi/testdata/openconfig/dns.txt`, I think we can just leave out the Payload tests here in this `internal/provider/openconfig/` package.

So generally, I would only add a `dns_test.go` if you really have some logic that you would want to test. One could e.g. add a table driven test func for the `validateDNSSpec` func, but other than that, I don't see much that would be worth a test.

rgildein: Hi @felix-kaestner, gnmi tests are not running currently, or they are? I run them manually by old Dockerfile, which was removed.
I added it, just because of this. So if it will be re-invented again latter, I'm ok to remove those `dns_test.go`.

felix-kaestner: https://github.com/ironcore-dev/network-operator/pull/409 will re-add the gnmi tests. So with this, I guess those payload tests in `dns_test.go` wouldn't be needed any longer. :)

rgildein: Nice, ok I will remove it :). Thanks

internal/provider/openconfig/testdata/dns.json
felix-kaestner, line 1: Should we also remove this file, now that it's not used anymore? The actual payload to be verified is also in `test/gnmi/testdata/openconfig/dns.txt`.

rgildein: absolutely! I forgot about that.

felix-kaestner (approved):

PR #472, opened 2026-07-30
Improve test reliability by waiting for resources to be fully deleted

adamtrizuljak-sap:
While working on https://github.com/ironcore-dev/network-operator/pull/426 I noticed that the tests are unreliable. Typical rate was 1-2 fails out of 10 runs of the test suite. Different tests would fail each time and it was hard to reproduce the failures reliably.

The root cause seems to be this. Envtest [doesn't run any controllers](https://book.kubebuilder.io/reference/envtest.html#testing-considerations), therefore the tests need to delete the created resources manually. This is already handled in the existing `AfterEach` blocks. However the deletion could happen too quickly, before the controller could finalize the resource(s). This would then trigger a cascade failure, where resources that should have been deleted are still present (or vice versa) and other related conditions.

For example in the Device tests, the added `Eventually(IsNotFound)` waits after deleting the primary resource and before deleting the Device in `AfterEach`. Without the wait, the Device could be deleted before the controller had a chance to finalize the resource — causing `GetDeviceByName` to fail, the finalizer to never be removed, and the provider state to never be cleaned up.

Also the `EthernetSegmentController` was registered in `suite_test.go` without `RequeueInterval`, meaning it would always run only 1 round of reconciliation and never get polled again.

This PR consists of the extracted test fixes from https://github.com/ironcore-dev/network-operator/pull/426 + several small additional fixes that were required to make this PR pass the tests on its own. 

Also adding the generated `cover.html` file to gitignore

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@ Dockerfile.cross
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
+cover.html
 
 # Go workspace file
 go.work
--- a/internal/controller/core/acl_controller_test.go
+++ b/internal/controller/core/acl_controller_test.go
@@ -8,6 +8,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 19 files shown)


adamtrizuljak-sap:
Extracting just the test fixes from the [Configured condition PR ](https://github.com/ironcore-dev/network-operator/pull/426)does not fully fix the tests. The fix must depend on some other code and logic changes that have been introduced in that PR. I would like to merge this as-is and keep building upon it.

felix-kaestner (approved):

PR #482, opened 2026-08-05
feat: Add skip-provisioning maintenance annotation for Device controller

nikatza:


internal/controller/core/device_controller_test.go
```diff
@@ -682,4 +602,230 @@ var _ = Describe("Device Controller", func() {
...
+				Type: corev1.SecretTypeBasicAuth,
+			}
+			Expect(k8sClient.Create(ctx, secret)).To(Succeed())
+			name = secret.Name
+			key = client.ObjectKey{Name: name, Namespace: metav1.NamespaceDefault}
+
+			testEvents = nil
+
+			DeferCleanup(func() {
+				Expect(client.IgnoreNotFound(k8sClient.Delete(ctx, secret))).To(Succeed())
```
adamtrizuljak-sap, line 630: Why `IgnoreNotFound`?  The secret should still exist

internal/controller/core/device_controller.go
```diff
@@ -429,6 +448,10 @@ func (r *DeviceReconciler) reconcileMaintenance(ctx context.Context, obj *v1alph
 	}
 
 	switch action {
+	case v1alpha1.DeviceMaintenanceSkipProvisioning:
+		// The annotation must stay present until the annotation is consumed.
```
felix-kaestner, line 452: Does this mean that when a Device is in phase==Running this annotation will stay there forever?

api/core/v1alpha1/groupversion_info.go
```diff
@@ -108,6 +108,11 @@ const (
 	// device-side operations. This is useful for recovering from terminal states (e.g., Failed)
 	// after manual intervention.
 	DeviceMaintenanceResetPhase = "reset-phase"
+	// DeviceMaintenanceSkipProvisioning is for devices provisioned out-of-band.
```
felix-kaestner, line 111:
```suggestion
	// DeviceMaintenanceSkipProvisioning is for devices provisioned outside of the operator.
```
The out-of-band could be a bit misleading w/ out-of-band connectivity.

internal/controller/core/device_controller_test.go
```diff
@@ -682,4 +602,230 @@ var _ = Describe("Device Controller", func() {
 			}).Should(Succeed())
 		})
 	})
+
+	Context("When using maintenance annotations", func() {
```
felix-kaestner, line 606: What is the reason we create a new Context here instead of just adding a new `It` case following the pattern we already have?

internal/controller/core/device_controller_test.go
nikatza: Yes, this is true. However, this is not testing the controller behavior but ensuring that the test environment is correct. The `IgnoreNotFound` makes the delete operation idempotent; meaning we ensure that "the secret doesn't exist" rather than trying to "delete an existing secret".

internal/controller/core/device_controller.go
nikatza: Indeed, I think its probably better if we remove the annotation  when Device is in Running phase and record an event. Made the changes for this.

internal/controller/core/device_controller.go
```diff
@@ -137,6 +137,14 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 			return ctrl.Result{}, reconcile.TerminalError(errors.New("provider does not support provisioning"))
 		}
 
+		if action, ok := obj.Annotations[v1alpha1.DeviceMaintenanceAnnotation]; ok && action == v1alpha1.DeviceMaintenanceSkipProvisioning {
+			// Transition to Running without removing the annotation.
+			// Annotation is removed by reconcileMaintenance on the next reconcile.
+			obj.Status.Phase = v1alpha1.DevicePhaseRunning
+			r.Recorder.Eventf(obj, nil, "Normal", "SkipProvisioning", "Maintenance", "Device in pending phase will skip provisioning due to maintenance annotation")
+			return ctrl.Result{}, nil
```
felix-kaestner, line 145: I think if we would just `break` here instead of returning, we would run the `reconcileMaintenance` to directly remove this annotation in the same reconcilation? Same applies to the `Provisioning` case.

nikatza: Currently the defer patches the object metadata first and the status second in two separate API calls. If we add the `break`, we remove the annotation (metadata) and set` phase=Running `(status) in the same reconcile. Since these are two sequential patches, the first patch's watch event shows the annotation is gone but the phase is still Pending. If the next reconcile reads from the informer cache before the second patch is processed, then we would enter normal provisioning.

swagner-de (approved):

PR #484, opened 2026-08-06
Add AGENTS.md

felix-kaestner:


AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+## Coding Philosophy
+
+- **YAGNI** — don't build it until it's needed. No scaffolding "for later"
+- **Reuse before writing** — check if the codebase, stdlib, or an existing dependency already solves it
+- **No premature abstraction** — no interfaces with one implementation, no factories for a single entities, no config for a value that never changes
+- **Deletion over addition** — remove complexity rather than add code
+- **Fewest files, shortest diff** — the minimal change that solves the problem wins
+- **Fix root causes** — fix bugs where all callers converge, not in each caller
+- **Edge-case correctness** — between two approaches of similar size, pick the one correct on edge cases
+- **Don't over-comment** — don't document trivial steps; avoid unnecessary empty lines between code segments
```
rgildein, line 184: <3

AGENTS.md
```diff
@@ -0,0 +1,215 @@
...
+## Commits
+
+- Run `git log` first and match the existing commit style
+- Plain imperative mood, no semantic prefixes (no feat:/fix:/chore:)
+- Max 72 characters per line
+
+## Development Workflow
+
+Always run against a dedicated [Kind](https://kind.sigs.k8s.io/) cluster, not a real dev/prod cluster.
+The Makefile passes `--context kind-$(KIND_CLUSTER_NAME)` to all kubectl calls automatically — no need to switch context globally.
```
rgildein: User can use any other k8s, e.g. I can spin VM with other k8s and ask Claude to run test on it. I would suggest to rephrase this to "Run agains Kind cluster only or any other user provided".

This is exactly one of reason why I'm using VM in #491, so Claude run commands in VM and do not touch view any of my context.

rgildein (commented):
Overall LGTM, only one small suggestion.

AGENTS.md
felix-kaestner: I agree that any k8s suffices (also e.g. minikube). However for the same reason I wouldn't make a suggestion to use a dedicated VM. While I do see a benefit for sandboxing coding agents in a VM, I'm not sure if we should make that recommendation within in this project. I'd rather then completely leave it up to a user/contributor on how to run k8s and how to sandbox their coding agent.

That being said, the default development environment uses Tilt, where we explicitly set some "allowed contexts" in https://github.com/ironcore-dev/network-operator/blob/main/Tiltfile.

AGENTS.md
```diff
@@ -0,0 +1,211 @@
...
+**After editing any `*.go` files:**
+
+```bash
+make fmt        # Format code
+make lint       # Lint code style
+make test       # Run unit tests (uses envtest: real K8s API + etcd)
+```
+
+Tests use **Ginkgo + Gomega** (BDD style) for controller and webhook tests — check `suite_test.go` in each package for setup.
+All other packages use the **standard library `testing` package** with table-driven tests where appropriate.
```
IvoGoman, line 46: I experienced some issues in the past that the agent was facing issues to find the kubebuilder assets. IDK if we want to tell it to use a certain ENV and prompt the user to fill that in their local dev environment? Currently the assets are only part of the build command but would still be required to execute tests

swagner-de (approved):

AGENTS.md
felix-kaestner: Updated and revered to the way local tools are installed.

IvoGoman: Does that suffice or should we tell here to point `KUBEBUILDER_ASSETS` to `setup-envtest use 1.36 -p path`?

felix-kaestner: I think I'd rather not include specific versions, as this will get outdated quick. I though the reference to the Makefile would already be sufficient for it to figure out that it should do that?

IvoGoman: The version was an example. Let's see if this is sufficient meanwhile.

PR #485, opened 2026-08-06
Implement ManagementAccess provider

rgildein:
# Description

Add EnsureManagementAccess and DeleteManagementAccess to the OpenConfig provider using openconfig-system YANG paths:
- gRPC server: /system/grpc-servers/grpc-server[name=<serverName>]/config
- SSH server:  /system/ssh-server/config

Unsupported fields (spec.grpc.gnmi, spec.ssh.sessionLimit) are rejected with a terminal UnsupportedFieldError, following the DNS provider pattern.

Adds a configurable `spec.grpc.serverName` field — defaults to `"gnmi"` on OpenConfig devices via `cmp.Or`. Raises `UnsupportedFieldError` on Cisco NX-OS where the server name is not configurable. Also introduces `DeleteManagementAccessRequest` to carry spec context (including `serverName`) into deletion.

<details>
<summary><strong>My prompt</strong></summary>

```bash
/plan Add Managementaccess to OpenConfig provider.
# context
The ManagementAccess CRD manages gRPC/gNMI and SSH server access on network devices. The NX-OS provider already have it, but OpenConfig not.
Use PR https://github.com/ironcore-dev/network-operator/pull/471 as inspiration.

# Goal
Add EnsureManagementAccess and DeleteManagementAccess to the OpenConfig provider using the standard OpenConfig YANG paths from:
 - openconfig-system-grpc (for gRPC server config)
 - openconfig-system (for SSH server config)

# Steps
1. check what parameters NetworkOperator consume for Managementaccess e.g. `ssh: timeout: 120s; sessionLimit: 10`
2. check OpenConfig paths https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-grpc-servers and https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-ssh-server
3. create EnsureManagementAccess and DeleteManagementAccess method
4. build and deploy Network-Operator
5. Use example configuretion @config/samples/v1alpha1_managementaccess.yaml to create new custom resource kind ManagementAccess
6. validate with gnmic that device was properly configured
7. @test/gnmi/testdata/openconfig/managementaccess.txt with config and expected output from gnmic
```

</details>

<details>
<summary><strong>Claude plan</strong></summary>

### Context

The `ManagementAccess` CRD manages gRPC/gNMI and SSH server access on network devices. The NX-OS provider already implements `EnsureManagementAccess` and `DeleteManagementAccess`, but the OpenConfig provider does not. The controller already checks if a provider implements `ManagementAccessProvider` (defined in `internal/provider/provider.go`) and sets a `NotImplementedReason` status when it doesn't. This task wires up the OpenConfig provider to configure devices using standard OpenConfig YANG paths from `openconfig-system-grpc` and `openconfig-system`.

---

### OpenConfig YANG Paths

**gRPC server** (named list, key: `name`):
```
openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config/enable           bool
openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config/port             uint16
openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config/certificate-id   string
openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config/network-instance string (VRF)
```

**SSH server** (singleton container):
```
openconfig-system:system/ssh-server/config/enable        bool
openconfig-system:system/ssh-server/config/timeout       uint32 (seconds)
```

---

#### New file: `internal/provider/openconfig/managementaccess.go`

**Compile-time assertion:**
```go
var _ provider.ManagementAccessProvider = (*Provider)(nil)
```

**Two structs implementing `gnmiext.DataElement`:**

1. `GRPCServer` — targets the grpc-server list item:
   - XPath: `openconfig-system:system/grpc-servers/grpc-server[name=<serverName>]/config`
   - JSON fields: `enable`, `port`, `certificate-id` (omitempty), `network-instance` (omitempty)
   - Populated from `spec.grpc.*`

2. `SSHServer` — targets the SSH server config container:
   - XPath: `openconfig-system:system/ssh-server/config`
   - JSON fields: `enable`, `timeout` (seconds, uint32)
   - Populated from `spec.ssh.*`

</details>

<details>
<summary><strong>Claude Test results</strong></summary>

```
  Test Report

  ┌──────────────────┬──────────────────┬───────────┬───────┬──────────────────────────────────────────────────────────────────────┬─────────────────┐
  │    CR Name       │      Kind        │ Namespace │ Ready │                           gNMI Path                                 │     Result      │
  ├──────────────────┼──────────────────┼───────────┼───────┼──────────────────────────────────────────────────────────────────────┼─────────────────┤
  │ managementaccess │ ManagementAccess │ default   │ True  │ openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config  │ ✓ value matches │
  │ managementaccess │ ManagementAccess │ default   │ True  │ openconfig-system:system/ssh-server/config                          │ ✓ value matches │
  └──────────────────┴──────────────────┴───────────┴───────┴──────────────────────────────────────────────────────────────────────┴─────────────────┘

  serverName tests:
  ┌──────────────────┬──────────────────┬───────────┬───────┬─────────────────────────────────────────────────────────────────────────┬─────────────────┐
  │    CR Name       │      Kind        │ Namespace │ Ready │                           gNMI Path                                    │     Result      │
  ├──────────────────┼──────────────────┼───────────┼───────┼─────────────────────────────────────────────────────────────────────────┼─────────────────┤
  │ managementaccess │ ManagementAccess │ default   │ True  │ openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config     │ ✓ value matches │
  │ managementaccess │ ManagementAccess │ default   │ True  │ openconfig-system:system/grpc-servers/grpc-server[name=my-gnmi]/config  │ ✓ value matches │
  └──────────────────┴──────────────────┴───────────┴───────┴─────────────────────────────────────────────────────────────────────────┴─────────────────┘

  Local Dev:
  ────────────────────────────────────────────────────────
  Vet:         ✓ passed
  Lint:        ✓ 0 issues
  Unit tests:  ✓ all passed
  gNMI tests:  ✓ 11/11 passed
  ────────────────────────────────────────────────────────
  Overall:     ✓ all checks passed
```

Applied spec:
```yaml
spec:
  deviceRef:
    name: leaf1
  grpc:
    certificateId: mytrustpoint
    enabled: true
    gnmi:
      keepAliveTimeout: 10m   # default — ignored by OpenConfig provider
      maxConcurrentCall: 8    # default — ignored by OpenConfig provider
    port: 9339
    vrfName: mgmt
  ssh:
    enabled: true
    sessionLimit: 32          # default — ignored by OpenConfig provider
    timeout: 120s
```

</details>

# Manual test result

```bash
$ gnmic -a 172.20.20.2 --port 57400 -u admin -p 'NokiaSrl1!' --skip-verify --encoding JSON_IETF get --path openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config
[
  {
    "source": "172.20.20.2",
    "timestamp": 1786538371750487362,
    "time": "2026-08-12T14:39:31.750487362+02:00",
    "updates": [
      {
        "Path": "openconfig-system:system/grpc-servers/grpc-server[name=gnmi]/config",
        "values": {
          "openconfig-system:system/grpc-servers/grpc-server/config": {
            "certificate-id": "mytrustpoint",
            "enable": true,
            "name": "gnmi",
            "network-instance": "mgmt",
            "port": 9339
          }
        }
      }
    ]
  }
]

$ gnmic -a 172.20.20.2 --port 57400 -u admin -p 'NokiaSrl1!' --skip-verify --encoding JSON_IETF get --path openconfig-system:system/ssh-server/config
[
  {
    "source": "172.20.20.2",
    "timestamp": 1786538371999025754,
    "time": "2026-08-12T14:39:31.999025754+02:00",
    "updates": [
      {
        "Path": "openconfig-system:system/ssh-server/config",
        "values": {
          "openconfig-system:system/ssh-server/config": {
            "enable": true,
            "timeout": 120
          }
        }
      }
    ]
  }
]
```


internal/provider/openconfig/managementaccess.go
```diff
@@ -0,0 +1,92 @@
...
+	return fmt.Sprintf("openconfig-system:system/grpc-servers/grpc-server[name=%s]/config", g.Name)
+}
+
+// SSHServer targets the OpenConfig ssh-server config container.
+type SSHServer struct {
+	Enable  bool   `json:"enable"`
+	Timeout uint32 `json:"timeout,omitempty"`
+}
+
+func (s *SSHServer) XPath() string {
```
felix-kaestner:
```suggestion
func (*SSHServer) XPath() string {
```
in such cases, we can omit the receiver variable name

internal/provider/openconfig/managementaccess.go
```diff
@@ -0,0 +1,92 @@
...
+
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/apistatus"
+	"github.com/ironcore-dev/network-operator/internal/provider"
+	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
+)
+
+var _ provider.ManagementAccessProvider = (*Provider)(nil)
+
+const grpcServerName = "gnmi"
```
felix-kaestner: I don't think this is a universal standard that is set by openconfig. As such, do we want to hardcode this name here, or should we rather allow to set this name from the spec? On a provider such as cisco nx-os which doesn't support multiple servers, this field could then perhaps be ignored?

rgildein: I was trying to follow what we have for NX-OS, where we do not set the name. My assumptions was that there is only one resource `kind: ManagementAccess` per device, so setting generic name (maybe we can go with even something like 'network-operator-gnmi'). Is that true or it's excepted to have multiple resources?

If you think it will be useful, we can do as you say add this to the spec and ignore it for Cisco provider. BTW, I think that Cisco support multiple servers, but without names.

felix-kaestner: I think we'll only have one `kind: ManagementAccess` resource per Device. However, I think some device platforms (like Nokia SRLinux) can have multiple named grpc servers. On Cisco NX-OS I'm only aware that there is a single grpc server without a name. Where did you read that Cisco NX-OS supports multiple grpc servers?

rgildein: [Here](https://www.cisco.com/c/en/us/td/docs/dcn/nx-os/nexus9000/105x/programmability/cisco-nexus-9000-series-nx-os-programmability-guide-105x/m-grpc-agent.html#guidelines-and-limitations-for-grpc-agent), it's not specifically mentioned how, but that two server can used.
Anyway, I think it makes sense to add name to the spec and raise error if it's used with Cisco provider.

hack/provider/main.go
```diff
@@ -1125,11 +1125,11 @@ func performDelete(ctx context.Context, prov provider.Provider, obj client.Objec
 			if len(refStore) == 0 {
 				return errors.New("evpninstance resource references vlan but no reference files provided (use --ref-files)")
 			}
-			obj := refStore.Get(resource.Spec.VLANRef.Name, resource.Namespace)
-			if obj == nil {
+			vlanObj := refStore.Get(resource.Spec.VLANRef.Name, resource.Namespace)
```
felix-kaestner, line 1128: Seems like a really unrelated change?

internal/provider/openconfig/managementaccess.go
```diff
@@ -0,0 +1,91 @@
...
+
+var _ provider.ManagementAccessProvider = (*Provider)(nil)
+
+func (p *Provider) EnsureManagementAccess(ctx context.Context, req *provider.EnsureManagementAccessRequest) error {
+	ma := req.ManagementAccess
+	if err := validateManagementAccessSpec(ma.Spec); err != nil {
+		return err
+	}
+	grpcServer := &GRPCServer{
+		Name:            cmp.Or(ma.Spec.GRPC.ServerName, "gnmi"),
```
felix-kaestner, line 26: Really nice trick!

api/core/v1alpha1/managementaccess_types.go
```diff
@@ -53,6 +52,14 @@ type GRPC struct {
 	// +kubebuilder:validation:ExclusiveMaximum=false
 	Port int32 `json:"port"`
 
+	// Name of the gRPC server instance on the device.
+	// If not specified, defaults to "gnmi" on OpenConfig devices.
+	// Not supported on Cisco NX-OS devices.
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	ServerName string `json:"serverName,omitempty"`
```
felix-kaestner, line 61: I think we need to make this Immutable once set.

The reasoning is that otherwise we can leave orphaned config on a device which we want to avoid. Assume you create a `ManagementAccess` resource with a grpc server and name "foo" and it creates a server named like that. Now you change the spec to have the grpc server named "bar". What will happen now is that it will again create a new grpc server, with a different name, and leave the previous gprc server running and untouched. This is because the server name is part of the YANG path, as such it identifies the tree which is updated. If it can change, then we might end up creating/updating an entirely different path.

hack/provider/main.go
rgildein: There is obj objec define above, never used below and it was siliently replaced here. Since, I need to use it I hit the issue.

api/core/v1alpha1/managementaccess_types.go
rgildein: Good point, I did not know that and did not realised that we can do that. It's not possible to do "update" for the name? If server exists we will rename it? I think it's better to be immutable, since it make more sense for user to create new resource rather that edit existing one, but I'm curious if "update" is possible.

felix-kaestner: > It's not possible to do "update" for the name? If server exists we will rename it?

So a rename in the yang model would actually be to delete the old element, e.g. `openconfig-system:system/grpc-servers/grpc-server[name=foo]`, and create a new one, e.g `openconfig-system:system/grpc-servers/grpc-server[name=bar]`. As the `name` field of the object is part of the path, so there is no way of doing something like an update to e.g. `openconfig-system:system/grpc-servers/grpc-server[name=foo]` with `{"name":"bar"}`. XPath and object field (`name` in this case) always match.

Now based on this, there is no easy way for us to know, that an existing `openconfig-system:system/grpc-servers/grpc-server[name=foo]` on the device was our operator in a previous reconcilation run. You just get the current kubernetes resource (now having `.spec.name=bar`). 

In order to actually perform the deletion of the previous element, you would need knowledge over a previous kubernetes resource version, that was successfully configured, before being changed by a user to the now current version. 😄

rgildein: ok, making it immutable is much simpler and make more sense in situations like this. Let me update it.

felix-kaestner (approved):

PR #490, opened 2026-08-10
Add Remote ConfigBackup type with S3 object storage

felix-kaestner:


docs/concepts/config-backup.md
```diff
@@ -30,7 +30,7 @@ spec:
     name: leaf-switch-1
   schedule: "0 2 * * *"
   type: Local
-  path: "<device-local-path>"
+  path: "bootflash:///backups/"
```
adamtrizuljak-sap: Is this vendor-specific? Maybe add a note about which vendor / OS this applies to. Same in `config/samples/v1alpha1_configbackup.yaml`

internal/controller/core/configbackup_controller.go
```diff
@@ -445,6 +528,166 @@ func (r *ConfigBackupReconciler) reconcile(ctx context.Context, s *configBackupS
...
+		return nil, fmt.Errorf("failed to get S3 credentials secret %s/%s: %w", ns, ref.Name, err)
+	}
+	accessKeyID, ok := secret.Data[S3AccessKeyID]
+	if !ok {
+		return nil, fmt.Errorf("secret %s/%s missing key %q", ns, ref.Name, S3AccessKeyID)
+	}
+	secretAccessKey, ok := secret.Data[S3SecretAccessKey]
+	if !ok {
+		return nil, fmt.Errorf("secret %s/%s missing key %q", ns, ref.Name, S3SecretAccessKey)
+	}
```
adamtrizuljak-sap, line 568: I think that these should also return TerminalError and set the condition to False - missing secret key is a user error. Instead of retrying this with exponential backoff, we should wait until the user deploys the secret with the correct structure, upon which the watch re-enqueues the controller.

docs/concepts/config-backup.md
felix-kaestner: Added a comment to both.

internal/controller/core/configbackup_controller.go
felix-kaestner: Agree. Changed to a terminal error.

adamtrizuljak-sap (approved):
LGTM, but someone else could have a look as well :)

PR #492, opened 2026-08-11
Implement OpenConfig provider for core resource types

felix-kaestner:
Add provider implementations for ACL, AAA, BGP, BGPPeer, ISIS, LLDP,
NTP, PrefixSet, RoutingPolicy, Syslog, and VRF resources targeting
Nokia SRLinux via gNMI.

Use SetBuilder for multi-element operations and direct client calls
for single-element operations. Guard unsupported fields with
UnsupportedFieldError.

### DIFF
--- a/internal/provider/openconfig/aaa.go
+++ b/internal/provider/openconfig/aaa.go
@@ -1,0 +1,222 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package openconfig
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/apistatus"
+	"github.com/ironcore-dev/network-operator/internal/provider"
+	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
+)
+
+var _ provider.AAAProvider = (*Provider)(nil)
+
+func (p *Provider) EnsureAAA(ctx context.Context, req *provider.EnsureAAARequest) error {
+	spec := req.AAA.Spec
+
+	if spec.Authorization != nil {
+		return apistatus.NewUnsupportedFieldError(apistatus.FieldViolation{
+			Field:       "spec.authorization",
+			Description: "openconfig provider does not support AAA authorization on SRLinux",
+		})
+	}
+
+	sb := new(gnmiext.SetBuilder)
+
+	for _, sg := range spec.ServerGroups {
+		if sg.Type == v1alpha1.AAAServerGroupTypeRADIUS {
+			return apistatus.NewUnsupportedFieldError(apistatus.FieldViolation{
+				Field:       "spec.serverGroups[].type",
+				Description: "openconfig provider does not support RADIUS server groups on SRLinux",
+			})
+		}
+
+		group := &AAAServerGroup{
+			Name: sg.Name,
+			Config: &AAAServerGroupConfig{
+				Name: sg.Name,
+				Type: AAAServerGroupTypeTACACS,
+			},
+			Servers: &AAAServers{},
+		}
+
+		for _, srv := range sg.Servers {
+			s := &AAAServer{
+				Address: srv.Address,
+				Config: &AAAServerConfig{
+					Address: srv.Address,
+				},
+			}
+			if srv.Timeout != nil {
+				s.Config.Timeout = uint16(srv.Timeout.Seconds())
+			}
+			if srv.TACACS != nil {
+				key := req.TACACSServerKeys[srv.Address]
+				s.TACACS = &AAAServerTACACS{
+					Config: &AAAServerTACACSConfig{
+						Port:      uint16(srv.TACACS.Port), //nolint:gosec
+						SecretKey: key,
+					},
+				}
+			}
+			group.Servers.Server.Set(s)
+		}
+		sb.Update(group)
+	}
+
+	if spec.Authentication != nil {
+		methods := make([]AAAMethodType, 0, len(spec.Authentication.Methods))
+		for _, m := range spec.Authentication.Methods {
+			methods = append(methods, toAAAMethod(m))
+		}
+		sb.Update(&AAAAuthenticationConfig{Methods: methods})
+	}
+
+	if spec.Accounting != nil {
+		methods := make([]AAAMethodType, 0, len(spec.Accounting.Methods))
+		for _, m := range spec.Accounting.Methods {
+			methods = append(methods, toAAAMethod(m))
+		}
+		sb.Update(&AAAAccountingConfig{Methods: methods})
+	}
+
+	return p.client.Do(ctx, sb)
+}
+
+func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest) error {
+	sb := new(gnmiext.SetBuilder)
+	for _, sg := range req.AAA.Spec.ServerGroups {
+		sb.Delete(&AAAServerGroup{Name: sg.Name})
+	}
+	if len(req.AAA.Spec.ServerGroups) == 0 {
+		sb.Delete(&AAAContainer{})
+	}
+	return p.client.Do(ctx, sb)
+}
+
+func toAAAMethod(m v1alpha1.AAAMethod) AAAMethodType {
+	switch m.Type {
+	case v1alpha1.AAAMethodTypeLocal:
+		return AAAMethodTypeLocal
+	case v1alpha1.AAAMethodTypeNone:
+		return AAAMethodTypeNone
+	case v1alpha1.AAAMethodTypeGroup:
+		return AAAMethodType(m.GroupName)
+	default:
+		return AAAMethodTypeLocal
+	}
+}
+
+// AAAServerGroupType represents the OpenConfig AAA server group type identity.
+type AAAServerGroupType string
+
+const (
+	AAAServerGroupTypeTACACS AAAServerGroupType = "openconfig-aaa:TACACS"
+)
+
+// AAAMethodType represents the AAA authentication/accounting method string.
+type AAAMethodType string
+
+const (
+	AAAMethodTypeLocal AAAMethodType = "local"
+	AAAMethodTypeNone  AAAMethodType = "none"
+)
+
+// Compile-time assertions.
+var (
+	_ gnmiext.DataElement = (*AAAServerGroup)(nil)
+	_ gnmiext.DataElement = (*AAAAuthenticationConfig)(nil)
+	_ gnmiext.DataElement = (*AAAAccountingConfig)(nil)
+	_ gnmiext.DataElement = (*AAAContainer)(nil)
+)
+
+// AAAContainer targets the full AAA container for deletion.
+type AAAContainer struct{}
+
+func (*AAAContainer) XPath() string { return "openconfig-system:system/aaa" }
+
+// AAAServerGroup targets a server-group entry.
+type AAAServerGroup struct {
+	Name    string                `json:"-"`
+	Config  *AAAServerGroupConfig `json:"config,omitempty"`
+	Servers *AAAServers           `json:"servers,omitempty"`
+}
+
+func (g *AAAServerGroup) XPath() string {
+	return fmt.Sprintf("openconfig-system:system/aaa/server-groups/server-group[name=%s]", g.Name)
+}
+
+// AAAServerGroupConfig holds the server-group config.
+type AAAServerGroupConfig struct {
+	Name string             `json:"name"`
+	Type AAAServerGroupType `json:"type"`
+}
+
+// AAAServers holds the server list.
+type AAAServers struct {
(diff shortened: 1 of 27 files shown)


rgildein:
## Test Report — OpenConfig Provider (Nokia SRL 26.7.1)

### Local Dev Checks

```
────────────────────────────────────────────────────────
Vet:         ✓ passed
Lint:        ✓ passed  (0 issues)
Unit tests:  ✓ 21 packages passed, 0 failed
gNMI tests:  ✓ 17 passed, 0 failed
             banner, bgp, dns, interface (×8), lldp,
             ntp, prefixset, routingpolicy, syslog, vrf
────────────────────────────────────────────────────────
```

---

### Real Device — Nokia SRL 26.7.1 @ 172.20.20.2:57400

| CR Name | Kind | Ready | gNMI Path | Result |
|---|---|---|---|---|
| lldp | LLDP | True | `openconfig-lldp:lldp/config` | ✓ enabled: true |
| pfx-loopbacks | PrefixSet | True | `...prefix-set[name=LOOPBACKS]` | ✓ 10.0.0.0/24, 10.1.0.0/16 |
| accept-loopbacks | RoutingPolicy | True | `...policy-definition[name=accept-loopbacks]` | ✓ statements 10 + 65535 |
| vrf-blue | VRF | True | `...network-instance[name=BLUE]/config` | ✓ type: L3VRF |
| vrf-default | VRF | True | `...network-instance[name=default]/config` | ✓ type: DEFAULT_INSTANCE |
| bgp | BGP | True | `...network-instance[name=default]/...bgp/global` | ✓ AS 65000, IPv4+IPv6 |
| ntp | NTP | True | `openconfig-system:system/ntp` | ✓ 2 servers, prefer + default |
| syslog | Syslog | True | `openconfig-system:system/logging/remote-servers` | ✓ INFORMATIONAL severity |
| dns | DNS | False | `openconfig-system:system/dns` | ✗ containerlab constraint (see below) |

**8/9 passed**

---

### gnmic outputs

<details>
<summary><code>gnmic get openconfig-lldp:lldp/config</code></summary>

```json
{ "enabled": true }
```
</details>

<details>
<summary><code>gnmic get ...prefix-set[name=LOOPBACKS]</code></summary>

```json
{
  "config": { "name": "LOOPBACKS" },
  "prefixes": {
    "prefix": [
      { "ip-prefix": "10.0.0.0/24", "masklength-range": "exact",
        "config": { "ip-prefix": "10.0.0.0/24", "masklength-range": "exact" } },
      { "ip-prefix": "10.1.0.0/16", "masklength-range": "exact",
        "config": { "ip-prefix": "10.1.0.0/16", "masklength-range": "exact" } }
    ]
  }
}
```
</details>

<details>
<summary><code>gnmic get ...policy-definition[name=accept-loopbacks]</code></summary>

```json
{
  "config": { "name": "accept-loopbacks" },
  "statements": {
    "statement": [
      { "name": "10",
        "conditions": { "match-prefix-set": { "config": { "prefix-set": "LOOPBACKS", "match-set-options": "ANY" } } },
        "actions": { "config": { "policy-result": "ACCEPT_ROUTE" } } },
      { "name": "65535",
        "actions": { "config": { "policy-result": "REJECT_ROUTE" } } }
    ]
  }
}
```
</details>

<details>
<summary><code>gnmic get ...network-instance[name=BLUE]/config</code></summary>

```json
{ "name": "BLUE", "type": "openconfig-network-instance-types:L3VRF" }
```
</details>

<details>
<summary><code>gnmic get ...network-instance[name=default]/config</code></summary>

```json
{ "name": "default", "type": "openconfig-network-instance-types:DEFAULT_INSTANCE" }
```
</details>

<details>
<summary><code>gnmic get ...network-instance[name=default]/protocols/protocol[BGP]/bgp/global</code></summary>

```json
{
  "config": { "as": 65000, "router-id": "10.0.0.1" },
  "afi-safis": {
    "afi-safi": [
      { "afi-safi-name": "openconfig-bgp-types:IPV4_UNICAST",
        "config": { "afi-safi-name": "openconfig-bgp-types:IPV4_UNICAST", "enabled": true } },
      { "afi-safi-name": "openconfig-bgp-types:IPV6_UNICAST",
        "config": { "afi-safi-name": "openconfig-bgp-types:IPV6_UNICAST", "enabled": true } }
    ]
  }
}
```
</details>

<details>
<summary><code>gnmic get openconfig-system:system/ntp</code></summary>

```json
{
  "config": { "enabled": true },
  "servers": {
    "server": [
      { "address": "192.0.2.1", "config": { "address": "192.0.2.1", "network-instance": "default", "prefer": true } },
      { "address": "192.0.2.2", "config": { "address": "192.0.2.2", "network-instance": "default" } }
    ]
  }
}
```
</details>

<details>
<summary><code>gnmic get openconfig-system:system/logging/remote-servers</code></summary>

```json
{
  "remote-server": [
    {
      "config": { "host": "192.0.2.100", "remote-port": 514 },
      "host": "192.0.2.100",
      "selectors": {
        "selector": [
          { "facility": "openconfig-system-logging:ALL", "severity": "INFORMATIONAL",
            "config": { "facility": "openconfig-system-logging:ALL", "severity": "INFORMATIONAL" } }
        ]
      }
    }
  ]
}
```
</details>

---

### Notes

**BGP** requires a `DEFAULT_INSTANCE` VRF CR to be applied first — SRLinux enforces that the `default` network-instance must exist with `type` set before a BGP protocol entry can be written.

**DNS** — containerlab pre-configures `mgmt` and `clab-default` DNS network-instances at boot. SRLinux enforces a unique `network-instance` leaf on the OpenConfig `system/dns` path, causing `FailedPrecondition` on any write. Passes cleanly on the mock server. Not a provider bug — a containerlab topology constraint that does not apply in production.

### Additional fixes in this commit

- `NTPSpec.SourceInterfaceName`, `SyslogSpec.Facilities`, `LogServer.VrfName` made optional in the API (were incorrectly required, blocking Nokia SRLinux usage)
- `NTP.servers[].vrfName` was silently accepted by the provider — now correctly returns `UnsupportedFieldError`
- Syslog severity values mapped from API strings (`Info`) to OpenConfig identities (`INFORMATIONAL`)

---

Signed-off-by: Claude <claude@anthropic.com>

internal/provider/openconfig/aaa.go
```diff
@@ -0,0 +1,223 @@
...
+		sb.Update(&AAAAccountingConfig{Methods: methods})
+	}
+
+	// Authorization — not supported on SRLinux OC (empty config options).
+	if spec.Authorization != nil {
+		return apistatus.NewUnsupportedFieldError(apistatus.FieldViolation{
+			Field:       "spec.authorization",
+			Description: "openconfig provider does not support AAA authorization on SRLinux",
+		})
+	}
```
rgildein: Would not make sense to put this on top of EnsureAAA function? -> failed fast

internal/provider/openconfig/aaa.go
```diff
@@ -0,0 +1,223 @@
...
+func (p *Provider) EnsureAAA(ctx context.Context, req *provider.EnsureAAARequest) error {
+	spec := req.AAA.Spec
+
+	sb := new(gnmiext.SetBuilder)
+
+	for _, sg := range spec.ServerGroups {
+		if sg.Type == v1alpha1.AAAServerGroupTypeRADIUS {
+			return apistatus.NewUnsupportedFieldError(apistatus.FieldViolation{
+				Field:       "spec.serverGroups[].type",
+				Description: "openconfig provider does not support RADIUS server groups on SRLinux",
```
rgildein, line 35: You are using "on SRLinux" in all messages, but OpenConfig could be used with multiple vendor. Could we remove it and add some note that, currently it's tested only for Nokia devices?

internal/provider/openconfig/acl.go
```diff
@@ -0,0 +1,213 @@
...
+var _ provider.ACLProvider = (*Provider)(nil)
+
+func (p *Provider) EnsureACL(ctx context.Context, req *provider.EnsureACLRequest) error {
+	spec := req.ACL.Spec
+
+	// Determine ACL type based on first entry's source address.
+	aclType := ACLTypeIPv4
+	if len(spec.Entries) > 0 {
+		if spec.Entries[0].SourceAddress.Is6() {
+			aclType = ACLTypeIPv6
```
rgildein: I do not think this is correct approach,  what if user used mixed types? Should we do the check for mixed address types here or via kubebuilder validation?

internal/provider/openconfig/acl.go
```diff
@@ -0,0 +1,213 @@
...
+
+		acl.Entries.Entry.Set(e)
+	}
+
+	return p.client.Update(ctx, acl)
+}
+
+func (p *Provider) DeleteACL(ctx context.Context, req *provider.DeleteACLRequest) error {
+	// Delete both possible types — only one will exist.
+	aclV4 := &ACLSet{Name: req.Name, ACLType: ACLTypeIPv4}
```
rgildein: Would it not make sense to check type also here, so we will not delete something what was not configured by the operator?

internal/provider/openconfig/aaa.go
rgildein: Actually reading below, I see that not all massages have "on SRLinux", so maybe it's correct approach.
Do you have any idea how will handle it if we will want to support multiple vendor with OpenConfig? For me it would be still better to tell that OpenConfig provider is now tested and supported only for Nokia devices and handle any further changes, when other device will be supported. WDYT?

internal/provider/openconfig/bgp.go
```diff
@@ -0,0 +1,347 @@
...
+	if req.BGP.Spec.AdminState == v1alpha1.AdminStateDown {
+		return apistatus.NewUnsupportedFieldError(apistatus.FieldViolation{
+			Field:       "spec.adminState",
+			Description: "openconfig provider does not support disabling BGP (adminState Down)",
+		})
+	}
+
+	spec := req.BGP.Spec
+
+	asn, err := asnToUint32(spec.ASNumber)
```
rgildein, line 35: Not related to this changes, but the spec.ASNumber should not be define as `string` instead of `IntOrString`, since there are `-` and `.`? To me the IntOrString represent `"1" ` or `1`. We can add some regex validation for string and with that the asnToUint32 would be much simpler.

internal/provider/openconfig/aaa.go
felix-kaestner: Yes, agree. We should document that the openconfig implementation is currently only tested targeting Nokia SRLinux. Once we want to add support for other device platforms for openconfig, we need to re-evaluate.

internal/provider/openconfig/acl.go
felix-kaestner: Agreed. This should be API validation rather than provider-only validation. Still added a provider-side guard for now such that `EnsureACL` rejects entries where source and destination use different IP families. Longer term I'd handle mixed-family ACL validation in a follow-up API change.

internal/provider/openconfig/aaa.go
felix-kaestner: I would keep the SRLinux-specific wording where the limitation was observed on SRLinux. Otherwise the error would imply a generic OpenConfig limitation, which may not be true for another vendor.

If we add another OpenConfig target later, we can introduce vendor/capability-specific handling where behavior differs.

internal/provider/openconfig/acl.go
rgildein: Can you create separate issue for it and link it here? thx

internal/provider/openconfig/bgp.go
felix-kaestner: The spec.ASNumber can carry both plain formatted and dotted notation as numbers (ref/ https://datatracker.ietf.org/doc/html/rfc5396). However the openconfig model represents them as a uint32 so we have the normalize the value to a uint32. Let's add validation to this spec field in a follow-up. However, I'm unsure how this would simplify the `asnToUint32` func. Can you elaborate?

internal/provider/openconfig/acl.go
felix-kaestner: My previous assumption was that a ACL with a given name can only exist once so there could never be a ipv4 and ipv6 acl with the same name at the same time. However, this might be a platform limitation and is not part of the openconfig model. So I've change the implementation here to specifically only delete the ACL for the given type.

internal/provider/openconfig/aaa.go
felix-kaestner: Fixed by moving the unsupported spec.authorization check to the top of EnsureAAA, before building any gNMI operations

internal/provider/openconfig/bgp.go
felix-kaestner: Created https://github.com/ironcore-dev/network-operator/issues/507 as a follow-up.

internal/provider/openconfig/acl.go
felix-kaestner: Done. https://github.com/ironcore-dev/network-operator/issues/506

internal/provider/openconfig/bgp.go
rgildein: Reading the rfc5396 and I realised that I understand the format wrongly. Now it make sense to me, that it's we are using `IntOrString` type :+1:. Adding validation to the scheme would improve it, but not replace usage of `IntOrString`. Resolving.

rgildein (approved):
LGTM

PR #510, opened 2026-08-18
Add Probe CRD and controller for fabric health assertions

felix-kaestner:
Introduce the Probe resource in the core API group for asserting operational health of network devices. A Probe runs one-shot or on a cron schedule and supports four assertion types:

- Ping: ICMP echo from device with optional VRF/source interface
- MACTableEntry: assert MAC exists in forwarding table
- RoutePresence: assert IP prefix exists in routing table
- VTEPPeerConnectivity: assert VTEP peers are up

Add ProbeProvider interface with per-type methods, controller with full reconcile loop (schedule, lock, connect, dispatch, status), reference resolution for Interface/VRF/VLAN with not-found, cross-device, and readiness checks. Add watchers for Device, Interface, VRF, and VLAN ready-state changes.

Add VLANSource and InterfaceSource/VRFSource shared types for name-or-ref resolution with CEL mutual exclusivity validation.

```bash
$ kubectl get netprobe
NAME           DEVICE   TYPE                   READY   AGE
mac-entry      leaf1    MACTableEntry          True    4m39s
ping-peer      leaf1    Ping                   True    12m
route-prefix   leaf1    RoutePresence          True    2m47s
vtep-peers     leaf1    VTEPPeerConnectivity   True    49s
```


felix-kaestner:
**Ping Probe**

```yaml
kind: Probe
spec:
  deviceRef:
    name: leaf1
  ping:
    address: 10.0.0.12
    count: 3
    packetSize: 1400
    sourceInterface:
      interfaceRef:
        name: lo0
    timeout: 5s
    vrf:
      name: default
  schedule: '* * * * *'
  type: Ping
status:
  conditions:
  - lastTransitionTime: "2026-08-20T14:48:01Z"
    message: 3/3 packets received, avg 9.66ms
    observedGeneration: 1
    reason: ProbeSuccessful
    status: "True"
    type: Ready
  lastRunTime: "2026-08-20T14:49:00Z"
  nextRunTime: "2026-08-20T14:50:00Z"
  ping:
    avgTime: 9.66ms
    maxTime: 10.723ms
    minTime: 8.319999ms
    received: 3
    sent: 3
```

**MAC Table Entry Probe**

```yaml
kind: Probe
spec:
  deviceRef:
    name: leaf1
  macTableEntry:
    macAddress: "00:00:00:00:00:02"
    vlan:
      id: 10
  type: MACTableEntry
status:
  conditions:
  - lastTransitionTime: "2026-08-20T14:55:00Z"
    message: MAC 00:00:00:00:00:02 found
    observedGeneration: 1
    reason: ProbeSuccessful
    status: "True"
    type: Ready
  lastRunTime: "2026-08-20T14:55:00Z"
```

**Route Prefix Route Probe**

```yaml
﻿apiVersion: networking.metal.ironcore.dev/v1alpha1
kind: Probe
spec:
  deviceRef:
    name: leaf1
  routePresence:
    prefix: 192.168.10.0/24
    vrf:
      name: default
  type: RoutePresence
status:
  conditions:
  - lastTransitionTime: "2026-08-20T14:56:52Z"
    message: prefix 192.168.10.0/24 found
    observedGeneration: 1
    reason: ProbeSuccessful
    status: "True"
    type: Ready
  lastRunTime: "2026-08-20T14:56:52Z"
```

**VTEP Connectivity Probe**

```yaml
kind: Probe
spec:
  deviceRef:
    name: leaf1
  type: VTEPPeerConnectivity
  vtepPeerConnectivity:
    expectedPeers:
    - 10.0.1.12
status:
  conditions:
  - lastTransitionTime: "2026-08-20T14:58:45Z"
    message: 1/1 expected VTEP peers up
    observedGeneration: 1
    reason: ProbeSuccessful
    status: "True"
    type: Ready
  lastRunTime: "2026-08-20T14:58:45Z"
```

api/core/v1alpha1/probe_types.go
```diff
@@ -0,0 +1,299 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceRef is immutable"
+	DeviceRef LocalObjectReference `json:"deviceRef"`
+
+	// ProviderConfigRef is a reference to a resource holding the provider-specific configuration of this probe.
+	// This reference is used to link the Probe to its provider-specific configuration.
+	// +optional
+	ProviderConfigRef *TypedLocalObjectReference `json:"providerConfigRef,omitempty"`
+
+	// Schedule is an optional cron expression (e.g., "*/5 * * * *").
+	// If omitted, the controller performs a one-shot probe execution.
```
nikatza: Without looking at the implementation it is not clear if this is one-shoot, or one-shoot per reconciliation.

internal/controller/core/probe_controller.go
```diff
@@ -0,0 +1,861 @@
...
+		}
+		return ctrl.Result{}, nil
+	}
+
+	device, err := deviceutil.GetDeviceByName(ctx, r, obj.Namespace, obj.Spec.DeviceRef.Name)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	if isPaused, requeue, err := paused.EnsureCondition(ctx, r.Client, device, obj); isPaused || requeue || err != nil {
```
nikatza: May be the probes should not be affected by the device condition: we might want to run them even when the device is `paused`, for example for debugging. I don't think this is a blocker per se.

internal/controller/core/probe_controller.go
```diff
@@ -0,0 +1,861 @@
...
+			handler.EnqueueRequestsFromMapFunc(r.deviceToProbes),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					return paused.DevicePausedChanged(e.ObjectOld, e.ObjectNew)
+				},
+				GenericFunc: func(e event.GenericEvent) bool {
+					return false
+				},
+			}),
+		).
```
nikatza, line 252: Related to my comment earlier: we might want to ignore the device condition for probe resources / have a different handling approach.

internal/controller/core/probe_controller.go
```diff
@@ -0,0 +1,861 @@
...
+		// Watches enqueues Probes when a referenced Interface's ready state changes.
+		Watches(
+			&v1alpha1.Interface{},
+			handler.EnqueueRequestsFromMapFunc(r.interfaceToProbes),
+			builder.WithPredicates(predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					oldIntf := e.ObjectOld.(*v1alpha1.Interface)
+					newIntf := e.ObjectNew.(*v1alpha1.Interface)
+					return conditions.IsReady(oldIntf) != conditions.IsReady(newIntf)
```
nikatza, line 261: Should this check for `configured` instead of `ready` and trigger a probe even if the interface is not ready?

internal/controller/core/probe_controller.go
```diff
@@ -0,0 +1,861 @@
...
+			s.Probe.Status.Ping.MinTime = &metav1.Duration{Duration: stats.MinTime}
+		}
+		if stats.AvgTime > 0 {
+			s.Probe.Status.Ping.AvgTime = &metav1.Duration{Duration: stats.AvgTime}
+		}
+		if stats.MaxTime > 0 {
+			s.Probe.Status.Ping.MaxTime = &metav1.Duration{Duration: stats.MaxTime}
+		}
+
+		if stats.Received == 0 {
```
nikatza: we could fail the assertion if there is flapping somewhere with 
```suggestion
		if stats.Received != stats.Sent {
```

internal/controller/core/probe_controller.go
```diff
@@ -0,0 +1,861 @@
...
+			if err != nil {
+				return "", err
+			}
+		}
+		routes, err := s.Provider.GetRouteTable(ctx, req)
+		if err != nil {
+			return "", err
+		}
+		targetPrefix := spec.RoutePresence.Prefix.String()
+		if !slices.ContainsFunc(routes, func(route provider.RouteEntry) bool {
```
nikatza, line 528: This performs an exact match lookup. I am wondering if this should be a longest prefix match instead and change/extend the API to support it. Any thoughts?

internal/provider/cisco/nxos/provider.go
```diff
@@ -4009,6 +4011,261 @@ func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest
...
+			return p.InterfaceIPAddr(ctx, addr4.Unnumbered, vrf)
+		}
+		for _, a := range addr4.AddrItems.AddrList {
+			if a.Type == IntfAddrTypePrimary {
+				ip, _, _ := strings.Cut(a.Addr, "/")
+				return ip, nil
+			}
+		}
+	}
+	addr6 := &AddrItem{ID: short, Vrf: vrf, Is6: true}
```
nikatza: Here we fetch the IPv4 addr if it exists and the IPv6 only if the former doesn't. Howerer,  we could set an ICMP probe could be targettng an IPv6 address on a dual-stack interface. I am aware that this is not yet supported, but we could still pass an AFI param to this function or return a slice with all addresses.

internal/provider/cisco/nxos/provider.go
```diff
@@ -4009,6 +4011,261 @@ func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest
...
+	}
+	if vrf == "" {
+		vrf = DefaultVRFName
+	}
+	addr4 := &AddrItem{ID: short, Vrf: vrf}
+	if err := p.client.GetConfig(ctx, addr4); err != nil && !errors.Is(err, gnmiext.ErrNil) {
+		return "", fmt.Errorf("failed to get IPv4 address for interface %q in VRF %q: %w", name, vrf, err)
+	}
+	if len(addr4.AddrItems.AddrList) > 0 {
+		if addr4.Unnumbered != "" {
```
nikatza: This is an interesting quirk. So, in the model we can have have both things configured but then `unnumbered` takes precedence? Could you please add a comment to the function that explains this? Thanks :)

internal/provider/cisco/nxos/provider.go
```diff
@@ -4009,6 +4011,261 @@ func (p *Provider) DeleteAAA(ctx context.Context, req *provider.DeleteAAARequest
...
+	var res *systempb.PingResponse
+	for {
+		resp, err := stream.Recv()
+		if err != nil {
+			break
+		}
+		res = resp
+	}
+	if res == nil {
+		return nil, fmt.Errorf("ping to %s returned no response", req.Address)
```
nikatza, line 4143: Would it help here to use of `grpc.WrapTerminalError`?  Or at least propagate the error code to the controller for logging?

nikatza (commented):
I left some comments but this is a really cool contribution!

api/core/v1alpha1/probe_types.go
felix-kaestner: Clarified.

internal/controller/core/probe_controller.go
felix-kaestner: Agree. However, I added a check for checking if the device is reachable and not even trying when the device is not reachable.

internal/controller/core/probe_controller.go
felix-kaestner: Also changed to only retrigger the probe when the device become unreachable/reachable.

internal/controller/core/probe_controller.go
felix-kaestner: I would actually think Ready is what we want here. As the outcome of the probe will only change if the interface is operational, right?

internal/controller/core/probe_controller.go
felix-kaestner: Fixed.

internal/provider/cisco/nxos/provider.go
felix-kaestner: The `AddrItem` will either represent a unnumbered reference or explicit ip assignment. Never both. So the fields inside the struct are mutually exclusive. Hence we can just check if the `Unnumbered` field is set, in which case we don't evaluate `AddrItems` in the struct.

internal/provider/cisco/nxos/provider.go
felix-kaestner: Changed to return the `error` in case it's non-nil. We should always have either a `res != nil` or `error != nil`. So this case of `res == nil` should then actually never happen. If it still, I think it's fine that we treat this as a normal error and retry (w/ exponential backoff).

internal/controller/core/probe_controller.go
felix-kaestner: I think exact matching is the right default for RoutePresence: it asserts that the specified route prefix is installed. LPM would need different and explicit semantics (probably a destination-address probe) and can be added later if there is a concrete use case. Or do you see a need to add this straight away?

internal/provider/cisco/nxos/provider.go
felix-kaestner: Changed to use the AFI of the destination IP Address and also documented that behavior on the api.

nikatza (approved):

PR #513, opened 2026-08-20
Add OpenConfig user provider

rgildein:
Add OpenConfig provider implementation for the `User` resource targeting the standard OpenConfig path:
`openconfig-system:system/aaa/authentication/users/user[username=X]/config`

## What changed

**Initial implementation:**
- `EnsureUser` / `DeleteUser` for the OpenConfig provider
- Supports `username`, `role`, and `ssh-key` — standard OpenConfig user config leaves
- Uses gNMI update (`Patch`) instead of replace — Juniper rejects gNMI replace for new list entries with `statement not found`; update works for both create and update
- `spec.roles` with more than one entry returns `UnsupportedFieldError` — the OpenConfig `role` leaf is a single value, confirmed on both Juniper and Nokia

**Password hashing:**
- Plaintext password from the Kubernetes Secret is hashed with SHA-512 crypt (`$6$`) before being sent to the device via the `password-hashed` gNMI field
- `password-hashed` is stored verbatim by Juniper, enabling idempotent reconciliation: the existing hash is read back, its salt extracted, and the password re-hashed — if the password is unchanged the output is identical and no gNMI Set is issued
- If the password changes, or the user does not exist yet, a fresh hash with a random salt is generated

## Testing

Validated against a real Juniper vJunos-Evolved device:
- User created with `username`, `password`, `role`, and `ssh-key`
- Hash stable across multiple reconcile cycles (no gNMI Set after initial write)
- Password update triggers a new hash and a gNMI Set

<details><summary><strong>Juniper vJunos-Evolved</strong></summary>

**Device:** Juniper vJunos-Evolved 26.2R1.7 via containerlab

**Verify user config:**
```bash
gnmic -a <device-address>:57401 -u admin -p 'admin@123' --skip-verify \
  --encoding JSON_IETF --type CONFIG get \
  --path 'openconfig-system:system/aaa/authentication/users/user[username=vjunos]'
```

</details>

<details><summary><strong>Nokia SR Linux</strong></summary>

Nokia SRL's OpenConfig user model does not expose `password` or `ssh-key` as writable leaves — only `username` and `role` are supported via OpenConfig. A separate Nokia-specific provider will be needed for password and SSH key management.

</details>


internal/apistatus/apistatus.go
```diff
@@ -34,12 +34,18 @@ const (
 	// configured out-of-band). [WrapTerminalError] does not promote these
 	// errors to terminal.
 	CodeFailedPrecondition
+
+	// CodeIgnoredField signals that one or more spec fields were silently
+	// ignored during realization because the provider does not support them.
+	// The resource is still considered successfully configured. [WrapTerminalError]
+	// does not promote these errors to terminal.
+	CodeIgnoredField
```
felix-kaestner: Honestly, this status code doesn't seem to follow the same semantics as all others do. Other status codes result in an `error` being returned (while this status code then returns `nil` from a reconcile) and also set the status condition to `false` (while this status code keeps it as `true`). As such, I would rather avoid having it.

rgildein: Agree, I removed it. As we agreed to not bend OpenConfig provider to work with Nokia device.

internal/provider/openconfig/user.go
```diff
@@ -0,0 +1,78 @@
...
+	Config   *UserConfig `json:"config,omitempty"`
+}
+
+func (u *User) XPath() string {
+	return fmt.Sprintf("openconfig-system:system/aaa/authentication/users/user[username=%s]", u.Username)
+}
+
+// UserConfig holds the user config container leaves.
+// Password is write-only — the device returns a hashed value that would never match
+// the plaintext, so we exclude it from unmarshal to avoid perpetual diffs.
```
felix-kaestner: Please have a look into what we are doing on the nxos provider in https://github.com/ironcore-dev/network-operator/blob/main/internal/provider/cisco/nxos/provider.go#L2620-L2641

We take the plaintext password as we retrieve it from the kubernetes secret and compute the hash ourselves (which we can do if the hash includes the algorithm and salt value). If the hash we compute from the plaintext value matches, what is stored in hashed form on the device. We can retain that value.

Otherwise, we would end up with a gnmi write on every reconcilation, which we definitely want to avoid.

internal/provider/openconfig/user.go
```diff
@@ -0,0 +1,78 @@
...
+	return p.client.Delete(ctx, &User{Username: req.Username})
+}
+
+// Compile-time assertion.
+var _ gnmiext.DataElement = (*User)(nil)
+
+// User targets an OpenConfig user entry.
+type User struct {
+	Username string      `json:"-"`
+	Config   *UserConfig `json:"config,omitempty"`
```
felix-kaestner:
```suggestion
	Config   *UserConfig `json:"config"`
```
This field is always present and should therefore not have an `omitempty` tag. See https://github.com/ironcore-dev/network-operator/blob/e21328df4f578010e609e15d2477946f6187d5a6/AGENTS.md?plain=1#L156-L161

internal/provider/openconfig/user.go
```diff
@@ -0,0 +1,78 @@
...
+	u := &User{
+		Username: req.Username,
+		Config: &UserConfig{
+			Username: req.Username,
+			Role:     req.Roles[0],
+			Password: req.Password,
+			SSHKey:   req.SSHKey,
+		},
+	}
+	return p.client.Patch(ctx, u)
```
felix-kaestner, line 44: Any particular reason we use a patch over an update here?

internal/provider/openconfig/user.go
rgildein: Ok so if user set up password we hashed and set hashed password on device. Make sense, I'll check the mechanism how we are hashing it.

internal/provider/openconfig/user.go
rgildein: Update was failing on Juniper device, I think it was to avoid changing username <==> creating new user. I can try out with update and share the message if you want.

internal/provider/openconfig/user.go
```diff
@@ -44,35 +87,19 @@ var _ gnmiext.DataElement = (*User)(nil)
...
-		return err
-	}
-	c.Username = a.Username
-	c.Role = a.Role
-	c.SSHKey = a.SSHKey
-	return nil
+	Username       string `json:"username"`
+	Role           string `json:"role,omitempty"`
+	PasswordHashed string `json:"password-hashed,omitempty"`
+	SSHKey         string `json:"ssh-key,omitempty"`
```
felix-kaestner, line 104: I don't think this field should have an "omitempty" tag, as it's always set. ref/ https://github.com/ironcore-dev/network-operator/blob/main/AGENTS.md?plain=1#L157-L164.

internal/provider/openconfig/user_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/go-crypt/crypt/algorithm/shacrypt"
```
felix-kaestner:
```suggestion
	"google.golang.org/grpc/status"
	"github.com/go-crypt/crypt/algorithm/shacrypt"
```
nit: should be a single third-party import block

felix-kaestner:
@rgildein when merging this PR, could you make sure to squash the commits into a single one so we keep the history linear and clean? 🙏

internal/provider/openconfig/user_test.go
rgildein: If I move it, lint is failing:
```bash
make lint
.../network-operator/bin/golangci-lint-custom run
internal/provider/openconfig/user.go:10:1: File is not properly formatted (gofmt)

^
1 issues:
* gofmt: 1
make: *** [Makefile:118: lint] Error 1
```

rgildein: Ah sorry this is the test file I was looking into user.go. Fixing.

felix-kaestner (approved):

PR #517, opened 2026-08-21
Implement Openconfig NTP provider

adamtrizuljak-sap:
Work in progress

- Add OpenConfig provider implementation for the NTP resource (`internal/provider/openconfig/ntp.go`) and a gnmi testdata file (`test/gnmi/testdata/openconfig/ntp.txtar`).
- ~Add optional field `NTPSpec.SourceAddress` and validation logic (see below)~
- Get the IP address of the `NTPSpec.SourceInterface` from the device using the GRPC client and use it to populate the `source-address` field in Openconfig
- Add a simple Containerlab setup with one SRLinux node + the associated network-operator `Device` and `NTP` resources

Applying [examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml](https://github.com/ironcore-dev/network-operator/pull/517/changes#diff-4fc85bc0e0c09287294d749ec44d3a80f7027069819d395d21a457bbec3075ce) produces the following config in SRLinux:

```json
--{ + running }--[  ]--
A:admin@srl# info system ntp | as json
{
  "admin-state": "enable",
  "server": [
    {
      "address": "de.pool.ntp.org",
      "prefer": true,
      "network-instance": "mgmt",
      "source-address": "172.20.20.2"
    },
    {
      "address": "pool.ntp.org",
      "network-instance": "mgmt",
      "source-address": "172.20.20.2"
    }
  ]
}
```

# Cisco vs Openconfig compatibility issue (solved)
`NTPSpec` defines the `SourceInterfaceName` field https://github.com/ironcore-dev/network-operator/blob/main/api/core/v1alpha1/ntp_types.go#L36. Openconfig has something similar `source-address` https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-ntp-servers-server-config-source-address

A way to handle it would be to look up the interface given by SourceInterfaceName , get its IPv4 address and put it into `source-address`. Does it make sense? Do you usually do such operations, or do you keep the provider code clean from reading any additional k8s resources?

Other way would be to just ignore `SourceInterfaceName` and not pass it into the Openconfig struct, or even throw a `apistatus.FieldViolation` saying that it's unsupported

## Solution

After [discussion](https://github.com/ironcore-dev/network-operator/pull/517#discussion_r3925691528) we decided to do the original proposal and obtain the IP address from the interface. However, if the management interface is used for NTP we can't obtain the address from an `Interface` resource, because we don't create such resource for the management interface. Therefore the provider reads the interface state directly from the device over GRPC and uses its IP address to populate the `source-address` field.

This solution does not change the NTPSpec and should work with any `SourceInterfaceName` that has an IP address (either manually configured, or obtained from DHCP).

## Previous solution - abandoned  

<strike>Discussed with @nikatza and @rgildein 

- Mark `NTPSpec.SourceInterfaceName` as optional
- Add optional field `NTPSpec.SourceAddress`
  - Maps to https://pubhub.devnetcloud.com/media/dme-docs-10-4-3/docs/System/datetime:NtpSrcIp/ in Cisco
  - Maps to https://openconfig.net/projects/models/schemadocs/yangdoc/openconfig-system.html#system-ntp-servers-server-config-source-address in Openconfig
- In Cisco controller, ensure that at least one of the fields is defined
- In Openconfig controller, throw an unsupported error if `SourceInterfaceName` is defined

Since both fields are optional, the change will be backwards-compatible with the previous NTPSpec and the generated custom resources.</strike>

### DIFF
--- a/Tiltfile
+++ b/Tiltfile
@@ -207,8 +207,8 @@ k8s_resource(new_name='mac-entry', objects=['mac-entry:probe'], trigger_mode=TRI
 k8s_resource(new_name='route-prefix', objects=['route-prefix:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 k8s_resource(new_name='vtep-peers', objects=['vtep-peers:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 
-k8s_yaml('./examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml')
-k8s_resource(new_name='ntp-clab', objects=['ntp-clab:ntp'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['containerlab'])
+k8s_yaml('./config/samples/nokia/srlinux/v1alpha1_ntp.yaml')
+k8s_resource(new_name='ntp-srlinux', objects=['ntp-srlinux:ntp'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['srlinux'])
 
 print('🚀 network-operator development environment')
 print('👉 Edit the code inside the api/, cmd/, or internal/ directories')
--- a/Tiltfile
+++ b/Tiltfile
@@ -207,6 +207,9 @@ k8s_resource(new_name='mac-entry', objects=['mac-entry:probe'], trigger_mode=TRI
 k8s_resource(new_name='route-prefix', objects=['route-prefix:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 k8s_resource(new_name='vtep-peers', objects=['vtep-peers:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 
+k8s_yaml('./examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml')
+k8s_resource(new_name='ntp-clab', objects=['ntp-clab:ntp'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['containerlab'])
+
 print('🚀 network-operator development environment')
 print('👉 Edit the code inside the api/, cmd/, or internal/ directories')
 print('👉 Tilt will automatically rebuild and redeploy when changes are detected')
--- a/api/core/v1alpha1/ntp_types.go
+++ b/api/core/v1alpha1/ntp_types.go
@@ -56,7 +56,6 @@ type NTPServer struct {
 	Prefer bool `json:"prefer,omitempty"`
 
 	// The name of the vrf used to communicate with the NTP server.
-	// Maps to network-instance in Openconfig
 	// +optional
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
--- a/api/core/v1alpha1/ntp_types.go
+++ b/api/core/v1alpha1/ntp_types.go
@@ -56,6 +56,7 @@ type NTPServer struct {
 	Prefer bool `json:"prefer,omitempty"`
 
 	// The name of the vrf used to communicate with the NTP server.
+	// Maps to network-instance in Openconfig
 	// +optional
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
--- a/config/samples/nokia/srlinux/v1alpha1_ntp.yaml
+++ b/config/samples/nokia/srlinux/v1alpha1_ntp.yaml
@@ -5,7 +5,7 @@ metadata:
     app.kubernetes.io/name: network-operator
     app.kubernetes.io/managed-by: kustomize
     networking.metal.ironcore.dev/device-name: leaf1
-  name: ntp-clab
+  name: ntp-srlinux
 spec:
   deviceRef:
     name: leaf1
--- a/examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml
+++ b/examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml
@@ -1,0 +1,19 @@
+apiVersion: networking.metal.ironcore.dev/v1alpha1
+kind: NTP
+metadata:
+  labels:
+    app.kubernetes.io/name: network-operator
+    app.kubernetes.io/managed-by: kustomize
+    networking.metal.ironcore.dev/device-name: leaf1-clab
+  name: ntp-clab
+spec:
+  deviceRef:
+    name: leaf1
+  sourceInterfaceName: mgmt0
+  servers:
+    - address: de.pool.ntp.org
+      prefer: true
+      vrfName: mgmt
+    - address: pool.ntp.org
+      prefer: false
+      vrfName: mgmt
--- a/examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml
+++ b/examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml
@@ -4,7 +4,7 @@ metadata:
   labels:
     app.kubernetes.io/name: network-operator
     app.kubernetes.io/managed-by: kustomize
-    networking.metal.ironcore.dev/device-name: leaf1-clab
+    networking.metal.ironcore.dev/device-name: leaf1
   name: ntp-clab
 spec:
   deviceRef:
--- a/internal/provider/openconfig/interface.go
+++ b/internal/provider/openconfig/interface.go
@@ -6,6 +6,7 @@ package openconfig
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
@@ -663,4 +664,28 @@ type SubinterfaceVlanDoubleTagged struct {
 type SubinterfaceVlanDoubleTaggedConfig struct {
 	InnerVlanID uint16 `json:"inner-vlan-id,omitempty"`
 	OuterVlanID uint16 `json:"outer-vlan-id,omitempty"`
+}
+
+type interfaceAddrs struct {
+	ifName  string
+	Address gnmiext.List[string, *IPv4Address] `json:"address"`
+}
+
+func (a *interfaceAddrs) XPath() string {
+	return fmt.Sprintf("openconfig-interfaces:interfaces/interface[name=%s]/subinterfaces/subinterface[index=0]/openconfig-if-ip:ipv4/addresses", a.ifName)
+}
+
+// interfaceIPAddr retrieves the first IPv4 address from the state of the named interface.
+func (p *Provider) interfaceIPAddr(ctx context.Context, name string) (string, error) {
+	addrs := &interfaceAddrs{ifName: name}
+	if err := p.client.GetState(ctx, addrs); err != nil {
+		if errors.Is(err, gnmiext.ErrNil) {
+			return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address A", name))
+		}
+		return "", fmt.Errorf("failed to get IPv4 address for interface %q: %w", name, err)
+	}
+	for _, a := range addrs.Address {
+		return a.IP, nil
+	}
+	return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address B", name))
 }
--- a/internal/provider/openconfig/interface.go
+++ b/internal/provider/openconfig/interface.go
@@ -680,12 +680,12 @@ func (p *Provider) interfaceIPAddr(ctx context.Context, name string) (string, er
 	addrs := &interfaceAddrs{ifName: name}
 	if err := p.client.GetState(ctx, addrs); err != nil {
 		if errors.Is(err, gnmiext.ErrNil) {
-			return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address A", name))
+			return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address", name))
 		}
 		return "", fmt.Errorf("failed to get IPv4 address for interface %q: %w", name, err)
 	}
 	for _, a := range addrs.Address {
 		return a.IP, nil
 	}
-	return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address B", name))
+	return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address", name))
 }
--- a/internal/provider/openconfig/ntp.go
+++ b/internal/provider/openconfig/ntp.go
@@ -85,7 +85,7 @@ func (s *NTPServer) Key() string { return s.Address }
 // NTPServerConfig holds the config container for a NTP server.
 type NTPServerConfig struct {
 	Address         string `json:"address"`
-	Prefer          bool   `json:"prefer,omitempty"`
+	Prefer          bool   `json:"prefer"`
 	NetworkInstance string `json:"network-instance,omitempty"` // Maps to VrfName
 	SourceAddress   string `json:"source-address,omitempty"`
 }
--- a/internal/provider/openconfig/ntp.go
+++ b/internal/provider/openconfig/ntp.go
@@ -1,0 +1,91 @@
+// SPDX-FileCopyrightText: 2026 SAP SE or an SAP affiliate company and IronCore contributors
+// SPDX-License-Identifier: Apache-2.0
+
+package openconfig
+
+import (
+	"context"
+
+	"github.com/ironcore-dev/network-operator/api/core/v1alpha1"
+	"github.com/ironcore-dev/network-operator/internal/provider"
+	"github.com/ironcore-dev/network-operator/internal/transport/gnmiext"
+)
+
+var _ provider.NTPProvider = (*Provider)(nil)
+
+func (p *Provider) EnsureNTP(ctx context.Context, req *provider.EnsureNTPRequest) error {
+	spec := req.NTP.Spec
+
+	n := &NTP{
+		Config: &NTPConfig{
+			Enabled: spec.AdminState == v1alpha1.AdminStateUp,
+		},
+	}
+
+	var sourceAddress string
+	if spec.SourceInterfaceName != "" {
+		var err error
+		sourceAddress, err = p.interfaceIPAddr(ctx, spec.SourceInterfaceName)
+		if err != nil {
+			return err
+		}
+	}
+
+	if len(spec.Servers) > 0 {
+		n.Servers = &NTPServers{}
+		for _, s := range spec.Servers {
+			n.Servers.Server.Set(&NTPServer{
+				Address: s.Address,
+				Config: &NTPServerConfig{
+					Address:         s.Address,
+					Prefer:          s.Prefer,
+					NetworkInstance: s.VrfName,
+					SourceAddress:   sourceAddress,
+				},
+			})
+		}
+	}
+
+	return p.client.Update(ctx, n)
+}
+
+func (p *Provider) DeleteNTP(ctx context.Context) error {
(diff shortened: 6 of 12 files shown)


test/gnmi/testdata/openconfig/ntp.txt
```diff
@@ -0,0 +1,43 @@
+# Banner PreLogin
```
felix-kaestner:
```suggestion
# NTP
```

examples/openconfig-containerlab/kubernetes/01-devices/kustomization.yaml
felix-kaestner, line 1: Why do we need all these files in `examples/openconfig-containerlab`?

If you just use the following containerlab topology:
```yaml
name: srlinux
topology:
  nodes:
    srl:
      kind: nokia_srlinux
      image: ghcr.io/nokia/srlinux:26.3.1
      startup-config: |-
        system grpc-server mgmt yang-models native
        system aaa authentication admin-user password admin
      ports:
        - 8022:22
        - 9339:57400
```

The default device sample in https://github.com/ironcore-dev/network-operator/blob/main/config/samples/v1alpha1_device.yaml connecting to `host.docker.internal` on port `9339` with `admin:admin` will work out of the box. This is already setup and integrated with the [Tiltfile](https://github.com/ironcore-dev/network-operator/blob/main/Tiltfile#L60-L65), so everything is already there.

api/core/v1alpha1/ntp_types.go
```diff
@@ -56,6 +65,7 @@ type NTPServer struct {
 	Prefer bool `json:"prefer,omitempty"`
 
 	// The name of the vrf used to communicate with the NTP server.
+	// Maps to NetworkInstance in Openconfig. If empty, will be set to `mgmt`.
```
felix-kaestner: We try to avoid such vendor specifics in our core apis, as these are meant to be vendor-agnostic and work across all vendors.

Also note that the management vrf is called differently on different platforms, so it might not always be called "mgmt".

api/core/v1alpha1/ntp_types.go
```diff
@@ -30,10 +30,19 @@ type NTPSpec struct {
...
+	// +kubebuilder:validation:MaxLength=63
+	SourceInterfaceName string `json:"sourceInterfaceName,omitempty"`
+
+	// Source IP address for all NTP traffic.
+	// For Cisco, either SourceInterfaceName or SourceAddress must be defined.
+	// +optional
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
-	SourceInterfaceName string `json:"sourceInterfaceName"`
+	SourceAddress string `json:"sourceAddress,omitempty"`
```
felix-kaestner: Could we just look up the source address by using the source interface name inside of the provider implementation and then use it as source address without the need to alter the API?

internal/provider/openconfig/ntp.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+	return nil
+}
+
+// Compile-time assertions.
+var _ gnmiext.DataElement = (*NTP)(nil)
+
+// DNS represents the OpenConfig /system/ntp container.
+type NTP struct {
+	Config  *NTPConfig  `json:"config,omitempty"`
```
felix-kaestner:
```suggestion
	Config  *NTPConfig  `json:"config"`
```
If a field is always set/present, we shouldn't have a "omitempty" tag on it. Applies also to the other types in this file. See https://github.com/ironcore-dev/network-operator/blob/76034710538f195d268f0c21ab50f384c25ba257/AGENTS.md?plain=1#L148-L164 for the usage guidelines on "omitempty".

internal/provider/openconfig/ntp.go
```diff
@@ -0,0 +1,106 @@
...
+	Servers *NTPServers `json:"servers,omitempty"`
+}
+
+func (*NTP) XPath() string { return "openconfig-system:system/ntp" }
+
+// NTPConfig holds the config container for NTP.
+type NTPConfig struct {
+	Enabled bool `json:"enabled"` // Maps to AdminState
```
felix-kaestner:
```suggestion
	Enabled bool `json:"enabled"`
```
nit: I think the mapping is already clear from the implementation itself, so these comments don't add to much additional value. Would just leave them out.

api/core/v1alpha1/ntp_types.go
adamtrizuljak-sap: This was done specifically for Nokia, but you're right that it should be as much vendor-agnostic as possible. Especially since we will likely be implementing the Nokia provider soon and therefore the Openconfig implementation should remain as pure as possible. I'll remove this.

On the other hand, the name `vrfName` is already vendor-specific :)

api/core/v1alpha1/ntp_types.go
adamtrizuljak-sap: We discussed this with @nikatza. In the current architecture the providers don't make requests to the K8s API to fetch more information, they purely translate the `EnsureXXRequest` to the vendor-specific configuration format. We didn't want to violate this principle.

Other option would be to look up the `Interface` in the NTP controller if  `SourceInterfaceName` and pass it into the `EnsureNTPRequest`.

internal/provider/openconfig/ntp.go
adamtrizuljak-sap: Yeah AdminState is probably obvious and common for all resources. But I'd keep the `NetworkInstance // Maps to VrfName`, since it's unique to NTP and not obvious

internal/provider/openconfig/ntp.go
adamtrizuljak-sap: I'll remove `omitempty` from everything except the `NTPServerConfig` fields which are marked as optional in `ntp_types.go`

examples/openconfig-containerlab/kubernetes/01-devices/kustomization.yaml
adamtrizuljak-sap: That's great! I missed the part where Tiltfile overrides the device IP. It's working nicely for me. 

Therefore, under `examples/openconfig-containerlab` I can keep just the `topology.clab.yml` and the NTP resource file. I think the NTP resource must remain separate because Cisco needs `vrfName: management` and Nokia needs `mgmt`. Unless we want to do some Tiltfile hacks as with the device IP...

api/core/v1alpha1/ntp_types.go
adamtrizuljak-sap: One catch in looking up the interface: When we use `sourceInterfaceName: mgmt0` (or another vendor-specific management interface name), we don't create an `Interface` resource for the management interface. Therefore there is no `Interface` resource to look up and get the IP address. Maybe the only way would be to read the address of the interface directly from the device using the GRPC client? Are you already doing this in some other controllers?

nikatza: Thanks a lot for pushing the OpenConfig provider, really nice :)

I think that what Felix here implies is that we shouldn't change the API to accommodate vendor-specific  behavior. Instead, the `EnsureNTP` method in the openconfig provider could fetch the configuration for the referenced interface and use the assigned IP address to configure NTP. Essentially something along the lines of [this function](https://github.com/ironcore-dev/network-operator/blob/76034710538f195d268f0c21ab50f384c25ba257/internal/provider/cisco/nxos/provider.go#L4085), or calling directly `GetConfig` on the appropiate structure that contains the addressing. This means that in `.spec.SourceInterfaceName` we don't give the name of a k8s `Interface` resource, but the name of the interface in the device itself (e.g., in Cisco `eth1/1`), which the provider can use.

One note with regard to API changes: even if the project is still in early stages we have already agreed on how to handle API changes in [these docs](https://github.com/ironcore-dev/network-operator/blob/76034710538f195d268f0c21ab50f384c25ba257/docs/designs/0006-api-deprecation-strategy.md). It's a bit tedious, but I think we can avoid this here and happily use the existing API :)

adamtrizuljak-sap: I've updated the provider to get the interface's IP from the device. And I've reverted all changes to the NTP API and the NXOS NTP provider.

api/core/v1alpha1/ntp_types.go
```diff
@@ -56,6 +56,7 @@ type NTPServer struct {
 	Prefer bool `json:"prefer,omitempty"`
 
 	// The name of the vrf used to communicate with the NTP server.
+	// Maps to network-instance in Openconfig
```
nikatza: Do we need this comment here? It does seem to bring any relevant information for the API user at first. May be the commit message or the provider implementation are more appropriate place.

config/samples/nokia/srlinux/v1alpha1_ntp.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: networking.metal.ironcore.dev/v1alpha1
```
nikatza, line 1: I fear these new files included in this folder as they don't really fit well here in this PR. I think the examples folder is reserved to more complex scenarios. If that is OK for you we could remove them and clarify in our weekly. What do you think?

internal/provider/openconfig/interface.go
```diff
@@ -664,3 +665,27 @@ type SubinterfaceVlanDoubleTaggedConfig struct {
...
 	OuterVlanID uint16 `json:"outer-vlan-id,omitempty"`
 }
+
+type interfaceAddrs struct {
+	ifName  string
+	Address gnmiext.List[string, *IPv4Address] `json:"address"`
+}
+
+func (a *interfaceAddrs) XPath() string {
+	return fmt.Sprintf("openconfig-interfaces:interfaces/interface[name=%s]/subinterfaces/subinterface[index=0]/openconfig-if-ip:ipv4/addresses", a.ifName)
```
nikatza, line 675: We have two hard-coded elements here: the index (0) and the AFI (IPv4). I think we should be able to fetch IPv6 or indicate the address family we want. Same goes for the index. What we could do is add these fields in `interfaceAddrs` (with tag `json:"-"`) and build our xpath using those values. 

If so the provider should be changed accordingly, e.g., use index 0 when the interface name matches some constraints (or just document why 0 is hard-coded). For the IPv4 and IPv6 AFI: you could make that  dependent on the AFI of the `NTPServers` in the spec.

internal/provider/openconfig/interface.go
```diff
@@ -664,3 +665,27 @@ type SubinterfaceVlanDoubleTaggedConfig struct {
...
+}
+
+// interfaceIPAddr retrieves the first IPv4 address from the state of the named interface.
+func (p *Provider) interfaceIPAddr(ctx context.Context, name string) (string, error) {
```
nikatza, line 679: We should be able to get IPv6 addresses too :)

internal/provider/openconfig/interface.go
```diff
@@ -664,3 +665,27 @@ type SubinterfaceVlanDoubleTaggedConfig struct {
...
+	addrs := &interfaceAddrs{ifName: name}
+	if err := p.client.GetState(ctx, addrs); err != nil {
+		if errors.Is(err, gnmiext.ErrNil) {
+			return "", apistatus.NewFailedPreconditionError(fmt.Sprintf("interface %q has no IPv4 address A", name))
```
nikatza: "IPv4 address A" --> "IPv4 address"?

internal/provider/openconfig/ntp.go
```diff
@@ -0,0 +1,91 @@
...
+	var sourceAddress string
+	if spec.SourceInterfaceName != "" {
+		var err error
+		sourceAddress, err = p.interfaceIPAddr(ctx, spec.SourceInterfaceName)
+		if err != nil {
+			return err
+		}
+	}
+
+	if len(spec.Servers) > 0 {
```
nikatza: This guard could be ignored.

examples/openconfig-containerlab/kubernetes/01-devices/kustomization.yaml
felix-kaestner: Just one thing to note, if we accumulate a resource sample for each provider, that will lead to many files quite quickly. So we might want to find a more sutable structure for that.

api/core/v1alpha1/ntp_types.go
felix-kaestner: I wouldn't say that `vrfName` is vendor specific. The term `VRF` is quite universally known, the term is even used in the linux kernel (https://docs.kernel.org/networking/vrf.html). And it also refers to the `VRF` resource type that we have defined (https://github.com/ironcore-dev/network-operator/blob/main/api/core/v1alpha1/vrf_types.go) within it's `spec.name`.

Tiltfile
```diff
@@ -207,6 +207,9 @@ k8s_resource(new_name='mac-entry', objects=['mac-entry:probe'], trigger_mode=TRI
 k8s_resource(new_name='route-prefix', objects=['route-prefix:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 k8s_resource(new_name='vtep-peers', objects=['vtep-peers:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 
+k8s_yaml('./examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml')
+k8s_resource(new_name='ntp-clab', objects=['ntp-clab:ntp'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['containerlab'])
```
nikatza: I would propose to leave the existing resource from samples and keep the tilt file compact.

api/core/v1alpha1/ntp_types.go
felix-kaestner:
```suggestion
	// Maps to network-instance in Openconfig.
```

examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: networking.metal.ironcore.dev/v1alpha1
+kind: NTP
+metadata:
+  labels:
+    app.kubernetes.io/name: network-operator
+    app.kubernetes.io/managed-by: kustomize
+    networking.metal.ironcore.dev/device-name: leaf1-clab
```
felix-kaestner:
```suggestion
    networking.metal.ironcore.dev/device-name: leaf1
```

internal/provider/openconfig/interface.go
felix-kaestner: What is "IPv4 address A" and "IPv4 address B" in this file?

nikatza (commented):
I have some minor comments, happy to discuss them if needed :) 🚀

examples/openconfig-containerlab/kubernetes/01-devices/kustomization.yaml
adamtrizuljak-sap: I understand. But I think that from a user perspective it would be helpful to have ready-to-go samples for as many providers (aka device vendors) as possible. And we as developers need some baseline resource sample(s) that can be deployed and tested against.

internal/provider/openconfig/ntp.go
```diff
@@ -0,0 +1,91 @@
...
+	return p.client.Delete(ctx, &NTP{})
+}
+
+// Compile-time assertions.
+var _ gnmiext.DataElement = (*NTP)(nil)
+
+// DNS represents the OpenConfig /system/ntp container.
+type NTP struct {
+	Config  *NTPConfig  `json:"config"`
+	Servers *NTPServers `json:"servers"`
```
felix-kaestner, line 62:
```suggestion
	Servers *NTPServers `json:"servers,omitempty"`
```
This field is optional and may be `nil` in which case I think we don't want to serialize `"null"` into the JSON payload. As we do an update (gnmi replace), omitting this field is also fine.

Tiltfile
```diff
@@ -207,6 +207,9 @@ k8s_resource(new_name='mac-entry', objects=['mac-entry:probe'], trigger_mode=TRI
 k8s_resource(new_name='route-prefix', objects=['route-prefix:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 k8s_resource(new_name='vtep-peers', objects=['vtep-peers:probe'], trigger_mode=TRIGGER_MODE_MANUAL, auto_init=False, labels=['samples'])
 
+k8s_yaml('./examples/openconfig-containerlab/kubernetes/01-devices/v1alpha1-leaf1-clab-ntp.yaml')
```
felix-kaestner: The naming here is a bit weird. This is nothing really containerlab specific in the sample.

internal/provider/openconfig/ntp.go
```diff
@@ -0,0 +1,91 @@
...
+	Address string           `json:"address"`
+	Config  *NTPServerConfig `json:"config"`
+}
+
+func (s *NTPServer) Key() string { return s.Address }
+
+// NTPServerConfig holds the config container for a NTP server.
+type NTPServerConfig struct {
+	Address         string `json:"address"`
+	Prefer          bool   `json:"prefer,omitempty"`
```
felix-kaestner:
```suggestion
	Prefer          bool   `json:"prefer"`
```
I think we don't want to omit a value of `false` (default).

internal/provider/openconfig/interface.go
adamtrizuljak-sap: I could do it, but in general we don't support IPv6 in network-operator?

config/samples/nokia/srlinux/v1alpha1_ntp.yaml
adamtrizuljak-sap: I created this file because Openconfig requires different settings from `config/samples/v1alpha1_ntp.yaml`.

As I wrote here https://github.com/ironcore-dev/network-operator/pull/517#discussion_r3958126631, my idea is to have ready-to-go samples for all providers which can be used during development for testing against a Containerlab setup or similar. If I only have `config/samples/v1alpha1_ntp.yaml` in the repo, I would need to overwrite it each time, discard the changes afterwards (and pay attention to not commit it) and do this for every resource that differs -> not a very streamlined and uniform development workflow.

Do you have an idea for better placement of such samples? Maybe move `config/samples/` into something like `config/samples/common` and place this file into `config/samples/openconfig`?

internal/provider/openconfig/interface.go
adamtrizuljak-sap: It's a simple lookup to get the first IP address of the interface given by `NTPSpec.SourceInterfaceName`, which we did to avoid adding the `NTPSpec.SourceAddress` field. I am not sure how to include the information (especially the interface index) without again modifying `NTPSpec`...

internal/provider/openconfig/interface.go
adamtrizuljak-sap: Oops, left-over from debug :)

adamtrizuljak-sap: Oops, left-over from debug :)

Tiltfile
adamtrizuljak-sap: Openconfig (actually Nokia SRLinux) needs `vrfName: mgmt`, whereas the default samples have `vrfName: management`.

See https://github.com/ironcore-dev/network-operator/pull/517#discussion_r3978308796 and maybe let's discuss it together with @nikatza?

internal/provider/openconfig/ntp.go
adamtrizuljak-sap: In the resource definition the field is required and has min length 1. So it should never be empty here?

https://github.com/ironcore-dev/network-operator/blob/f0111fe9206ef20f91ea47f7701629c059fa1173/api/core/v1alpha1/ntp_types.go#L42

config/samples/nokia/srlinux/v1alpha1_ntp.yaml
adamtrizuljak-sap: As discussed in #557, I moved these samples into vendor-specific path  `config/samples/nokia/srlinux/`

internal/provider/openconfig/interface.go
adamtrizuljak-sap: #564 has since happened. But we decided on Thursday that we don't need to support IPv6 for NTP source address yet.

internal/provider/openconfig/interface.go
nikatza: Please add a TODO  or update the commit body :)

nikatza (approved):

felix-kaestner (approved):

internal/provider/openconfig/interface.go
adamtrizuljak-sap: Done :)

PR #522, opened 2026-08-25
feat(test/gnmi): Enable `cisco-nxos-gnmi`

nikatza:
Details in commit messages. This PR enables preliminaty set of gnmi tests for cisco nxos.

### DIFF
--- a/test/gnmi/testdata/cisco-nxos-gnmi/aaa.txtar
+++ b/test/gnmi/testdata/cisco-nxos-gnmi/aaa.txtar
@@ -1,0 +1,158 @@
+-- aaaconfigs/settings --
+apiVersion: nx.cisco.networking.metal.ironcore.dev/v1alpha1
+kind: AAAConfig
+metadata:
+  name: settings
+  namespace: default
+spec:
+  loginErrorEnable: true
+  keyEncryption: Type6
+
+-- secrets/tacacs-keys --
+apiVersion: v1
+kind: Secret
+metadata:
+  name: tacacs-keys
+  namespace: default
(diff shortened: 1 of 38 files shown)


felix-kaestner:
Please note that this project doesn't follow semantic commits: https://github.com/ironcore-dev/network-operator/blob/main/AGENTS.md?plain=1#L53-L57

test/gnmi/testdata/cisco-nxos-gnmi/aaa.txtar
```diff
@@ -0,0 +1,160 @@
+# NX-OS AAA TACACS+ with provider config (loginErrorEnable, keyEncryption)
```
felix-kaestner: nit: do we need that line? Also I guess "NX-OS" is already clear from the package? (Same applies to all test files).

test/gnmi/testdata/cisco-nxos-gnmi/bgp_peer.txtar
felix-kaestner, line 1: Shouldn't this just be called `bgp_peer.go`?

felix-kaestner (approved):

test/gnmi/testdata/cisco-nxos-gnmi/aaa.txtar
nikatza: I had always in mind to have more complex tests cases describing functional roles involving multiple resources, e.g.,  border-leaf, core-router. This is the reason why this is there, but we can remove it for now.

PR #528, opened 2026-08-28
Complete EVPN Fabric sample and reconciliation support

felix-kaestner:
_Details in each commit.._


internal/controller/evpn/fabric_controller.go
```diff
@@ -244,15 +250,133 @@ func (r *FabricReconciler) reconcile(ctx context.Context, fabric *evpnv1alpha1.F
...
+	}
+
+	switch {
+	case totalAdjacencies == 0:
+		conditions.Set(fabric, metav1.Condition{
+			Type:    evpnv1alpha1.UnderlayConvergedCondition,
+			Status:  metav1.ConditionFalse,
+			Reason:  evpnv1alpha1.NoResourcesReason,
+			Message: "No underlay adjacencies expected",
+		})
```
IvoGoman, line 319: The condition message seems off to me. Should it be `No underlay adjacencies created yet` to signal the reconciliation is not yet done, analogous to the overlay?

Otherwise, if no underlay adjacencies is a valid outcome. I would expect a case for `fullAdjacencies == 0 && totalAdjacencies > 0` and one for `fullAdjacencies == 0 && totalAdjacencies == 0`.

felix-kaestner: Adjusted the wording.

internal/controller/core/ospf_controller.go
```diff
@@ -327,6 +329,26 @@ func (r *OSPFReconciler) reconcile(ctx context.Context, s *ospfScope) (reterr er
...
+
+		if !intf.HasIPv4() {
+			conditions.Set(s.OSPF, metav1.Condition{
+				Type:    v1alpha1.ConfiguredCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  v1alpha1.IPAddressingNotFoundReason,
+				Message: fmt.Sprintf("interface %q has no IPv4 configuration", ref.Name),
+			})
+			return reconcile.TerminalError(fmt.Errorf("interface %q has no IPv4 configuration", ref.Name))
+		}
```
nikatza, line 350: At the moment the [this controller watches](https://github.com/ironcore-dev/network-operator/blob/67e3ff5db8e3b596c6b3d3a744e198a5f5ec23a1/internal/controller/core/ospf_controller.go#L268) would trigger only on configuration change. Since we can have a `configured` interface without addressing, should we add a watcher?

internal/controller/core/ospf_controller_test.go
```diff
@@ -118,6 +119,70 @@ var _ = Describe("OSPF Controller", func() {
 		})
 	})
 
+	Context("When a referenced interface has no IPv4 configuration", func() {
```
nikatza, line 124: Could we add to this test a change in the interface config, namely that the update the address configuration?

internal/controller/core/ospf_controller.go
felix-kaestner: Exactly this was added? 🤔 https://github.com/ironcore-dev/network-operator/blob/239b5b4c4bfcbb329c911b3f41e1e9dc2e5a06bf/internal/controller/core/ospf_controller.go#L277-L280

internal/controller/core/ospf_controller_test.go
felix-kaestner: Added.

internal/controller/core/ospf_controller.go
nikatza: I didn't catch that!

nikatza (approved):

PR #530, opened 2026-08-28
Add device provisioning documentation

swagner-de:
Document the platform-agnostic provisioning contract in
docs/concepts/provisioning.md: the Device resource, lifecycle phases,
HTTP endpoints, source validation, the ProvisioningProvider hook
interface, and password handling. Document Cisco NX-OS specifics
(POAP, install all, password hash formats) in docs/concepts/ztp-nxos.md,
with the reference POAP boot script at hack/ztp/nxos.py.

Signed-off-by: Sebastian Wagner <sebastian.wagner02@sap.com>

### DIFF
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -82,6 +82,8 @@ export default withMermaid({
                     { text: 'Config Backups', link: '/concepts/config-backup' },
                     { text: 'Pausing Reconciliation', link: '/concepts/pausing' },
                     { text: 'Numbered Resources', link: '/concepts/numbered-resources' },
+                    { text: 'Device Provisioning', link: '/concepts/provisioning' },
+                    { text: 'ZTP for Cisco NX-OS', link: '/concepts/ztp-nxos' },
                 ],
             },
             {
--- a/docs/concepts/provisioning.md
+++ b/docs/concepts/provisioning.md
@@ -1,0 +1,212 @@
+# Device Provisioning
+
+The network operator can bootstrap a device automatically on first boot, from initial network configuration through OS image upgrade and final handoff to the operator. This page describes the platform-agnostic provisioning contract: the `Device` resource, the lifecycle phases, the HTTP endpoints a device talks to, and the provider hook interface a new platform must implement.
+
+Platform-specific mechanics (how the device obtains and runs the boot script, which on-device commands perform the upgrade, which password hash formats are supported) live on the per-platform pages. For Cisco NX-OS see [Zero-Touch Provisioning for Cisco NX-OS](ztp-nxos.md).
+
+## Overview
+
+Provisioning is driven by a small boot script that runs on the device early in its boot sequence, before it has any operator-managed configuration. The script speaks HTTP(S) to the operator's provisioning server: it identifies the device by serial number, receives an image URL, credentials and a token, then downloads and installs the image and reports progress back. Once the device reboots onto the target image and becomes reachable, the operator takes over and reconciles the device's configuration resources.
+
+Because any platform can supply its own boot script and its own provider implementation, the operator side of this contract stays the same across vendors. What differs per platform is packaged behind the provider hook interface described below.
+
+The design rests on an explicit assumption: provisioning works on any vendor's device that can (a) run a boot script early in its boot sequence which is able to configure the device, and (b) reach the operator over HTTP(S). Platforms that meet these two requirements can plug in to the operator-side contract.
+
+Provisioning is currently implemented only by the Cisco NX-OS provider (see [Zero-Touch Provisioning for Cisco NX-OS](ztp-nxos.md)). The contract described here is the interface a future platform provider would implement.
+
+The overall flow, independent of platform, looks like this:
+
+```mermaid
+sequenceDiagram
+    participant SW as Device
+    participant DHCP as DHCP Server
+    participant OP as Network Operator
+    participant IMG as Image Server
+
+    SW->>DHCP: Request network settings
+    DHCP-->>SW: IP, DNS, NTP, boot-script location
+    Note over SW: Fetch and run boot script
+    SW->>OP: GET /provisioning/config?serial=<serial>
+    OP-->>SW: image URL, checksum, credentials, token
+    SW->>IMG: Download image
+    IMG-->>SW: image
+    SW->>OP: PUT /provisioning/status-report (DownloadingImage, UpgradeStarting, RebootingDevice)
+    Note over SW: Install image and reboot
+    SW->>OP: Reachable on new image
+    OP-->>SW: Reconcile configuration
+```
+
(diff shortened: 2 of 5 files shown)


docs/.vitepress/config.mts
```diff
@@ -82,6 +82,8 @@ export default withMermaid({
                     { text: 'Config Backups', link: '/concepts/config-backup' },
                     { text: 'Pausing Reconciliation', link: '/concepts/pausing' },
                     { text: 'Numbered Resources', link: '/concepts/numbered-resources' },
+                    { text: 'Device Provisioning', link: '/concepts/provisioning' },
+                    { text: 'ZTP for Cisco NX-OS', link: '/concepts/ztp-nxos' },
```
felix-kaestner, line 86: Shouldn't "ZTP for Cisco NX-OS" be a sub-section under the the "Device Provisioning"?

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
+# Device Provisioning
+
+The network operator can bootstrap a device automatically on first boot, from initial network configuration through OS image upgrade and final handoff to the operator. This page describes the platform-agnostic provisioning contract: the `Device` resource, the lifecycle phases, the HTTP endpoints a device talks to, and the provider hook interface a new platform must implement.
```
felix-kaestner, line 3: Lifecycle phases and the http endpoints were already documented in https://github.com/ironcore-dev/network-operator/blob/main/docs/tutorials/device-onboarding.md#device-lifecycle-phases. Please condense this information to a single location.

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
...
+      checksumType: MD5
+    bootScript:
+      configMapRef:
+        name: boot-script
+        key: script
+```
+
+The `provisioning.image` section tells the operator which image the device should run. When the boot script contacts the operator, this is what gets returned in the config response.
+
+The `provisioning.bootScript` holds the boot script the device runs, supplied inline or from a referenced Secret or ConfigMap. The operator ships an optional built-in TFTP server (beta) that serves this script directly: the device requests a filename encoding its serial (`serial-<serial>` or `<serial>.<ext>`), the server resolves the matching `Device`, reads `spec.provisioning.bootScript`, and returns its contents as-is. With source validation enabled it also checks that the client IP matches the device endpoint and the serial matches `status.serialNumber`. If you already run your own TFTP infrastructure, delivery can stay external to the operator instead.
```
felix-kaestner, line 71: Where do we take the "beta" state of the built-in TFTP Server from?

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
...
+|--------|---------|
+| `DownloadingImage` | Image download in progress |
+| `UpgradeStarting` | Checksum verified, install command running |
+| `RebootingDevice` | Install complete, rebooting |
+| `ImageDownloadFailed` | Download or checksum failure |
+| `UpgradeFailed` | Install command failed |
+
+### `GET /provisioning/mtls-client-ca` and `GET /provisioning/device-certificate`
+
+Both are optional. They let the device fetch the operator's mTLS client CA and a per-device certificate so the operator can later connect over an authenticated channel. A device that does not need certificates can ignore these; the operator returns `404` when there is nothing to hand out, which the boot script treats as "skip".
```
felix-kaestner, line 172: "per-device" is nothing the operator does or enforces. It just serves what a user sets in the Device spec. How these server certificates are created, is up to the user.

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
...
+
+## Provisioning provider interface
+
+A platform plugs into provisioning by implementing `ProvisioningProvider` (`internal/provider/provider.go`):
```
felix-kaestner, line 176: Anything in `/internal` is intentionally private and no public API (ref/ https://github.com/golang-standards/project-layout/tree/master/internal). As such I'm not sure we want to document that in a public documentation. With code in `/internal` we reserve the possibility to change that interface at any time w/o external guarantees.

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
...
+
+- `Reprovision` resets the device and re-enables its provisioning mechanism so it can run through the sequence again.
+- `HashProvisioningPassword` turns a plaintext password into a device-native hash plus an algorithm label (see below).
+- `VerifyProvisioned` checks whether the device has finished provisioning and is running the expected image.
+
+## Password handling
+
+The operator never sends a plaintext password to the device over the provisioning channel. That channel may be unauthenticated or otherwise not fully trusted (the device has not yet been secured, and the operator does not verify the device's identity at this stage), so shipping a cleartext credential over it would be unsafe.
+
+Instead the operator reads the admin credential from the device's endpoint Secret and passes the plaintext to the provider's `HashProvisioningPassword` hook. The hook returns two things: a hash in the device's native on-box format, and an algorithm label identifying which hash format it is.
```
felix-kaestner, line 194: We don't have any "admin" semantics here. This is just a user account from the device spec from our perspective.

docs/concepts/provisioning.md
```diff
@@ -0,0 +1,212 @@
...
+
+Both are placed in the config response as `userAccounts[].hashedPassword` and `userAccounts[].hashAlgorithm`. The boot script maps the algorithm label to the platform's corresponding on-device password type and configures the account with the pre-hashed value; the plaintext never leaves the operator.
+
+## Observing provisioning progress
```
felix-kaestner, line 198: This just shows normal `kubectl` usage, nothing specific to our network-operator. As such, do we even want to have that in our documentation. Users can also choose an GUI or any other way of looking at kubernetes resources.

docs/concepts/ztp-nxos.md
```diff
@@ -0,0 +1,37 @@
+# Zero-Touch Provisioning for Cisco NX-OS
+
+Zero-Touch Provisioning (ZTP), referred to as POAP (Power-On Auto Provisioning) on Cisco NX-OS, lets a switch bootstrap itself automatically on first boot, without manual intervention.
+
+This page documents only the Cisco NX-OS specifics. For the platform-agnostic model (the end-to-end flow, DHCP, the `Device` resource, lifecycle phases, HTTP endpoints, source validation, and the provider hook interface) see [Device Provisioning](provisioning.md).
+
+## Boot script
+
+The network operator ships its own POAP boot script at [`hack/ztp/nxos.py`](https://github.com/ironcore-dev/network-operator/blob/main/hack/ztp/nxos.py). It is not Cisco's reference POAP script; it talks to the operator's provisioning API and uses `install all` (see below). Site-specific values like the provisioning server URL have been replaced with placeholders and must be adapted to your environment before use.
```
felix-kaestner, line 9: We don't do that by default (like this sentence implies). This is just an example. Users might set their own version, however they desire.

This whole document is written in this way that the "network operator provides" which is not the case. We should rephrase this to state this as an example for reference.

PR #536, opened 2026-08-31
Add handling of firmware upgrades for nxos

IvoGoman:
- adds new DeviceMaintenance key "firmware-upgrade"
    - pauses reconciliation while upgrade is running
    - adds DeviceMaintenanceFirmwareTargetAnnotation to specify the target firmware version

    nxos provider impletementation
    - ensure enough space, extends NX-API timeout during upgrade.
    - image is copied & verified.
    - show incompatibility-all nxos & show install all impact nxos are run to check software & hw
    - after install & reload the device is checked for version. When boot image matches the target image the upgrade is considered done.

### DIFF
--- a/internal/controller/core/device_controller.go
+++ b/internal/controller/core/device_controller.go
@@ -6,9 +6,11 @@ package core
 import (
 	"cmp"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"math/rand/v2"
+	"net/url"
 	"regexp"
 	"slices"
 	"strings"
@@ -246,6 +248,12 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 	}
 
 	if err := r.reconcileMaintenance(ctx, obj, conn); err != nil {
+		// ErrUpgradeInProgress signals that a long-running maintenance step was
+		// issued and the operation must be resumed on a subsequent reconcile, so
+		// it must requeue rather than terminate.
+		if errors.Is(err, provider.ErrUpgradeInProgress) {
+			return ctrl.Result{}, err
+		}
 		return ctrl.Result{}, reconcile.TerminalError(err)
 	}
 
@@ -468,7 +476,8 @@ func (r *DeviceReconciler) reconcileMaintenance(ctx context.Context, obj *v1alph
 
 	case v1alpha1.DeviceMaintenanceReboot,
 		v1alpha1.DeviceMaintenanceFactoryReset,
-		v1alpha1.DeviceMaintenanceReprovision:
+		v1alpha1.DeviceMaintenanceReprovision,
+		v1alpha1.DeviceMaintenanceFirmwareUpgrade:
 
 		prov := r.Provider()
 		if err := prov.Connect(ctx, conn); err != nil {
@@ -538,6 +547,39 @@ func (r *DeviceReconciler) reconcileMaintenance(ctx context.Context, obj *v1alph
 				return fmt.Errorf("failed to prepare device for reprovisioning: %w", err)
 			}
 			obj.Status.Phase = v1alpha1.DevicePhasePending
+
+		case v1alpha1.DeviceMaintenanceFirmwareUpgrade:
+			mp, ok := prov.(provider.MaintenanceProvider)
+			if !ok {
+				r.Recorder.Eventf(obj, nil, "Warning", "MaintenanceUnsupported", "Maintenance", "Provider does not support firmware upgrade operation: %s", action)
+				return nil
+			}
+			targetFirmware, err := r.getTargetFirmware(obj)
+			if err != nil {
+				return err
+			}
+
+			err = mp.UpgradeFirmware(ctx, conn, targetFirmware)
+			if errors.Is(err, provider.ErrUpgradeInProgress) {
+				conditions.Set(obj, metav1.Condition{
+					Type:    v1alpha1.ReadyCondition,
+					Status:  metav1.ConditionFalse,
+					Reason:  v1alpha1.MaintenanceInProgressReason,
+					Message: "Firmware upgrade in progress",
+				})
+				r.Recorder.Eventf(obj, nil, "Normal", "FirmwareUpgradeInProgress", "Maintenance", "Device firmware upgrade is in progress")
+				return err
+			}
+			if err != nil {
+				conditions.Set(obj, metav1.Condition{
+					Type:    v1alpha1.ReadyCondition,
+					Status:  metav1.ConditionFalse,
+					Reason:  v1alpha1.MaintenanceFailedReason,
+					Message: fmt.Sprintf("Failed to upgrade firmware: %v", err),
+				})
+				r.Recorder.Eventf(obj, nil, "Warning", "FirmwareUpgradeFailed", "Maintenance", "Device firmware upgrade has failed: %v", err)
+				return fmt.Errorf("failed to upgrade firmware: %w", err)
+			}
 		}
 
 	default:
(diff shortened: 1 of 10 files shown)


internal/controller/core/device_controller.go
```diff
@@ -246,6 +248,12 @@ func (r *DeviceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ c
 	}
 
 	if err := r.reconcileMaintenance(ctx, obj, conn); err != nil {
+		// ErrUpgradeInProgress signals that a long-running maintenance step was
+		// issued and the operation must be resumed on a subsequent reconcile, so
+		// it must requeue rather than terminate.
+		if errors.Is(err, provider.ErrUpgradeInProgress) {
+			return ctrl.Result{}, err
```
felix-kaestner: Should this use a `RequeueAfter` instead of an error?

internal/controller/core/device_controller.go
```diff
@@ -538,6 +547,39 @@ func (r *DeviceReconciler) reconcileMaintenance(ctx context.Context, obj *v1alph
...
+
+		case v1alpha1.DeviceMaintenanceFirmwareUpgrade:
+			mp, ok := prov.(provider.MaintenanceProvider)
+			if !ok {
+				r.Recorder.Eventf(obj, nil, "Warning", "MaintenanceUnsupported", "Maintenance", "Provider does not support firmware upgrade operation: %s", action)
+				return nil
+			}
+			targetFirmware, err := r.getTargetFirmware(obj)
+			if err != nil {
+				return err
```
felix-kaestner, line 560: Should this be wrapped in a terminal error? As I can see the `getTargetFirmware`, all error cases represent an invalid annotation value. So that would mean there is no point in retrying this. Once a user edits the resources to correct the value, it would automatically trigger a new reconcilation again.

internal/paused/paused.go
```diff
@@ -107,6 +107,12 @@ func computeCondition(device *v1alpha1.Device, obj Object) metav1.Condition {
 				condition.Message = "Device is not reachable: " + cond.Message
 				return condition
 			}
+			if device.GetAnnotations()[v1alpha1.DeviceMaintenanceAnnotation] == v1alpha1.DeviceMaintenanceFirmwareUpgrade {
```
felix-kaestner: Shall we just generally Pause here, when we find any maintenance annotation? For things like reboot/factory reset it would be similar. Currently there we simply have the case that the device eventually becomes unreachable and resources are paused because of this. But if we have this condition now anyways, I think we could just always pause when there is a maintenance annotation present.

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	"github.com/ironcore-dev/network-operator/internal/transport/nxapi"
+)
+
+// upgradeTimeout is the NX-API client timeout for long-running firmware
+// commands (copy and install), which block synchronously for minutes.
+const upgradeTimeout = 20 * time.Minute
+
+const (
+	// nxosDefaultSessionTimeout is the default NX-API session timeout in seconds.
+	nxosDefaultSessionTimeout = 300
```
felix-kaestner:
```suggestion
	DefaultSessionTimeout = 300
```
nit: we usually omit "nxos" in variable names and comments, as this is already clear from the package name. ref/ https://google.github.io/styleguide/go/decisions#package-vs-exported-symbol-name

internal/provider/provider.go
```diff
@@ -71,6 +75,21 @@ type DevicePort struct {
 	Transceiver string
 }
 
+// TargetFirmware represents the firmware image to be applied to the device, including its URL and optional checksum.
+// This JSON is passed as the value of the DeviceMaintenanceFirmwareTargetAnnotation on the Device resource
```
felix-kaestner: The last sentence is somewhat misleading, as this is on a struct type. I think I would just leave this sentence out, as essentially the provider code doesn't care how/where the value is provided (if it comes from annotation now or a spec later).

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+func (p *Provider) UpgradeFirmware(ctx context.Context, _ *deviceutil.Connection, target provider.TargetFirmware) error {
+	logger := logr.FromContextOrDiscard(ctx)
+
+	upgraded, err := p.isUpgraded(ctx, target)
+	switch {
+	case err != nil:
+		return err
+	case upgraded:
+		return nil
+	}
```
felix-kaestner:
```suggestion
	if err != nil || upgraded {
		return err
	}
```
wouldn't this be identical?

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	// The copy and install commands block for several minutes, so they run on a
+	// clone of the NX-API client that only differs in its longer timeout.
+	nxapiUpgrade, err := p.nxapi.Clone(nxapi.WithTimeout(upgradeTimeout))
+	if err != nil {
+		return fmt.Errorf("failed to create long-timeout nxapi client: %w", err)
+	}
+
+	// Disable POAP and extend NX-API session timeouts before long-running commands.
+	if _, err := p.nxapi.Do(ctx, nxapi.NewRequest(
+		"configure",
```
felix-kaestner:
```suggestion
```
I think this is implicit and doesn't need to be done.

Same also in other places, where I think we can simply omit the "configure".

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	switch {
+	case sum == "":
+		// absent — copy below.
+	case target.MD5 == "":
+		haveValidImage = true // no checksum to compare; presence is enough.
+	case strings.EqualFold(sum, target.MD5):
+		haveValidImage = true
+	default:
+		logger.V(1).Info("Stale image on bootflash, deleting", "file", targetFileName)
+		if _, err := p.nxapi.Do(ctx, nxapi.NewRequest("delete bootflash:"+targetFileName+" no-prompt")); err != nil {
```
felix-kaestner:
```suggestion
		if _, err := p.nxapi.Do(ctx, nxapi.NewRequest("delete bootflash:///"+targetFileName+" no-prompt")); err != nil {
```
I though you need the path to me like this, no?

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	size, err := remoteImageSize(ctx, target.URL)
+	if err != nil {
+		return err
+	}
+	dir, err := p.ListDirectory(ctx, "bootflash:")
+	if err != nil {
+		return fmt.Errorf("nxos firmware: dir bootflash: failed: %w", err)
+	}
+	if size > dir.Bytesfree {
+		// TODO: Check if more than the current target and the current running image are present on the bootflash and delete them to free up space.
```
felix-kaestner: We might also use this path for the ConfigBackup local storage, so I'm not sure if we can that easily delete stuff in this directory. Generally, I'm fine to just error here.

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+
+	if target.MD5 != "" {
+		sum, err := p.fileMD5(ctx, targetFileName)
+		if err != nil {
+			return fmt.Errorf("nxos firmware: verify md5: %w", err)
+		}
+		if !strings.EqualFold(sum, target.MD5) {
+			return fmt.Errorf("nxos firmware: md5 mismatch after copy: got %s want %s", sum, target.MD5)
+		}
+	}
```
felix-kaestner: Given that we duplicate some logic from above again, do we want to restructure the function approach, like:

```
if file.NotExists() { ... }
if Checksum.Matches(expected) { return err }
return nil
```

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	).WithMethod(nxapi.MethodCLIASCII))
+	if err != nil {
+		return fmt.Errorf("nxos firmware: install impact check failed: %w", err)
+	}
+	logCLIResult(logger, impactRes, "Install impact check result")
+	return nil
+}
+
+// doUpgrade saves the running config, installs the firmware without
+// reload, and resets the session timeout to the default.
```
felix-kaestner:
```suggestion
// reload.
```
We don't seem to do that here.

internal/provider/cisco/nxos/firmware_test.go
```diff
@@ -0,0 +1,408 @@
...
+
+	got := cleanCLIOutput(in)
+	if got != want {
+		t.Errorf("cleanCLIOutput mismatch:\ngot:\n%s\n\nwant:\n%s", got, want)
+	}
+}
+
+// fakeGNMI is a minimal gnmiext.Client that returns a canned running version.
```
felix-kaestner: I think we already create a mock of the `gnmiext.Client` for testing, why not use that?

internal/provider/cisco/nxos/firmware_test.go
```diff
@@ -0,0 +1,408 @@
...
+		t.Fatalf("expected ErrUpgradeInProgress for unreachable device, got %v", err)
+	}
+}
+
+// nxapiStub starts an httptest server that responds to each NX-API batch using
+// the provided handler, which maps the list of commands to a JSON body string
+// (the ".result.body" payload) for each command, in order. It returns both the
+// client and the connection so tests can pass conn to UpgradeFirmware for the
+// long-timeout client it creates internally.
+func nxapiStub(t *testing.T, handler func(cmds []string) []string) (*nxapi.Client, *deviceutil.Connection) {
```
felix-kaestner, line 83: Instead of mocking a real nxapi response, do we want to do it similarly as with the `gnmiext.Client` and just convert the client to an interface (with lowercase `*client` serving as the implementation) and then have a mock auto-generated which we then use for testing?

internal/provider/cisco/nxos/firmware.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	msg := strings.ToLower(rpcErr.Error())
+	return strings.Contains(msg, "no such file") || strings.Contains(msg, "not found")
+}
+
+// isDeviceUnreachable reports whether err indicates the device is temporarily
+// unreachable (as opposed to a logical error), covering both NX-API transport
+// errors and gNMI/gRPC unavailability. This is expected while the device is
+// rebooting after an install and should be treated as an in-progress state.
+func isDeviceUnreachable(err error) bool {
```
felix-kaestner: Before calling the `UpgradeFirmware` method, we call `Connect`/`Disconnect` on the Provider (https://github.com/ironcore-dev/network-operator/blob/main/internal/controller/core/device_controller.go#L473-L477), which will already perform an gnmi capabilities request. So I would think, that when a device's gnmi api is unreachable, we don't even call the `UpgradeFirmware` as the controller would already run into a `DeadlineExceeded` on the `Connect` and then just return.

internal/controller/core/device_controller.go
IvoGoman: That should be handled by the adjusted error handling on the return, where everything not `ErrInProgress` is wrapped as terminal.

internal/provider/cisco/nxos/firmware.go
IvoGoman: yes

internal/provider/cisco/nxos/firmware.go
IvoGoman: At some point it would make sense to have some sort of cleanup in here to avoid having to manually delete the firmware files. For now the error should be sufficient. But might need to add this to our backlog.

internal/provider/cisco/nxos/firmware.go
IvoGoman: Ah yes, that moved out since it was not persisted between restarts.

internal/provider/cisco/nxos/firmware.go
felix-kaestner: What about putting all firmware files in a specific directory or just prefixing the file with a certain string, that we then use to delete files matching that prefix.

internal/provider/cisco/nxos/firmware_test.go
felix-kaestner: @IvoGoman you resolve this comment but didn't make any adjustments, so do I just see this as a "Don't care"? 😄

IvoGoman: Is this not what you were talking about? https://github.com/ironcore-dev/network-operator/pull/536/changes/a7b79c6f8920c2f0fd7f1cadcdcbaf919586c300#diff-ae3ffff00bc94fd6c131e25d6dea06f1f5bf2c28e57a00e02b7f2c62122b18d7R34-R39

felix-kaestner: No, I meant to create a client mock like in https://github.com/ironcore-dev/network-operator/blob/main/internal/transport/gnmiext/client.go#L134 to have a client_mock.go file like [this](https://github.com/ironcore-dev/network-operator/blob/main/internal/transport/gnmiext/client_mock.go).

felix-kaestner (approved):
@IvoGoman I guess just some linting that needs to be fixed.

PR #537, opened 2026-08-31
Validate reserved VLANs on Cisco NX-OS

felix-kaestner:
Read the configured VLAN reservation before realizing VLAN and NVE resources. Return a failed precondition when requested IDs overlap the configurable 128-VLAN block or the permanently reserved 4093-4095 range.

### DIFF
--- a/api/core/v1alpha1/vlan_types.go
+++ b/api/core/v1alpha1/vlan_types.go
@@ -24,11 +24,11 @@ type VLANSpec struct {
 	// +optional
 	ProviderConfigRef *TypedLocalObjectReference `json:"providerConfigRef,omitempty"`
 
-	// ID is the VLAN ID. Valid values are between 1 and 4094.
+	// ID is the VLAN ID. Valid values are between 1 and 4095.
 	// Immutable.
 	// +required
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=4094
+	// +kubebuilder:validation:Maximum=4095
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Name is immutable"
 	ID int16 `json:"id"`
 
--- a/internal/provider/cisco/nxos/provider.go
+++ b/internal/provider/cisco/nxos/provider.go
@@ -2857,7 +2857,34 @@ func (p *Provider) DeleteSyslog(ctx context.Context) error {
 	)
 }
 
+// ValidateReservedVLANs rejects VLANs reserved for internal use by NX-OS.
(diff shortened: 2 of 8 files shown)


api/core/v1alpha1/vlan_types.go
```diff
@@ -24,11 +24,11 @@ type VLANSpec struct {
 	// +optional
 	ProviderConfigRef *TypedLocalObjectReference `json:"providerConfigRef,omitempty"`
 
-	// ID is the VLAN ID. Valid values are between 1 and 4094.
+	// ID is the VLAN ID. Valid values are between 1 and 4095.
 	// Immutable.
 	// +required
 	// +kubebuilder:validation:Minimum=1
-	// +kubebuilder:validation:Maximum=4094
+	// +kubebuilder:validation:Maximum=4095
```
nikatza, line 31: Should this be 4092 since "4093-4095 are always for internal use"?

internal/provider/cisco/nxos/provider.go
```diff
@@ -2857,7 +2857,34 @@ func (p *Provider) DeleteSyslog(ctx context.Context) error {
...
+		if vlan >= start && vlan <= end || vlan >= 4093 {
+			return apistatus.NewFailedPreconditionError(fmt.Sprintf("VLAN %d is reserved for internal device use", vlan))
+		}
+	}
+	return nil
+}
+
 func (p *Provider) EnsureVLAN(ctx context.Context, req *provider.VLANRequest) error {
+	if err := p.ValidateReservedVLANs(ctx, []int16{req.VLAN.Spec.ID}); err != nil {
+		return err
```
nikatza, line 2885: Should this be a `TerminalError` instead? On the other hand, the reserved range might be changed externally, so I guess we can use this code and exponentially back-off or we add a new watcher to the `nve` controller to monitor changes on resources of type `system`.

nikatza (approved):

api/core/v1alpha1/vlan_types.go
felix-kaestner: 4093-4095 are only reserved on NX-OS. Not on other vendors. This is a core API.

internal/provider/cisco/nxos/provider.go
felix-kaestner: Yes, had the same though. However, setting up some watcher like this would not be that straight forward, as this would be the NX-specific `System` resource, which would re-trigger the core api. So I rather opted for using a precondition failed, as this setting could otherwise also be done manually from the outside.

PR #540, opened 2026-09-02
add cisco-nxos-gnmi enforcePriv & combine globals

IvoGoman:
add globals to set enforcePriv
combine InformSrcIf, TrapsSrcIf into the globals tree

Signed-off-by: Ivo Gosemann <ivo.gosemann@sap.com>

### DIFF
--- a/api/core/v1alpha1/snmp_types.go
+++ b/api/core/v1alpha1/snmp_types.go
@@ -63,6 +63,10 @@ type SNMPSpec struct {
 	// +listType=set
 	// +kubebuilder:validation:MinItems=1
 	Traps []string `json:"traps,omitempty"`
+
+	// EnforceEncryption enforces SNMP message encryption for all users.
+	// +optional
+	EnforceEncryption bool `json:"enforceEncryption,omitempty"`
 }
 
 type SNMPCommunity struct {
--- a/config/samples/v1alpha1_snmp.yaml
+++ b/config/samples/v1alpha1_snmp.yaml
@@ -12,6 +12,7 @@ spec:
   contact: IronCore
   location: rack1
   sourceInterfaceName: mgmt0
+  enforceEncryption: true
   communities:
     - name: snmpcollector
       group: network-operator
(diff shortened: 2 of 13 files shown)


config/samples/v1alpha1_snmp.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   contact: IronCore
   location: rack1
   sourceInterfaceName: mgmt0
+  enforcePrivacy: "yes"
```
felix-kaestner:
```suggestion
  enforcePrivacy: true
```

api/core/v1alpha1/snmp_types.go
```diff
@@ -63,6 +63,10 @@ type SNMPSpec struct {
 	// +listType=set
 	// +kubebuilder:validation:MinItems=1
 	Traps []string `json:"traps,omitempty"`
+
+	// EnforcePrivacy enforces SNMP message encryption (privacy) for all users.
```
felix-kaestner: Maybe there is a slightly better name for this API field as the functionality is to actually enable message encryption in SNMPv3. So maybe something like `EnforceEncryption: true ` would fit better?

internal/provider/cisco/nxos/snmp.go
```diff
@@ -29,14 +29,19 @@ func (*SNMPSysInfo) XPath() string {
 	return "System/snmp-items/inst-items/sysinfo-items"
 }
 
-// SNMPSrcIf represents the SNMP source interface configuration on a NX-OS device.
-type SNMPSrcIf struct {
-	Type   MessageType    `json:"-"`
-	Ifname Option[string] `json:"ifname"`
+// SNMPGlobals represents the SNMP global configuration on a NX-OS device.
+type SNMPGlobals struct {
+	EnforcePriv  Option[string] `json:"enforcePrivacy"`
```
felix-kaestner: How do you feel about creating yet another type to represent a boolean on Cisco NX-OS in our glorious "_100-ways to have a boolean on NX-OS_" file in https://github.com/ironcore-dev/network-operator/blob/main/internal/provider/cisco/nxos/admin.go?

felix-kaestner (approved):

PR #552, opened 2026-09-08
Add PIM logNeighborChanges field

rgildein:
# Add PIM log-neighbor-changes to CRD

## Description

Add PIM logNeighborChanges field

- Added `logNeighborChanges *bool` field to `PIMSpec` in `api/core/v1alpha1/pim_types.go`
- Added `LogNbhChng *bool` (`json:"logNbhChng,omitempty"`) to the NX-OS `PIMDom` struct in `internal/provider/cisco/nxos/pim.go` — field is omitted when nil (user did not configure it), avoiding spurious gNMI Sets
- Wired the field in `EnsurePIM` (`internal/provider/cisco/nxos/provider.go`): `dom.LogNbhChng = req.PIM.Spec.LogNeighborChanges`
- Added `pim_dom_log` test fixture + golden file in `internal/provider/cisco/nxos/pim_test.go` and `testdata/pim_dom_log.json`
- Updated `test/gnmi/testdata/cisco-nxos-gnmi/pim.txtar` with `logNeighborChanges: true` in spec and `"logNbhChng": true` in expected gNMI state

### Phase 1 — local checks

| Check | Result |
|---|---|
| Lint | ✓ 0 issues |
| Unit tests | ✓ 21/21 packages passed |
| gNMI integration tests | ✓ 20/20 passed |

### Phase 2 — real device (Cisco NX-OS, `10.1.1.1:57403`)

Applied `PIM` CR with `logNeighborChanges: true` referencing loopback `lo99`.

**Operator reconcile log (first apply):**
```
Updating  path=System/pim-items/inst-items/dom-items/Dom-list[name=default]
          payload={"name":"default","adminSt":"enabled","logNbhChng":true}
```

**gnmic GET response** (`System/pim-items/inst-items/dom-items/Dom-list[name=default]`):
```json
[
  {
    "source": "10.1.1.1:57403",
    "timestamp": 1788878844530406667,
    "time": "2026-09-08T16:47:24.530406667+02:00",
    "updates": [
      {
        "Path": "System/pim-items/inst-items/dom-items/Dom-list[name=default]",
        "values": {
          "System/pim-items/inst-items/dom-items/Dom-list": [
            {
              "adminSt": "enabled",
              "if-items": {
                "If-list": [
                  {
                    "id": "lo99",
                    "ipAddr": "10.99.99.99/32",
                    "pimSparseMode": true
                  }
                ]
              },
              "jpDelay": 100,
              "logNbhChng": true,
              "mtu": 1500,
              "name": "default"
            }
          ]
        }
      }
    ]
  }
]
```

**Idempotency:** On subsequent reconciles (after `logNbhChng` was already set), the operator emitted:
```
Configuration is already up-to-date  path=System/pim-items/inst-items/dom-items/Dom-list[name=default]
```
No gNMI Set was issued — idempotency confirmed.

## Notes

The NX-OS YANG field name is `logNbhChng` (bool), not a string enum. The platform default is `false`.
Since the field is a pointer with `omitempty`, omitting `logNeighborChanges` from the CR spec sends no value to the device, leaving the platform default intact.

### DIFF
--- a/internal/provider/cisco/nxos/pim.go
+++ b/internal/provider/cisco/nxos/pim.go
@@ -27,8 +27,9 @@ func (*PIM) XPath() string {
 }
 
 type PIMDom struct {
-	Name    string  `json:"name"`
-	AdminSt AdminSt `json:"adminSt"`
+	Name       string  `json:"name"`
+	AdminSt    AdminSt `json:"adminSt"`
+	LogNbhChng *bool   `json:"logNbhChng,omitempty"`
 }
 
 func (*PIMDom) IsListItem() {}
--- a/internal/provider/cisco/nxos/provider.go
+++ b/internal/provider/cisco/nxos/provider.go
@@ -2389,6 +2389,7 @@ func (p *Provider) EnsurePIM(ctx context.Context, req *provider.EnsurePIMRequest
 	if req.PIM.Spec.AdminState == v1alpha1.AdminStateDown {
 		dom.AdminSt = AdminStDisabled
 	}
+	dom.LogNbhChng = req.PIM.Spec.LogNeighborChanges
 	sb.Patch(dom)
 
 	rpItems := new(StaticRPItems)
--- a/api/core/v1alpha1/pim_types.go
+++ b/api/core/v1alpha1/pim_types.go
@@ -29,6 +29,10 @@ type PIMSpec struct {
 	// +kubebuilder:default=Up
 	AdminState AdminState `json:"adminState,omitempty"`
 
+	// LogNeighborChanges enables logging when a PIM neighbor is added or removed.
+	// +optional
+	LogNeighborChanges *bool `json:"logNeighborChanges,omitempty"`
+
 	// RendezvousPoints defines the list of rendezvous points for sparse mode multicast.
 	// +optional
 	// +listType=map
--- a/api/core/v1alpha1/zz_generated.deepcopy.go
+++ b/api/core/v1alpha1/zz_generated.deepcopy.go
@@ -3535,6 +3535,11 @@ func (in *PIMSpec) DeepCopyInto(out *PIMSpec) {
 		*out = new(TypedLocalObjectReference)
 		**out = **in
 	}
+	if in.LogNeighborChanges != nil {
+		in, out := &in.LogNeighborChanges, &out.LogNeighborChanges
+		*out = new(bool)
+		**out = **in
+	}
 	if in.RendezvousPoints != nil {
 		in, out := &in.RendezvousPoints, &out.RendezvousPoints
 		*out = make([]RendezvousPoint, len(*in))
--- a/charts/network-operator/templates/crd/pim.networking.metal.ironcore.dev.yaml
+++ b/charts/network-operator/templates/crd/pim.networking.metal.ironcore.dev.yaml
@@ -115,6 +115,10 @@ spec:
                 minItems: 1
                 type: array
                 x-kubernetes-list-type: atomic
+              logNeighborChanges:
+                description: LogNeighborChanges enables logging when a PIM neighbor
+                  is added or removed.
+                type: boolean
               providerConfigRef:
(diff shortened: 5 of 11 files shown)


felix-kaestner:
@rgildein did you check whether this field is some vendor-agnostic standard that we want to add to the "core" api types or something Cisco-specific hence should rather go to a provider config? ref/ https://github.com/ironcore-dev/network-operator/blob/eb3ad0cbc1979b4081ff4350d171abde26413ed5/docs/designs/0002-provider-interface.md?plain=1#L50-L62

rgildein:
> @rgildein did you check whether this field is some vendor-agnostic standard that we want to add to the "core" api types or something Cisco-specific hence should rather go to a provider config? ref/
> 
> https://github.com/ironcore-dev/network-operator/blob/eb3ad0cbc1979b4081ff4350d171abde26413ed5/docs/designs/0002-provider-interface.md?plain=1#L50-L62

Good call, you are right. I check it and there is no alternative for this in OpenConfig. What would be an option here? Going with custom resource for Cisco or refer Cisco specific resource in PIM?

rgildein:
@felix-kaestner after some conversation with @swagner-de, we agreed that it's ok to put this to the CRD schema instead of creating separate resource for such small "knob". Are you ok with it or do you have another opinion? Thanks

felix-kaestner:
> @felix-kaestner after some conversation with @swagner-de, we agreed that it's ok to put this to the CRD schema instead of creating separate resource for such small "knob". Are you ok with it or do you have another opinion? Thanks

@rgildein I thought the conclusion of our discussions was that this option would rather go into a dedicated "PIMConfig" provider config reference for Cisco NX-OS?

internal/provider/cisco/nxos/pim.go
```diff
@@ -27,8 +27,9 @@ func (*PIM) XPath() string {
 }
 
 type PIMDom struct {
-	Name    string  `json:"name"`
-	AdminSt AdminSt `json:"adminSt"`
+	Name       string  `json:"name"`
+	AdminSt    AdminSt `json:"adminSt"`
+	LogNbhChng *bool   `json:"logNbhChng,omitempty"`
```
felix-kaestner:
```suggestion
	LogNbhChng bool   `json:"logNbhChng"`
```
This field seems to always be present, with a default value of `false`. 

Also, please pay special attention when using "omitempty". ref/ https://github.com/ironcore-dev/network-operator/blob/main/AGENTS.md?plain=1#L148-L164

internal/provider/cisco/nxos/provider.go
```diff
@@ -2444,6 +2451,7 @@ func (p *Provider) EnsurePIM(ctx context.Context, req *provider.EnsurePIMRequest
 	if req.PIM.Spec.AdminState == v1alpha1.AdminStateDown {
 		dom.AdminSt = AdminStDisabled
 	}
+	dom.LogNbhChng = cfg.Spec.LogNeighborChanges
```
```suggestion
	dom.LogNbhChng = false
	if cfg.Spec.LogNeighborChanges != nil {
	    dom.LogNbhChng = *cfg.Spec.LogNeighborChanges
	}
```

internal/provider/cisco/nxos/pim.go
rgildein: You are right, I moved it from core (where it was empty) and forgot to remove it :+1:

internal/provider/cisco/nxos/provider.go
rgildein: This will not be needed since we removed the omitempty?

internal/provider/cisco/nxos/pim.go
rgildein: Sorry misunderstand with omitempty we are talking about, now I'm getting it and I do not realised that not having default value could cause that we are simple ignore configuration during another write. Also now commit below make more sense to me also.
Resolving.

felix-kaestner (approved):

PR #554, opened 2026-09-09
SuppressARP per VNI

adamtrizuljak-sap:
# Summary

The NVE-level `suppressARP` is a device-wide default for all VNIs. There was no way to override ARP suppression individually per VNI — either to explicitly enable it for a specific VNI while leaving others to inherit the default, or to disable it for a VNI while the global default has it enabled.

This PR adds an optional `suppressARP` field to `EVPNInstanceSpec`. When set, it overrides the NVE-level setting for that VNI's gNMI path. When unset (`nil`), the NVE-level default remains in effect.

## API change (`api/core/v1alpha1/evpninstance_types.go`):

```yaml
spec:
  suppressARP: true   # or false; omit to inherit NVE-level default
```

## Provider mapping (`internal/provider/cisco/nxos/`):

The NX-OS DME API represents this field as a string on Nw-list, not a boolean. The provider maps:

|`spec.suppressARP` | gNMI value |
|--------|--------|
| nil (unset) | "off" — inherit NVE-level setting |
| true | "enabled" |
| false | "disabled" |
			
`VNI.SuppressARP` is therefore a plain `string` field, avoiding any bool↔string unmarshal mismatch when reading device state back for idempotency checks.

# Changes
- `api/core/v1alpha1/evpninstance_types.go` — adds `SuppressARP *bool` to `EVPNInstanceSpec`
- `api/core/v1alpha1/zz_generated.deepcopy.go` — regenerated deep-copy
- `config/crd/bases/, charts/.../crd/` — regenerated CRD manifests
- `docs/api-reference/index.md` — updated API reference table
- `internal/provider/cisco/nxos/nve.go` — adds `VNI.SuppressARP` string and string constants
- `internal/provider/cisco/nxos/provider.go` — maps *bool → string constant in `EnsureEVPNInstance`
- `internal/provider/cisco/nxos/testdata/` — updates marshal testdata to include `suppressARP` in all VNI payloads
-` test/gnmi/testdata/cisco-nxos-gnmi/evpninstance.txtar` — updates expected gNMI state to include `"suppressARP": "off"`

# Testing
Updated `testdata/vni.json` to include `"suppressARP": "DME_UNSET_PROPERTY_MARKER"` (reflecting the new zero-value serialization of `Option[bool]`)
Added golden files and fixtures for explicit true and false cases (vni_suppress_arp_true, vni_suppress_arp_false)

## Simulated device

Tested against a simulated Cisco N9K-C9300v in Clabernets. Results of reading the path using `gnmic -a 127.0.0.1 --port 9339 -u *** -p *** --skip-verify get -e json --values-only --path 'System/eps-items/epId-items/Ep-list[epId=1]/nws-items/vni-items/Nw-list[vni=100010]/suppressARP'`

|`spec.suppressARP` | gNMI value |
|--------|--------|
| unset | `["off"]` |
| `true` | `["enabled"]` |
| `false` | `["disabled"]` |

# YANG

Reference https://github.com/YangModels/yang/blob/main/vendor/cisco/nx/10.6-3/Cisco-NX-OS-device.yang

Path: `System/eps-items/epId-items/Ep-list[epId=1]/nws-items/vni-items/Nw-list[vni={vni}]/suppressARP`

Allowed values of `suppressARP`: 
```yang
typedef nvo_SuppressARPT {
    type enumeration {
        enum off {
            value 0;
            description "Arp suppression not configured";
        }
        enum enabled {
            value 1;
            description "Arp suppression enabled";
        }
        enum disabled {
            value 2;
            description "Arp suppression disabled";
        }
    }
    default "off";
}
```


internal/provider/cisco/nxos/nve.go
```diff
@@ -64,10 +64,17 @@ func (n *NVE) XPath() string {
...
+	suppressARPOff      = "off"
+	suppressARPEnabled  = "enabled"
+	suppressARPDisabled = "disabled"
+)
+
 type VNI struct {
 	AssociateVrfFlag bool           `json:"associateVrfFlag"`
 	McastGroup       Option[string] `json:"mcastGroup"`
 	Vni              int32          `json:"vni"`
+	SuppressARP      string         `json:"suppressARP"`
```
felix-kaestner:
```suggestion
type SuppressARP string
const (
	SuppressARPOff      SuppressARP = "off"
	SuppressARPEnabled  SuppressARP = "enabled"
	SuppressARPDisabled SuppressARP = "disabled"
)

type VNI struct {
    AssociateVrfFlag bool           `json:"associateVrfFlag"`
    McastGroup       Option[string] `json:"mcastGroup"`
    Vni              int32          `json:"vni"`
    SuppressARP      SuppressARP         `json:"suppressARP"`
```
could we make this a real string enumeration?

internal/provider/cisco/nxos/testdata/vni_suppress_arp_false.json
felix-kaestner, line 1: nit: I don't think we need a variant for each of the different values for each fiel. The `internal/provider/cisco/nxos/testdata/vni.json` would already be sufficient in my view to showcase that we set this payload. Exact string values are easily checkable from the code itself.

felix-kaestner (approved):

PR #557, opened 2026-09-11
Move samples into config/samples/common

adamtrizuljak-sap:
Update samples directory layout to support per-vendor samples:
- Move the current samples that are common for all platforms from `config/samples` into `config/samples/common`
- `config/samples/cisco/nx` stays as-is
- Update Tiltfile
- Nokia SRLinux samples will be placed into `config/samples/nokia/srlinux` in a future PR

I would also propose to change the Tiltfile labels to `samples-common`, `samples-cisco-nx` or similar for a clearer layout of resources in Tilt


felix-kaestner (commented):
I think all samples we currently have in `config/samples` were developed/tested for Cisco NX-OS. As such, I would either move all them into `../cisco/nx` or keep then in their current location and only add a new `../nokia/srlinux` directory with Nokia SRLinux ones.

adamtrizuljak-sap:
> I think all samples we currently have in `config/samples` were developed/tested for Cisco NX-OS. As such, I would either move all them into `../cisco/nx` or keep then in their current location and only add a new `../nokia/srlinux` directory with Nokia SRLinux ones.

I wouldn't move the common samples into `../cisco/nx`, as that directory contains the provider-specific configurations with `apiVersion: nx.cisco.networking.metal.ironcore.dev`. 

So actually, it probably makes sense to leave the samples where they are and just put Nokia-specific samples into `../nokia/srlinux`. Therefore this PR can be discarded.

PR #558, opened 2026-09-11
Isolate per-device provider state in controller tests

felix-kaestner:
The test provider was a singleton shared by all controllers. Every test creates a distinct Device via GenerateName, but the provider stored all state (DNS, BGP, NTP, LLDP, etc.) in flat fields on a single struct. This caused state bleed between test suites and intermittent failures.

Refactor the test provider into three parts:

- DeviceState: per-device mutable state (all resource fields, ConnectFailure, LastRebootTime), each with its own mutex
- DeviceStore: shared map of device name to DeviceState, with a StateFor(name) accessor that creates entries on demand
- Provider: lightweight per-reconcile instance created by the ProviderFunc factory, capturing the device name during Connect via the new Connection.DeviceName field

Add DeviceName to deviceutil.Connection and set it in GetDeviceConnection so the mock provider can identify the device without changing any controller code.


adamtrizuljak-sap:
The idea seems sound. I tried to run the tests from this branch about 50x and all controllers seem to be working fine, except for the `ConfigBackup` controller which had 3 failures. You may want to look at it, but it may be out of scope of this PR to fix it.

[configbackup_fail.txt](https://github.com/user-attachments/files/32384685/configbackup_fail.txt)

felix-kaestner:
Nice catch @adamtrizuljak-sap. I added another commit in https://github.com/ironcore-dev/network-operator/pull/558/changes/fe078610812a67666c03623a4871d08c3d2115f2 to address this.

sven-rosenzweig (approved):

PR #562, opened 2026-09-14
Remove fabric name from managed resources names

nikatza:
The fabric controller prepends the fabric name to the generated resources names as in the following example:
BGPPeer/ora-1-fabric-ora-1-sw1101b-brd-ora-1-sw1100c-spine


felix-kaestner:
I think having the `Fabric` name being part of the resource names is actually a common pattern, see e.g. how a typical Deployment -> ReplicaSet -> Pod passes down the name. ref/ [https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/](https://kubernetes.io/docs/concepts/workloads/controllers/replicaset/#:~:text=When%20the%20control%20plane%20creates%20new%20Pods%20for%20a%20ReplicaSet%2C%20the%20%2Emetadata%2Ename%20of%20the%20ReplicaSet%20is%20part%20of%20the%20basis%20for%20naming%20those%20Pods%2E). 

For the described example of `ora-1-fabric-ora-1-sw1101b-brd-ora-1-sw1100c-spine`, the problem is rather that `ora-1` is part of all device names (coming from netbox in our case) and even part of the Fabric resource name. It could simply be omitted everywhere in our case, as the "ora-1" part is already clear from our internal target cluster/environment.

Generally speaking. If I have a Fabric resource called just `fabric` and two devices named `leaf1` and `spine1`, then the name `fabric-leaf1-spine1` is actually not that bad and also matches the style of how Pods of a Deployment end up being named.

nikatza:
Won't do (see comment above).

PR #564, opened 2026-09-14
feat: add ipv6 link-local address to Interface

ofrzeta:


felix-kaestner:
@ofrzeta Thanks a lot for your contribution! Looks pretty great. 🚀 

Would you be able to add a sample for an IPv6 Interface into the [config/samples/v1alpha1_interface.yaml](https://github.com/ironcore-dev/network-operator/blob/main/config/samples/v1alpha1_interface.yaml) file and also add it to the Tiltfile like [this](https://github.com/ironcore-dev/network-operator/blob/main/Tiltfile#L75)?

ofrzeta:
Hi @felix-kaestner, thanks for having a look. I've add the samples to the YAML and Tiltfile.

internal/provider/openconfig/interface.go
```diff
@@ -313,9 +322,42 @@ func (p *Provider) EnsureSubinterface(ctx context.Context, req *provider.EnsureI
...
+	v, ok := ipv6.(provider.IPv6AddressList)
+	if !ok {
+		return nil
+	}
+
+	addrs := &IPv6Addresses{}
+	for _, prefix := range v {
+		ip := prefix.Addr().String()
+		addrType := IPv6AddressTypeGlobalUnicast
+		if prefix.Addr().IsLinkLocalUnicast() {
```
felix-kaestner: This condition is not possible, through the check in https://github.com/ofrzeta/network-operator/blob/4ca5e0db50819379bd3d9371968cbeb2c8ac38eb/internal/webhook/core/v1alpha1/interface_webhook.go#L171, right?

internal/provider/openconfig/interface.go
```diff
@@ -500,6 +543,49 @@ type InterfaceIPv4Config struct {
 	Enabled bool `json:"enabled"`
 }
 
+// InterfaceIPv6 holds the IPv6 container of a subinterface.
+type InterfaceIPv6 struct {
+	Addresses *IPv6Addresses       `json:"addresses,omitempty"`
+	Config    *InterfaceIPv6Config `json:"config,omitempty"`
```
felix-kaestner:
```suggestion
	Addresses *IPv6Addresses       `json:"addresses"`
	Config    *InterfaceIPv6Config `json:"config"`
```
nit: As these fields are always set, we don't need the "omitempty" tag, as it won't have an effect. This also applies to the `IPv6Addresses.Address`, `IPv6Address.config` and `IPv6AddressConfig.type` fields.

internal/provider/openconfig/interface.go
ofrzeta: Right. Gonna remove it.

felix-kaestner (approved):

felix-kaestner:
@ofrzeta could you please squash all your commits into a single one and make sure it has a `Signed-off-by` line in it at the end. 🙏 This is an organization requirement currently blocking the merge.

PR #565, opened 2026-09-15
Diff list entries individually on Patch to remove stale config

felix-kaestner:
When the provider uses Patch (gNMI update) on a struct containing list fields, the device merges entries into existing config but never removes entries that are no longer desired. Stale list entries survive across reconciliations.

Fix this by removing list fields from parent structs and managing each list entry as a standalone DataElement. The provider now fetches current entries from the device, patches each desired entry individually, and deletes any stale entries not present in the desired set.


internal/provider/cisco/nxos/provider.go
```diff
@@ -1670,9 +1715,38 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
 	// Add the address items last, as they depend on the interface being created first.
 	if addr != nil {
 		sb.Patch(addr)
+		for _, a := range addrs {
+			sb.Patch(a)
+		}
+		currentAddrs := &IntfAddrItems{ID: name, Vrf: vrf}
+		if err := p.client.GetConfig(ctx, currentAddrs); err != nil && !errors.Is(err, gnmiext.ErrNil) {
```
sven-rosenzweig: nit: I think a short comment: "removing stale entries" would increase understandability. 
Same for in the ensure bgp, intf, ldldp, nve methods.

internal/provider/cisco/nxos/provider.go
```diff
@@ -1670,9 +1715,38 @@ func (p *Provider) EnsureInterface(ctx context.Context, req *provider.EnsureInte
...
+			return err
+		}
+		for _, a := range currentAddrs.AddrList {
+			if _, ok := addrs.Get(a.Key()); !ok {
+				a.ID = name
```
sven-rosenzweig: nit: Comment like setting the key values to delete from our gnmi.List would help.

sven-rosenzweig (approved):

PR #568, opened 2026-09-16
Add ConsoleConnection CRD and controller for console health checks

felix-kaestner:
Implement a ConsoleConnection resource that verifies serial console connectivity to devices via SSH console servers. The controller supports three verification strategies (Wait, SendCRLF, SendChar) and reports five health states: ConsoleServerUnreachable, ConsoleServerAuthFailure, Dead, Alive, and Verified.

Includes envtest integration tests with an in-process SSH server, sample manifests for all verification strategies, and Tiltfile integration.


internal/controller/core/consoleconnection_controller.go
```diff
@@ -0,0 +1,478 @@
...
+		last := obj.CreationTimestamp.UTC()
+		if obj.Status.LastCheckTime != nil {
+			last = obj.Status.LastCheckTime.UTC()
+		}
+
+		// If the next scheduled check is in the future, requeue until that time.
+		// Otherwise, continue to check now.
+		if now, next := time.Now().UTC(), schedule.Next(last); next.After(now) {
+			obj.Status.NextCheckTime = &metav1.Time{Time: next}
+			r.Recorder.Eventf(obj, nil, "Normal", "Scheduled", "Reconcile", "Next console check scheduled at %s", next.Format(time.RFC3339))
```
adamtrizuljak-sap: Do we need to log this as a k8s event? I think that a simple info or debug log would be enough here.

On the other hand, recording the result of a check as an event is fine.

internal/controller/core/consoleconnection_controller.go
```diff
@@ -0,0 +1,478 @@
...
+			}
+			next := schedule.Next(time.Now().UTC())
+			obj.Status.NextCheckTime = &metav1.Time{Time: next}
+			r.Recorder.Eventf(obj, nil, "Normal", "Scheduled", "Reconcile", "Next console check scheduled at %s", next.Format(time.RFC3339))
+			res.RequeueAfter = time.Until(next)
+		}()
+	}
+
+	if schedule == nil && obj.Status.LastCheckTime != nil {
+		r.Recorder.Eventf(obj, nil, "Normal", "CheckCompleted", "Reconcile", "One-shot check already completed at %s", obj.Status.LastCheckTime.String())
```
adamtrizuljak-sap: Also here, I would just log this...

api/core/v1alpha1/consoleconnection_types.go
```diff
@@ -0,0 +1,206 @@
...
+	// DeviceRef is a reference to the Device this console connection targets.
+	// The Device object must exist in the same namespace.
+	// Immutable.
+	// +required
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="DeviceRef is immutable"
+	DeviceRef LocalObjectReference `json:"deviceRef"`
+
+	// Endpoint contains the console server connection details.
+	// +required
+	Endpoint ConsoleEndpoint `json:"endpoint"`
```
adamtrizuljak-sap: Since we already have `DeviceRef`, it would be nice if we could obtain the endpoint address directly from it and save the user from having to repeat the device IP address in another resource.

adamtrizuljak-sap (commented):
Looks good upon the first pass.

PR #569, opened 2026-09-16
add (Ext)CommunitySet and NXOS support

IvoGoman:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


internal/controller/core/communityset_controller.go
```diff
@@ -0,0 +1,302 @@
...
+			reterr = kerrors.NewAggregate([]error{reterr, err})
+		}
+	}()
+
+	conn, err := deviceutil.GetDeviceConnection(ctx, r, device)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	var cfg *provider.ProviderConfig
```
felix-kaestner: Can you look into https://github.com/ironcore-dev/network-operator/commit/8b8c868cebfbb65c256a939b70015cedb3573ddf, as we decoubled the provider config ref fetching to not be part of the finalization and hence moved after the `if !obj.DeletionTimestamp.IsZero() {` check. (Applies to both controllers of this PR).

internal/controller/core/communityset_controller.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1alpha1.CommunitySet{}, v1alpha1.DeviceRefIndexKey, func(obj client.Object) []string {
+		o := obj.(*v1alpha1.CommunitySet)
+		return []string{o.Spec.DeviceRef.Name}
+	}); err != nil {
+		return err
+	}
+
+	return ctrl.NewControllerManagedBy(mgr).
```
felix-kaestner: Here we are missing a Watch on the ProviderConfigRef like https://github.com/ironcore-dev/network-operator/blob/main/internal/controller/core/interface_controller.go#L302-L311

api/core/v1alpha1/communityset_types.go
```diff
@@ -0,0 +1,109 @@
...
+	return c.Status.Conditions
+}
+
+// SetConditions implements conditions.Setter.
+func (c *CommunitySet) SetConditions(conditions []metav1.Condition) {
+	c.Status.Conditions = conditions
+}
+
+// +kubebuilder:object:root=true
+
```
felix-kaestner: If we support provider config refs then we would need to also have a structure like https://github.com/ironcore-dev/network-operator/blob/main/api/core/v1alpha1/interface_types.go#L654-L663.

internal/provider/cisco/nxos/communityset.go
```diff
@@ -0,0 +1,106 @@
...
+
+var (
+	_ provider.CommunitySetProvider    = (*Provider)(nil)
+	_ provider.ExtCommunitySetProvider = (*Provider)(nil)
+)
+
+// CommunityList represents a named BGP standard community-list on NX-OS.
+type CommunityList struct {
+	Name     string            `json:"name"`
+	Mode     string            `json:"mode"`
```
felix-kaestner:
```go
    Type     CommunityType     `json:"-"`
```
What about adding a type field like this and then use it inside of the `XPath()` method to construct the path. Then we don't need two structs and everything and can just use one.

felix-kaestner (commented):
Given that `CommunitySet` and `ExtCommunitySet` have the same api+controller implementation and that even the provider implementation for NX-OS is identical (just a different YANG XPath), shall we just merge the two and have a single `CommunitySet` API that is able to express both, via a 

```go
//+kubebuilder:Enum=Regular;Extended
Type CommunityType `json:"type"`
```

as part of the spec?

internal/controller/core/extcommunityset_controller.go
```diff
@@ -0,0 +1,302 @@
...
+		return ctrl.Result{}, err
+	}
+
+	var cfg *provider.ProviderConfig
+	if obj.Spec.ProviderConfigRef != nil {
+		cfg, err = provider.GetProviderConfig(ctx, r, obj.Namespace, obj.Spec.ProviderConfigRef)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
rgildein: Not actually related to only these changes, but would not be possible to make function from this part? I assume that most of if not all Reconcile are following pattern like this.

rgildein (commented):
LGTM, I reviewed it and come only with question, but it's not related with only this PR and tested against nx-os device, works fine. 

One more thing is do we want to add this also for OpenConfig provider? It's even possible? If not I would suggest to create these CRDs under the nx api.

PR #575, opened 2026-09-16
Support unnumbered interface-based BGP peering

ofrzeta:
Allow a BGPPeer to name an Interface instead of a peer address, so an eBGP session can run over the peers' IPv6 link-local addresses and the transit link needs no addressing of its own. The referenced Interface must set spec.ipv6.useLinkLocalOnly.

spec.address becomes optional and is mutually exclusive with the new spec.interfaceRef; spec.localAddress is meaningless for such a peer. Both rules are enforced by CEL and by the webhook. A peer's identity, its address, interfaceRef or bgpRef, is immutable.


felix-kaestner:
Hey @ofrzeta could you be so kind and rebase this PR, there seem to be some conflicts. 🙏

PR #584, opened 2026-09-18
Allow finalization without provider config

nikatza:
More info in commit message.


felix-kaestner:
> Provider-specific configuration affecting device model paths outside the main resource may remain on the device because that configuration is unavailable during cleanup. 

Ideally, the clean up of configuration on the device, done in the `finalize` of the reconcilers, will make sure to clean all paths eventually written, regardless of previous desired config. So as long as the ProviderConfig doesn't include any identifying criteria of a YANG path, I think we should be good to just always delete/cleanup the yang tree and as such don't even need to pass the ProviderConfig at all to the `Delete...` methods of the provider implementation.

nikatza:
> > Provider-specific configuration affecting device model paths outside the main resource may remain on the device because that configuration is unavailable during cleanup.
> 
> Ideally, the clean up of configuration on the device, done in the `finalize` of the reconcilers, will make sure to clean all paths eventually written, regardless of previous desired config. So as long as the ProviderConfig doesn't include any identifying criteria of a YANG path, I think we should be good to just always delete/cleanup the yang tree and as such don't even need to pass the ProviderConfig at all to the `Delete...` methods of the provider implementation.

I seems this is actually possible with the models we have. I have updated this PR. Thanks for looking at this.

internal/provider/cisco/nxos/intf_test.go
```diff
@@ -4,12 +4,51 @@
...
+			})
+			return nil
+		},
+	}
+	p := &Provider{client: client}
+	req := &provider.DeleteInterfaceRequest{Interface: &corev1alpha1.Interface{
+		Spec: corev1alpha1.InterfaceSpec{Name: "Port-Channel1", Type: corev1alpha1.InterfaceTypeAggregate},
+	}}
+
+	for range 2 {
```
felix-kaestner: Why is there a range over 2?

internal/provider/cisco/nxos/lldp.go
```diff
@@ -24,6 +32,11 @@ func (*LLDP) XPath() string {
 
 func (*LLDP) IsListItem() {}
 
+func (l *LLDP) Default() {
+	l.HoldTime = NewOption(defaultLLDPHoldTime)
```
felix-kaestner: If these values have a platform default, then I wonder why we use an `Option` type at all? Shouldn't we rather then use the platform default in case no explicit value is provided from the spec. Because if we `Default()` this configuration anyways, then it seems like we can't really remove these fields (like a "DME_UNSET_PROPERTY_MARKER" would) anyways.

internal/provider/cisco/nxos/lldp_test.go
```diff
@@ -22,3 +30,44 @@ func init() {
...
+		UpdateFunc: func(_ context.Context, elements ...gnmiext.DataElement) error {
+			for _, element := range elements {
+				paths = append(paths, "update "+element.XPath())
+			}
+			return nil
+		},
+	}
+	p := &Provider{client: client}
+
+	for range 2 {
```
felix-kaestner: Same here, I don't get why we range over 2.

internal/provider/cisco/nxos/term_test.go
```diff
@@ -14,3 +22,25 @@ func init() {
...
+			}
+			return nil
+		},
+	}
+	p := &Provider{client: client}
+
+	if err := p.DeleteManagementAccess(t.Context(), new(provider.DeleteManagementAccessRequest)); err != nil {
+		t.Fatalf("DeleteManagementAccess() error = %v", err)
+	}
+	want := new(VTYAccessClass).XPath()
```
felix-kaestner: Here we don't range. So this is what I would also expect in the other test functions.

internal/provider/cisco/nxos/intf_test.go
nikatza: This is wrong. Fixed.

felix-kaestner: Did you push your changes?

internal/provider/cisco/nxos/lldp.go
nikatza: Indeed, I have removed the `Option` and use the defaults.

felix-kaestner (approved):

